PR #37, opened 2020-05-05
Change the DB used in the CI from MySQL to PostgreSQL

sky-joker:
##### SUMMARY
Using MySQL with Zabbix introduces the following bugs.

https://support.zabbix.com/browse/ZBX-17357

This bug does not occur when using PostgreSQL, so change DB from MySQL to PostgreSQL.
After making this change, the case of the username will be strictly evaluated.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
docker-compose.yml
integration/targets/setup_zabbix/defaults/main.yml

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,32 +2,29 @@
 version: '3.7'
 services:
   zabbix-db:
-    image: mysql:5.7
+    image: postgres:latest
     environment:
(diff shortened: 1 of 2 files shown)


D3DeFi:
Makes sense :) Thanks for implementing it

PR #47, opened 2020-05-13
Fix to run zabbix 5.0 in CI

sky-joker:
##### SUMMARY

This PR fixes `docker-compose` and `ansible-test.yml` to make zabbix 5.0 work with CI.  
It was changed container listen port from 80 to 8080 by Zabbix developer from zabbix version 5.0, so it is necessary to separate the listen port number according to the version of zabbix.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

github/workflows/ansible-test.yml  
docker-compose.yml  
README.md

##### ADDITIONAL INFORMATION

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -26,11 +26,17 @@ jobs:
   integration:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        zabbix_version:
-          - "3.0"
-          - "4.0"
-          - "4.4"
+        zabbix_container:
+          - version: "3.0"
+            port: "80"
+          - version: "4.0"
+            port: "80"
+          - version: "4.4"
+            port: "80"
+          - version: "5.0"
+            port: "8080"
 
     steps:
       - name: Check out code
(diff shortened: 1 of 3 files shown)


D3DeFi:
Merging because I plan to push fixes for zabbix_host soon and would rather have them failing test for other modules rather than not having 5.0 tests at all

good work and thank you

D3DeFi:
Tests in `zabbix_host` exited early and left _ExampleHost_ configured on the zabbix system, thus followed by fail in `zabbix_template` regarding _Host already exists_.

As of now, I wouldn't consider it a bug as the check is written in a way that it expects a new host to be created. So only `zabbix_proxy` fix remaining and we should have 5.0 CI passing reliably.

D3DeFi:
Seems that ubuntu-4.0-latest image got updated 30 minutes ago and is now exposing zabbix-web via tcp/8080 as 5.0-latest does. 4.4-latest and 3.0-latest, however, are still using tcp/80 :(

Lets hope that all of them gets eventually bumped to tcp/8080 so we can have it unified once again.

sky-joker:
> I will submit PR for it once all our major versions are using the same port

okay :)
Thank you!

PR #51, opened 2020-05-18
Updated zabbix_host module to support Zabbix 5.0

### LINKED ISSUE #46: host.create with SNMP interface not working with Zabbix 5.0
##### SUMMARY

Zabbix 5.0 introduced a change in the host interface handling. The [new documentation](https://www.zabbix.com/documentation/current/manual/api/reference/hostinterface/object) mentions a new `details` parameter which is used only for SNMP interfaces.

In order to create a new host with a SNMP interface this new property has to be set. Unfortunately the documentation seems to be not clear about which fields are mandatory and which are optional. There is [ZBX-17719](https://support.zabbix.com/browse/ZBX-17719) handling this topic.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

zabbix_host

##### ANSIBLE VERSION

```paste below
ansible 2.9.6
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/ansible/.venv/lib/python3.6/site-packages/ansible
  executable location = /root/ansible/.venv/bin/ansible
  python version = 3.6.8 (default, Aug  7 2019, 17:28:10) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

I am trying to create zabbix hosts for our HPE iLOs. They are being monitored via SNMP.
For this procedure I am using the following task in an ansible playbook.

```yaml
  zabbix_host:
    server_url: '{{ monitoring_server }}'
    login_user: '{{ monitoring_user }}'
    login_password: '{{ monitoring_password }}'
    host_name: '{{ ilo_ip }}'
    visible_name: '{{ dns_fqdn }}'
    host_groups: '{{ monitoring_groups | default(["iLO"]) }}'
    link_templates: '{{ monitoring_templates | default(["Template Server HP iLO SNMPv2"]) }}'
    interfaces:
    - type: 2
      main: 1
      useip: 1
      ip: '{{ ilo_ip }}'
      dns: ""
    status: '{{ monitoring_status | default("enabled") }}'
    state: '{{ monitoring_state | default("present") }}'
    validate_certs: no
```

##### EXPECTED RESULTS

The playbook should create the host. That has worked on Zabbix 4.4, while it is not working on Zabbix 5.0.

##### ACTUAL RESULTS

I get an error stating "Error -32602: Invalid params., Incorrect arguments passed to function."

```paste below
fatal: [*** -> ***]: FAILED! => {"changed": false, "msg": "Failed to create host ***: ('Error -32602: Invalid params., Incorrect arguments passed to function. while sending {\"jsonrpc\": \"2.0\", \"method\": \"host.create\", \"params\": {\"host\": \"***\", \"interfaces\": [{\"type\": 1, \"main\": 1, \"useip\": 1, \"ip\": \"127.0.0.1\", \"dns\": \"\", \"port\": \"10050\"}, {\"type\": 2, \"main\": 1, \"useip\": 1, \"ip\": \"***\", \"dns\": \"\", \"port\": \"161\"}], \"groups\": [{\"groupid\": \"22\"}], \"status\": 0, \"tls_connect\": 1, \"tls_accept\": 1, \"name\": \"***\"}, \"auth\": \"***\", \"id\": 5}', -32602)"}
```

D3DeFi:
##### SUMMARY
Zabbix 5.0 support for `zabbix_host` module.

Game breaking changes were:
- `hostinterface.get` now returns `details: []` for each interface type, even when not needed
- `macro.get` now returns `type: 0` for text macros and new type `type: 1` is for secret macros
- secret macros return from API without `value` property so there is no way to check if user wants to
change the value of macro. We can either always change it or require user to change something alongside it like `description`. I went with the later to not break idempotency of the module.

Fixes #46

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/zabbix_host.py

##### ADDITIONAL INFORMATION
I've finally added `options` for `interfaces` into `module.argument_spec`. The downside is that ansible now prints warnings when type is an integer value as it is converted into a string:

```yaml
  interfaces:
    - type: 1
       main: 1
...omitted...
```

This was added in a good faith by me in the past (`interfaces[0]['type'] = 'agent'`). Users shouldn't need to bother with API details and it would make sense to change every numerical value to either bool or string in the future, but it cannot be done without breaking backwards compatibility. 

Just for example - we would have `interfaces[0]['main'] = True` instead of `interfaces[0]['main'] = 1`. I don't know if this makes sense and if we want to pursue this in the future or if I should convert `interfaces[0]['type']` back to an integer. `zabbix_action` is using this logic and doesn't use numerical values anywhere where it doesn't make sense.

SNMP details are required, Zabbix documentation is broken - https://support.zabbix.com/browse/ZBX-17719

There is an additional issue, where `details` should be a dictionary and is accepted by an API as such. When `details` is empty, API returns empty list - `details: []` instead, hence the conversion in the code to empty dict.

### DIFF
--- a/plugins/modules/zabbix_host.py
+++ b/plugins/modules/zabbix_host.py
@@ -89,8 +89,10 @@ options:
             - List of interfaces to be created for the host (see example below).
             - For more information, review host interface documentation at
             - U(https://www.zabbix.com/documentation/4.0/manual/api/reference/hostinterface/object)
+        default: []
         suboptions:
             type:
+                type: str
                 description:
                     - Interface type to add
                     - Numerical values are also accepted for interface type
@@ -98,7 +100,7 @@ options:
                     - 2 = snmp
                     - 3 = ipmi
                     - 4 = jmx
-                choices: ['agent', 'snmp', 'ipmi', 'jmx']
+                choices: ['agent', '1', 'snmp', '2', 'ipmi', '3', 'jmx', '4']
                 required: true
             main:
                 type: int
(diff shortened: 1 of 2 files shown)


D3DeFi:
Failing integration tests for `zabbix_proxy` on 5.0 as expected. Other than that, looks like it might work.

@toerb @JohannesKreuzer if you can give this changes some tests on your side, it would be very much appreciated.

toerb:
@D3DeFi looks good! New hosts are created properly and already existing one are not changed if there is no need for it. This solves the problem described in #46.

sky-joker:
Excellent!!
Thank you @D3DeFi for this patch :)

LGTM

> Failing integration tests for zabbix_proxy on 5.0 as expected. Other than that, looks like it might work.

I agree with you.

Issue #50, opened 2020-05-17
Integration test error occur of zabbix_proxy with zabbix 5.0


sky-joker:
##### SUMMARY

The integration test result for zabbix_proxy, changed occurs when performing the same task that has not been parameters changed.  
It is necessary review the configuration comparison process.

```
(venv) [root@zbxcont zabbix]# ansible-test integration zabbix_proxy
(snip)
TASK [zabbix_proxy : test - create new passive Zabbix proxy server] ************************************************************************************************
[WARNING]: The value 10051 (type int) in a string field was converted to '10051' (type string). If this does not look like what you expect, quote the entire value
to ensure it does not change.
changed: [testhost]

TASK [zabbix_proxy : assert that proxy was created] ****************************************************************************************************************
ok: [testhost] => {
    "changed": false,
    "msg": "All assertions passed"
}

TASK [zabbix_proxy : test - create same passive Zabbix proxy server] ***********************************************************************************************
changed: [testhost]

TASK [zabbix_proxy : assert that nothing has been changed] *********************************************************************************************************
fatal: [testhost]: FAILED! => {
    "assertion": "not zbxproxy_existing is changed",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}

PLAY RECAP *********************************************************************************************************************************************************
testhost                   : ok=7    changed=2    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

##### ISSUE TYPE

- Bug Report

##### COMPONENT NAME

zabbix_proxy

##### ANSIBLE VERSION

```
(venv) [root@zbxcont zabbix]# ansible --version
ansible 2.10.0.dev0
```

##### OS / ENVIRONMENT

* zabbix 5.0

##### ADDITIONAL INFORMATION

D3DeFi:
@sky-joker just to point you to a right direction and save you some time. I am dealing with probably the same problem in `zabbix_host` right now:

Zabbix 5.0 now contains `details` property for `hostinterface` objects:
https://www.zabbix.com/documentation/current/manual/api/reference/hostinterface/object#host_interface

and since Zabbix API is returning `details: []` for every interface type and not just `type=2`, it may as well be the case for proxy with a passive host interface.

Btw, API should return `details: {}` as it is not an array, but an dictionary, but not a big problem with proxies as they don't use other than `type='agent'` interfaces.

sky-joker:
Thank you @D3DeFi for tell me that information.

[...]

PR #53, opened 2020-05-19
Updated zabbix_proxy module to support Zabbix 5.0

sky-joker:
##### SUMMARY
This PR is a modification of the `zabbix_proxy` module to support Zabbix 5.0.

Fixes https://github.com/ansible-collections/community.zabbix/issues/50

The interface configuration comparison process didn't work well because `details(list)` was converted to `str`.
I've fixed `details(list)` to convert to a string and compare.
Converting `details(list)` property to a string does not affect the creation or update of the `proxy(type=passive)` because the `details(list)` property is not required when creating or updating.

https://www.zabbix.com/documentation/current/manual/api/reference/proxy/object#proxy_interface

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/zabbix_proxy.py

### DIFF
--- a/plugins/modules/zabbix_proxy.py
+++ b/plugins/modules/zabbix_proxy.py
@@ -204,6 +204,7 @@ RETURN = r''' # '''
 import traceback
 import atexit
 
+from distutils.version import LooseVersion
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 try:
     from zabbix_api import ZabbixAPI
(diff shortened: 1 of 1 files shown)


sky-joker:
The sanity test is failed because https://status.quay.io/ is down.
If `quay.io` is recovered, I'll run the test again.

D3DeFi:
Thanks for looking into it! :)

PR #63, opened 2020-05-23
Add zabbix_maintenance integration test

sky-joker:
##### SUMMARY
This PR adds zabbix_maintenance module integration test.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_maintenance

### DIFF
--- a/tests/integration/targets/zabbix_maintenance/meta/main.yml
+++ b/tests/integration/targets/zabbix_maintenance/meta/main.yml
@@ -1,0 +1,2 @@
+dependencies:
+  - setup_zabbix
--- a/tests/integration/targets/zabbix_maintenance/tasks/main.yml
+++ b/tests/integration/targets/zabbix_maintenance/tasks/main.yml
@@ -1,0 +1,229 @@
+---
+# New host create test from here
+- name: "test - Create a new host"
+  zabbix_host:
+    server_url: "{{ zabbix_server_url }}"
+    login_user: "{{ zabbix_login_user }}"
+    login_password: "{{ zabbix_login_password }}"
+    host_name: example
+    host_groups:
+      - Linux servers
+    interfaces:
+      - type: 1
+        main: 1
+        useip: 1
+        ip: 10.1.1.1
+        dns: ""
+        port: 10050
+  register: create_host_result
+
+- assert:
+    that:
+      - create_host_result.changed is sameas true
+
+- name: "test - Create maintenance with a host_name param"
+  zabbix_maintenance:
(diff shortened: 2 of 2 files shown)


D3DeFi:
I wonder why this works. Module seems to set `start_time` of the maintenance to something like`time.now()` every time it is run.

sky-joker:
Thank you @D3DeFi for merging.  
Sorry, I late in checking your comment.  

I noticed that when I tried to update only the desc parameter, it didn't update.  
The cause of this is under investigation.

> Module seems to set start_time of the maintenance to something liketime.now() every time it is run.

It would be nice to add `start_time` and `end_time` parameters to this module, what do you think?

D3DeFi:
> It would be nice to add `start_time` and `end_time` parameters to this module, what do you think?

Can you please expand the idea a bit further? I am not sure if I fully understand how it would work. 

Do we, for example, require users to depend on the ansible fact `ansible_date_time.time`? But in that case `start_time` would always change as well when rerunning the playbook multiple times.

Only use case that came to my mind is that if you for example wanted to schedule maintenance each day for a specific hour, then this would make sense. On the other hand, you can still schedule recurring maintenance once and be done with it. Maybe users are handling something like this already by running playbook at the specific hour via cron/Ansible Tower/etc..

I presume that most popular use case right now is to put server to a maintenance, do some grunt work like OS upgrade and then remove the maintenance. Judging by the design of the module, I believe that it was written with this intention as well. But there is downside that it is not idempotent right now :(

sky-joker:
Thank you for your opinion.  

> I presume that most popular use case right now is to put server to a maintenance, do some grunt work like OS upgrade and then remove the maintenance.

I see, Is it like this? (write to playbook to the following processing like)

* register for the maintenance
* something process execute(update and so on)
* unregister for the maintenance

For example, I thought it would be useful to be able to specify the following parameters in the module as well.

```yaml
- zabbix_zabbix_maintenance:
    (snip)
    start_time: yyyy/mm/dd HH:MM:ss
    end_time: yyyy/mm/dd HH:MM:ss
    (snip)
```

This is in anticipation of registering for maintenance in advance.  
There may not be much demand for it, but...

D3DeFi:
> There may not be much demand for it, but...

Well, I ain't against it :) Would you be interested in proposing a feature PR for this? If so, it would be welcome. I am sure it would find its uses eventually. But just to be sure, let's leave default behavior as it is right now.

sky-joker:
Thank you :)
I'll think about this feature a bit more.

Issue #49, opened 2020-05-15
zabbix_template creates template in check mode


chemnic:
##### SUMMARY

The module zabbix_template creates a new template when in check mode with zabbix 4.0.
I use template_json to import a new template with just the enough data to create it.
The normal mode (not in check mode) works as expected: Shows "changed" when the template doesn't exist, and just "ok" when it already exists.
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

zabbix_template
##### ANSIBLE VERSION

```paste below
ansible 2.9.9
  config file = /home/chestrada/ansible/vfmsa_ansible/ansible.cfg
  configured module search path = ['/home/chestrada/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/chestrada/xxx/ansible/lib/python3.6/site-packages/ansible
  executable location = /home/chestrada/xxx/ansible/bin/ansible
  python version = 3.6.9 (default, Apr 18 2020, 01:56:04) [GCC 8.4.0]

```

##### CONFIGURATION

```paste below
ANSIBLE_PIPELINING(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = True
ANSIBLE_SSH_ARGS(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = -o ControlMaster=auto -o ControlPersist=1800s -o PreferredAuthentications=publickey
CACHE_PLUGIN(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = ~/ansible_cache
CACHE_PLUGIN_TIMEOUT(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = 86400
DEFAULT_CALLBACK_PLUGIN_PATH(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = ['/home/chestrada/py_ansible/lib/python3.6/site-packages/ara/plugins/callba
DEFAULT_FORKS(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = 50
DEFAULT_GATHERING(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = smart
DEFAULT_INTERNAL_POLL_INTERVAL(/home/chestrada/ansible/vfmsa_ansible/ansible.cfg) = 0.001
[...]

PR #69, opened 2020-06-04
dump_template from zabbix_template module should not exit module even…

D3DeFi:
… in check mode

##### SUMMARY
Old behavior found in #49 

I believe that `zabbix_template` should return actual template with `state=dump` and `--check-mode`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME


sky-joker:
LGTM :)

Thank you @D3DeFi

PR #72, opened 2020-06-09
Renaming the names of the roles

dj-wasabi:
##### SUMMARY

As it provides error while using a dash (-) in the role name, we will use an underscore (_)

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

roles zabbix*

##### ADDITIONAL INFORMATION

```paste below

```


dj-wasabi:
@D3DeFi  Can i merge? Or should I wait until you have finished the ci part that roles are excluded during runs?

D3DeFi:
> @D3DeFi Can i merge? Or should I wait until you have finished the ci part that roles are excluded during runs?

merge. Sorry I was away

D3DeFi:
ouchie, integration tests are failing because it uses `roles/zabbix_proxy` instead of `test/integration/targets/zabbix_proxy`.

D3DeFi:
I've pushed fix for sanity tests and will continue searching for workaround on integrations tests tomorrow as it seems to be a bug in `ansible-test` itself. If nothing else works we will simply rename roles in `test/integration/targets` to something like `zabbix_proxy_module`

Issue #15, opened 2020-04-24
Initial commit into Ansible Galaxy


ericsysmin:
##### SUMMARY
This collection is not yet available in Ansible Galaxy, can we have someone upload it?

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below

```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

D3DeFi:
Hi Eric and thanks for raising this issue, I agree that it is the high time to move with this.

@gundalow is this possible now? Last time we spoke on a chat you mentioned something about zuul tests if I remember correctly and that we need to wait for them to be able to release on galaxy.

Another problem is that I don't have permissions to release into the https://galaxy.ansible.com/community namespace.

sky-joker:
Thank you @D3DeFi and @ericsysmin
I was just about to create an issue about this question.

gundalow:
Hi,
I've manually built and release `0.1.0` https://galaxy.ansible.com/community/zabbix
It's still on my todo list to add automatic building on git tag, let me see if I can push that towards the top of my list.

sky-joker:
Thank you @gundalow :)

ericsysmin:
@gundalow thank you, was trying to use this collection as I am moving everything to use collections. :)

PR #74, opened 2020-06-10
Renamed integration roles to prevent ansible-test confusing them with…

D3DeFi:
… roles from roles/ directory

##### SUMMARY
Since we decided to rename roles to `zabbix_`, `ansible-test` seems to prefer `roles/*` over `test/integration/targets/*`.

Not even sure if this is intended feature, but I've opened a bug report in ansible/ansible#69981

To fix this on our side, I've decided to go with KISS principle and simply rename our integration test roles rather than look for more hacking solution like this:

`tests/integration/targets/zabbix_host -> tests/integration/targets/test_zabbix_host`

Follow up to #72

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/integration/targets/*


D3DeFi:
@sky-joker can you take a look please? Does this make sense to you?

D3DeFi:
I will open a PR from `devel-roles-migration` to `master` once we merge this simple rename. I wouldn't be waiting for a fix in ansible as that can take some time and this migration already took us long enough.

I would like to tag `0.2.0` during the weekend and ask for galaxy release. Do you think it is achievable @sky-joker ?

sky-joker:
I wonder do I need to ask @gundalow to make a config that automatically pushes to the galaxy when I add a tag?

https://github.com/ansible-collections/community.zabbix/issues/15#issuecomment-619150845

D3DeFi:
Thanks for looking at this!

> I wonder do I need to ask @gundalow to make a config that automatically pushes to the galaxy when we add a tag?

We probably still need to

gundalow:
Yup, I'm a bit behind on that :(
If you are happy to wait a but before pushing a gut tag, we can test here next week?

D3DeFi:
I believe we can @gundalow :) At least I will have more time to format CHANGELOG entries :)

sky-joker:
> If you are happy to wait a but before pushing a gut tag, we can test here next week?

@gundalow, please test next week :)

PR #80, opened 2020-06-13
Devel roles migration

D3DeFi:
##### SUMMARY
Merge changes approved in `devel-roles-migration` to `master`. PR for CHANGELOG entries will follow.

This precedes tagging of `0.2.0` and releasing collection with roles on Galaxy! hoorayy! :)

Fixes #24 
Fixes #52

PR #85, opened 2020-06-15
updated CHANGELOG.md to reflect changes made in the collection

D3DeFi:
##### SUMMARY
Changelog before 0.2.0 release

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,36 @@
 
 ## devel
 
+### New roles:
+  - `zabbix_agent` previously known as [dj-wasabi/zabbix-agent](https://galaxy.ansible.com/dj-wasabi/zabbix-agent)
+  - `zabbix_javagateway` previously known as [dj-wasabi/zabbix-javagateway](https://galaxy.ansible.com/dj-wasabi/zabbix-javagateway)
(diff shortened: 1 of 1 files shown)


D3DeFi:
@dj-wasabi @sky-joker do we want to do these steps before releasing?:

- prepare some migration documentation for user (e.g. to rename roles, use FQCNs, and so on)
- prepare some explicit documentation about which ansible roles and python modules are required for each role/module - our current doc seems to be a little bit vague on this one?

Or do it after release?

I also think that we have some time until Ansible 2.10 will be released as users would probably not be migrating to collections until then. At least not in a masses.

dj-wasabi:
I would say, lets release the 0.2.0 with what we have now. In the mean time we can get some experience and fixing some issues that people have now before the masses are using the collection and have issues. My 2 cents.

D3DeFi:
my bad requesting review on already approved changes in #80

D3DeFi:
> I would say, lets release the 0.2.0 with what we have now. In the mean time we can get some experience and fixing some issues that people have now before the masses are using the collection and have issues. My 2 cents.

I agree, I am going to push git tag now

PR #100, opened 2020-06-20
adjusted CI settings and naming to indicate that they do not include …

D3DeFi:
…roles

##### SUMMARY
Modified CI settings to explicitly state that they have nothing to do with roles (as to prevent confusion) and that they only do testing of relevant changes.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
.github/workflows/ansible-test.yml

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -1,6 +1,19 @@
-name: CI
+name: plugins
 on:
-- pull_request
+  push:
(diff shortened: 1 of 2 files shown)


D3DeFi:
When we merge CI for roles it will be more clear which Actions are for which components:
https://github.com/ansible-collections/community.zabbix/actions

PR #106, opened 2020-06-23
add an own bug_report template

rockaut:
##### SUMMARY
It's basically the same from https://github.com/ansible-collections/.github/tree/master/.github/ISSUE_TEMPLATE but with "Zabbix Version" added.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,59 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: 
+labels: 
+assignees: 
+
+---
+
+<!--- Verify first that your issue is not already reported on GitHub -->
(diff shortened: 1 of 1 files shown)


D3DeFi:
@gundalow is it ok if we adjust issue template? Maybe some more in the future to suite more our needs

gundalow:
Yup, you can adjust this however you want.

For example, add a question about what Zabbix version is installed, any relevant zabbix configuration, etc.

As long as the file name & title matches whats in https://github.com/ansible-collections/.github/tree/master/.github/ISSUE_TEMPLATE the files in the community.zabbix repo will be used instead of the generic version

D3DeFi:
That means we need to have other templates present in the directory as well, right?

gundalow:
Nop, you can just override the ones you want to

PR #105, opened 2020-06-23
Run CI against the correct branch

gundalow:
##### SUMMARY
This is likely my fault, though we've been running CI against a branch that hasn't been updated in a long time.

I expect a load of things to fail with this change, though we can push further changes to this branch till all is green.

**2.10**
- [x] ERROR: plugins/modules/zabbix_action.py:994:103: E225: missing whitespace around operator
- [x] ERROR: plugins/modules/zabbix_group_info.py:98:8: ansible-deprecated-no-collection-name: No collection name found in call to Display.deprecated or AnsibleModule.deprecate
- [X] ERROR: plugins/modules/zabbix_host_info.py:193:8: ansible-deprecated-no-collection-name: No collection name found in call to Display.deprecated or AnsibleModule.deprecate
- [X] ERROR: plugins/modules/zabbix_template.py:730:8: ansible-deprecated-no-collection-name: No collection name found in call to Display.deprecated or AnsibleModule.deprecate
- [X] ERROR: plugins/modules/zabbix_proxy.py:0:0: deprecation-collection-missing: AnsibleModule.argument_spec.interface.options.main: If removed_in_version or removed_at_date is specified, removed_from_collection must be specified as well for dictionary value @ data['argument_spec']['interface']['options']['main']. Got {'type': 'int', 'default': 0, 'removed_in_version': '2.14'}
- [X] ERROR: plugins/modules/zabbix_proxy.py:0:0: deprecation-collection-missing: AnsibleModule.argument_spec.interface.options.type: If removed_in_version or removed_at_date is specified, removed_from_collection must be specified as well for dictionary value @ data['argument_spec']['interface']['options']['type']. Got {'type': 'int', 'default': 0, 'removed_in_version': '2.14'}
- [X] ERROR: plugins/modules/zabbix_proxy.py:0:0: invalid-deprecated-version: Argument 'main' in argument_spec found in interface has an invalid removed_in_version number '2.14': invalid semantic version '2.14'
- [X] ERROR: plugins/modules/zabbix_proxy.py:0:0: invalid-deprecated-version: Argument 'type' in argument_spec found in interface has an invalid removed_in_version number '2.14': invalid semantic version '2.14'

**2.11**
- [X] Copy 2.10 ignore list

**General**
- [x] junit_xml missing. Check we have `--requirements` specified
- [x] For integration tests, which OS?
- [X] sanity tests Ansible-base 2.10
- [X] sanity tests Ansible-base devel (2.11)
[...]

Issue #4, opened 2020-03-14
about the integration test method


sky-joker:
##### PURPOSE

Here we discuss the integration test method for the Zabbix module.  
It provisioning for test environment uses docker-compose? or `ansible-test integration` command? (and so on)

D3DeFi:
I believe that `ansible-test` is not an option. If I understood it right, it creates isolated containers with a base OS in a non routable network. The other thing is that the CI will be unnecessary slow if we are to set up MySQL and Zabbix services ourselves.

My opinion here is that we will benefit greatly from a pre-built Zabbix docker containers, but just a minimal setup possible with zabbix-server + mysql containers and maybe zabbix-web if that is required for a working API. 

The final goal is to have a single script (or rather playbook) that will be run from a Github Actions and can be run locally by anyone without installing a ton of dependencies. Pseudo code can go like so:

```bash
docker-compose
for plugin in plugins/modules/; test plugin
cleanup
```

We can probably reuse your [container provisioner](https://github.com/sky-joker/ansible-zabbix-containers-provisioner/blob/master/zabbix-servers-compose.yml) for this purpose. What do you think?

gundalow:
Have a look at how Grafana integration tests work
They use GitHub actions to spin up the Docker container for the three currently supported Grafana versions: https://github.com/ansible-collections/grafana/blob/master/.github/workflows/ansible-test.yml#L55-L59

Note that when using this you can't use the `--docker` option to ansible-test as you will end up with the Zabbix container and ansible-test container in isolated networks: https://github.com/ansible-collections/grafana/blob/master/.github/workflows/ansible-test.yml#L80

sky-joker:
Thank you @gundalow for infomation!  
To run Zabbix containers, use official Zabbix images and SQL images.  
So I think it would be better to use docker-compose to solve container dependencies easily.  
If you do CI like Grafana, I think we need to create a single container image.  

@D3DeFi

> My opinion here is that we will benefit greatly from a pre-built Zabbix docker containers, but just a minimal setup possible with zabbix-server + mysql containers and maybe zabbix-web if that is required for a working API.

I think so too.  
I tried making a sample for CI.

https://github.com/sky-joker/community.zabbix-sample1-ci  

**Sample docker-compose**

https://github.com/sky-joker/community.zabbix-sample1-ci/blob/master/docker-compose.yml

**Sample Workflow**

[...]

PR #110, opened 2020-06-24
Refactor GitHub action

gundalow:
##### SUMMARY

Follow on from https://github.com/ansible-collections/community.zabbix/pull/105

My aim in refactoring this is we can use it to replace https://github.com/ansible-collections/collection_template/tree/master/.github/workflows which is how new Collection repos will be created
- [X] Full use of `matrix` to make it easier to update in the future
- [X] Move to `actions/checkout@v2`
- [x] Review `on` `push` and `pull_request`, not sure why we have both
- [x] Zabbix team to decide which mix of Python/Zabbix/Ansible version to test for integration
- [x] Add more comments
- [ ] Add `/rebase` workflow

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -50,7 +50,7 @@ jobs:
         working-directory: ./ansible_collections/community/zabbix
 
   integration:
-    runs-on: ubuntu-latestDISABLE
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
@@ -63,36 +63,46 @@ jobs:
             port: "8080"
           - version: "5.0"
             port: "8080"
+        ansible:
+          # - stable-2.9 # Only if your collection supports Ansible 2.9
+          - stable-2.10
+          #- devel
+        python:
+          - 2.7
+          - 3.7
+          #- 3.8 blocked by https://github.com/ansible/ansible/pull/70155
 
     steps:
       - name: Check out code
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
           path: ansible_collections/community/zabbix
 
-      - name: Set up Python 3.6
+      - name: Set up Python ${{ matrix.ansible }}
         uses: actions/setup-python@v1
         with:
-          python-version: 3.6
+          python-version: ${{ matrix.python }}
+
+      - name: Install ansible-base (${{ matrix.ansible }})
+        run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
 
       - name: Install dependencies
         run: pip install docker-compose zabbix-api
 
-      - name: Install ansible-base (stable-2.10)
-        run: pip install https://github.com/ansible/ansible/archive/stable-2.10.tar.gz --disable-pip-version-check
-
       - name: Zabbix container server provisioning
         run: docker-compose up -d
+        working-directory: ./ansible_collections/community/zabbix
         env:
           zabbix_version: ${{ matrix.zabbix_container.version }}
           zabbix_port: ${{ matrix.zabbix_container.port }}
 
       - name: Run integration test
-        run: ansible-test integration -v --color --retry-on-error --continue-on-error --diff --requirements --coverage
+        run: ansible-test integration -v --color --retry-on-error --continue-on-error --diff --python ${{ matrix.python }} --requirements --coverage
+        working-directory: ./ansible_collections/community/zabbix
 
       - name: Generate coverage report.
         run: ansible-test coverage xml -v --requirements --group-by command --group-by version
-        #working-directory: ./ansible_collections/community/zabbix
+        working-directory: ./ansible_collections/community/zabbix
 
       - uses: codecov/codecov-action@v1
         with:
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -18,7 +18,7 @@ on:
 
 jobs:
   sanity:
-    name: Sanity (${{ matrix.ansible }})
+    name: Sanity (${{ matrix.ansible }}) py${{ matrix.python }}
     strategy:
       matrix:
         ansible:
@@ -45,7 +45,8 @@ jobs:
         run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
 
       - name: Run sanity tests
-        run: ansible-test sanity --docker -v --color --python 3.6
+        run: ansible-test sanity --docker -v --color --python ${{ matrix.python }}
+        working-directory: ./ansible_collections/community/zabbix
 
   integration:
     runs-on: ubuntu-latestDISABLE
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -1,5 +1,6 @@
 name: plugins
 on:
+# FIXME: Is `push` and `pull_request` right
   push:
     paths:
       - 'plugins/**'
@@ -21,20 +22,24 @@ jobs:
     strategy:
       matrix:
         ansible:
-        - stable-2.10
-        - devel
+          # - stable-2.9 # Only if your collection supports Ansible 2.9
+          - stable-2.10
+          - devel
+        python:
+          - 2.7
+          - 3.8
     runs-on: ubuntu-latest
     steps:
 
       - name: Check out code
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
           path: ansible_collections/community/zabbix
 
-      - name: Set up Python 3.6
+      - name: Set up Python ${{ matrix.ansible }}
         uses: actions/setup-python@v1
         with:
-          python-version: 3.6
+          python-version: ${{ matrix.python }}
 
       - name: Install ansible-base (${{ matrix.ansible }})
         run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
@@ -43,7 +48,7 @@ jobs:
         run: ansible-test sanity --docker -v --color --python 3.6
 
   integration:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latestDISABLE
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -115,7 +115,7 @@ jobs:
       # As we need to connect to an existing docker container we can't use `--docker` here as the VMs would be on different
       # (non-routing) networks, so we run them locally and ensure any required dependencies are installed via `--requirements`
       - name: Run integration test
-        run: ansible-test integration -v --color --retry-on-error --continue-on-error --diff --python ${{ matrix.python }} --requirements --coverage --allow-unstable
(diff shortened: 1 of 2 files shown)


webknjaz:
> * [ ]  Review `on` `push` and `pull_request`, not sure why we have both

When you do `git push`, an event happens _in the repo where this is done_. So if anything is pushed to `master` or other branches, or tags, in this repo (this happens when you merge PRs too!), the `push` event appears in the GitHub platform and workflows in this repo will be triggered (given they declare `on: push`).

When there's a PR from a branch inside of this repo, you'll see "double runs" — when the branch gets updated both `push` and `pull_request` events happen in the same repo and they both trigger workflow runs.

When a PR is coming from a forked repo, updates to the PR branch will generate `push` in that forked repo and this upstream one will not "see" it but it will get a `pull_request` event.

All other CIs run builds on both `push` and `pull_request` because they are different events. Also, all CIs (and `actions/checkout@2` too) do this clever thing — when they run on `pull_request` and not normal `push`, they'll try to merge that branch into `master` so that the build would work with that updated version and not the branch `HEAD` as it is, whenever possible.
GH even has a poorly documented ref in their read-only namespace [`+refs/pull/<ID>/merge`](https://www.jvt.me/posts/2019/01/19/git-ref-github-pull-requests/) that holds a "phantom" merge commit if the PR is mergeable.

So my answer to this question would be yes, you need both events there. Unless you're absolutely sure that there's no way code get into `master` w/o getting through a PR review with branch protection. Also, if you want some custom behavior like publishing to Galaxy on tag pushes, you may use a `create` event instead. Oh, and there's also `schedule` that allows you to rerun the workflows on `cron`, you may want to set it up too.

.github/workflows/plugins.yml
```diff
@@ -1,5 +1,6 @@
 name: plugins
 on:
+# FIXME: Is `push` and `pull_request` right
```
webknjaz, line 3: https://github.com/ansible-collections/community.zabbix/pull/110#issuecomment-649454402

.github/workflows/plugins.yml
```diff
@@ -17,30 +18,36 @@ on:
 
 jobs:
   sanity:
-    name: Sanity (${{ matrix.ansible }})
+    name: Sanity (${{ matrix.ansible }}) py${{ matrix.python }}
```
webknjaz, line 21: how about
```suggestion
    name: Sanity (Ⓐ${{ matrix.ansible }}+py${{ matrix.python }})
```

.github/workflows/plugins.yml
```diff
@@ -17,30 +18,36 @@ on:
...
         ansible:
-        - stable-2.10
-        - devel
+          # - stable-2.9 # Only if your collection supports Ansible 2.9
+          - stable-2.10
+          - devel
+        python:
+          - 2.7
+          - 3.7
+          #- 3.8 blocked by https://github.com/ansible/ansible/pull/70155
```
webknjaz, line 31: I think I'd prefer this style, but up to you:
```suggestion
          - 3.8
          exclude:
          - python: 3.8  # blocked by https://github.com/ansible/ansible/pull/70155
```

.github/workflows/plugins.yml
gundalow: Nice, thanks

.github/workflows/plugins.yml
gundalow: Ah, this is much more readable. thans

.github/workflows/plugins.yml
```diff
@@ -17,30 +18,36 @@ on:
...
 
       - name: Install ansible-base (${{ matrix.ansible }})
         run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
 
       - name: Run sanity tests
-        run: ansible-test sanity --docker -v --color --python 3.6
+        run: ansible-test sanity --docker -v --color --python ${{ matrix.python }}
+        working-directory: ./ansible_collections/community/zabbix
 
   integration:
```
webknjaz, line 54: You may want to also improve the integration job names. Currently they don't fit in the UI here: https://github.com/ansible-collections/community.zabbix/runs/805153167?check_suite_focus=true.

.github/workflows/plugins.yml
```diff
@@ -56,36 +65,46 @@ jobs:
...
     steps:
       - name: Check out code
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
           path: ansible_collections/community/zabbix
 
-      - name: Set up Python 3.6
+      - name: Set up Python ${{ matrix.ansible }}
         uses: actions/setup-python@v1
```
webknjaz, line 84:
```suggestion
        uses: actions/setup-python@v2
```

.github/workflows/plugins.yml
```diff
@@ -56,36 +65,46 @@ jobs:
...
       - name: Zabbix container server provisioning
         run: docker-compose up -d
+        working-directory: ./ansible_collections/community/zabbix
         env:
           zabbix_version: ${{ matrix.zabbix_container.version }}
           zabbix_port: ${{ matrix.zabbix_container.port }}
 
       - name: Run integration test
-        run: ansible-test integration -v --color --retry-on-error --continue-on-error --diff --requirements --coverage
+        run: ansible-test integration -v --color --retry-on-error --continue-on-error --diff --python ${{ matrix.python }} --requirements --coverage
```
webknjaz, line 102: Any reason to run sanity in docker but this one in the VM directly?

.github/workflows/plugins.yml
```diff
@@ -1,46 +1,55 @@
...
-        - devel
+          # - stable-2.9 # Only if your collection supports Ansible 2.9
+          - stable-2.10
+          - devel
+        python:
+          - 2.7
+          - 3.7
+          - 3.8
+          exclude:
+          - python: 3.8  # blocked by https://github.com/ansible/ansible/pull/70155
```
webknjaz, line 33: Looks like I suggested a poorly indented block before. This should fix it:
```suggestion
        exclude:
        - python: 3.8  # blocked by https://github.com/ansible/ansible/pull/70155
```

.github/workflows/plugins.yml
D3DeFi: Hi @webknjaz and yes, we run integration tests against multiple versions of Zabbix servers (haven't found prettier solution via Github Actions than "stack"s being deployed by [docker-compose](https://github.com/ansible-collections/community.zabbix/blob/master/docker-compose.yml)) as we are not interested in testing on any major OS distro since we should be OS independent.

Getting each component of Zabbix working together was a real pain and this was far simpler solution since we are use this easily for local development too.

webknjaz: I meant the missing `--docker` option here.

D3DeFi: my bad for misinterpreting. Result would be non-routable network between test container and zabbix stack if I recall correctly

[See @gundalow comment](https://github.com/ansible-collections/community.zabbix/issues/4#issuecomment-599468228)

webknjaz: Ah, I see. I wonder if using GHA native services would work: https://help.github.com/en/actions/configuring-and-managing-workflows/about-service-containers#mapping-docker-host-and-service-container-ports

D3DeFi: Heh, I've already tried that in the same issue, although I don't recall much about it now - [comment](https://github.com/ansible-collections/community.zabbix/issues/4#issuecomment-602087010)

.github/workflows/plugins.yml
gundalow: TIL `actions/setup-python@v2`, thanks :)

.github/workflows/plugins.yml
webknjaz: @gundalow this causes YAML parsing error ^

gundalow:
1) Hitting some intermittent issue with `test_zabbix_maintenance` does that need disabling?
2) For the integration tests, what mix of Ports, Python_version, Zabbix_version, ansible_version would you like?]
I'm guessing we only need to test one portnumber vs ansible_version

gundalow:
I think this is good to merge now.
Thanks to @webknjaz @D3DeFi for their help with this.

We may trim down the text matrix in a future PR once we've got more feedback.

.github/workflows/plugins.yml
```diff
@@ -1,49 +1,70 @@
...
 on:
+# Run CI against all pushes (direct commits) and Pull Requests
   push:
-    paths:
-      - 'plugins/**'
-      - 'tests/**'
-      - 'scripts/**'
-      - '.github/workflows/plugins.yml'
-      - 'docker-compose.yml'
+#    paths:
```
D3DeFi, line 5: What about these? We preciously added them so integration tests are not run everytime a change is made in the roles. But we may change the logic to ignore_paths

.github/workflows/plugins.yml
```diff
@@ -56,37 +77,55 @@ jobs:
...
           - version: "5.0"
             port: "8080"
+        ansible:
+          # - stable-2.9 # Only if your collection supports Ansible 2.9
+          - stable-2.10
+          - devel
+        python:
+          - 2.7
+          - 3.7
+          #- 3.8 blocked by https://github.com/ansible/ansible/pull/70155
```
D3DeFi, line 87: exclude:
          - python: 3.8  # blocked by https://github.com/ansible/ansible/pull/70155

Instead?

.github/workflows/plugins.yml
```diff
@@ -1,49 +1,70 @@
...
       - name: Run sanity tests
-        run: ansible-test sanity --docker -v --color --python 3.6
+        run: ansible-test sanity --docker -v --color --python ${{ matrix.python }}
+        working-directory: ./ansible_collections/community/zabbix
+
+# FIXME IN TEMPLATE: UNIT TESTS WOULD GO HERE
 
   integration:
     runs-on: ubuntu-latest
+    name: I (${{ matrix.zabbix_container.version}}:${{ matrix.zabbix_container.port }} Ⓐ${{ matrix.ansible }}+py${{ matrix.python }}})
```
D3DeFi, line 67: Port can go away altogether, no need to have it in the name

.github/workflows/plugins.yml
webknjaz: How about keeping them just for PRs?

gundalow: @D3DeFi Good call on `ignore_paths`. I hope at some point we will have `ansible-test` validate galaxy.yml etc, so this feels safer.

Maybe in the future we might see some Molecule testing for these `roles` https://github.com/ansible-collections/community.kubernetes/blob/master/.github/workflows/ci.yml#L67-L99 ...

.github/workflows/plugins.yml
gundalow: @D3DeFi Thank you, that looks a lot better.

gundalow:
@webknjaz @D3DeFi Thank you for your help with this. I hope this is good.

If you can approve I'll merge, then I'll update https://github.com/ansible-collections/collection_template based on this (though adding a `ansible-galaxy collection install`, adding units)

.github/workflows/plugins.yml
webknjaz: @D3DeFi it is now possible to use `--docker` if you synchronize with the newest collection template workflow @ https://github.com/ansible-collections/collection_template/blob/main/.github/workflows/ansible-test.yml that uses a unified action solving this problem.

PR #113, opened 2020-06-25
Integration tests for zabbix_action module with few adjusted fixes

D3DeFi:
…bix_action module

##### SUMMARY
- adds integration tests for `zabbix_action` module
- `remote_command` of type SSH always required all of the options (username, pass, ssh_*_file and so on) regardless of auth method set in `ssh_auth_type`
- removed `selectInventory: extend` from API calls where it is silently ignored (or fails like with `zbx.script.get`)
- documentation adjustements
- it was unclear what options are available for `inventory` parameter in `*operations` so unified that to `manual` and `automatic`
- Edit: I've decided to skip 3.0 testing completly since the module states it only supports 3.4+ version and 3.4 is already unsupported, 4.0 is covered by these tests

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/zabbix_action
tests/integration/targets/test_zabbix_action

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,17 @@
 
 ### Improvements
 #### Modules:
-  - zabbix_action - fixed error on changed API fields for Zabbix 5.0 [GitHub Issue](https://github.com/rockaut/community.zabbix/edit/fix_92)
+  - zabbix_action - fixed error on changed API fields `*default_message` and `*default_subject` for Zabbix 5.0 (PR [#92](https://github.com/ansible-collections/community.zabbix/pull/92)).
+  - zabbix_action - parameter `ssh_auth_type` for SSH `remote_command` operation now correctly identifies which other parameters are required (PR [#113](https://github.com/ansible-collections/community.zabbix/pull/113)).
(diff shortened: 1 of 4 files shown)


D3DeFi:
Now this Codecov doesn't make sense at all to me. I would say it is comparing my branch against something else than `master` as this PR is not touching other modules at all.

gundalow:
Could you please provide some more details? We haven't been able to use codecov like this before as the main ansible repo is too big.

D3DeFi:
> Could you please provide some more details? We haven't been able to use codecov like this before as the main ansible repo is too big.

Yes, sorry @gundalow. What I meant by that is (as I am still grasping codecov) is that this one specific coverage report doesn't make sense to me, because I am not touching anything related to lets say `zabbix_host` module, yet it says that I will reduce its coverage with this patch?

D3DeFi:
@sky-joker can you look at this please? I don't plan to enhance it further for now, so merge as it is if you agree and ignore codecov. There is too much other work pending for me to play with this particular PR

sky-joker:
>  I don't plan to enhance it further for now, so merge as it is if you agree and ignore codecov. 

okay.
I'll ignore the `codecov` result.

LGTM :)
Thank you @D3DeFi for this patch!

D3DeFi:
Thank you for looking into it @sky-joker  :)

PR #115, opened 2020-06-26
Fixed getenforce binary path

offmindby:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_agent/tasks/Linux.yml
##### ADDITIONAL INFORMATION

Running by default role code is checking for getenforce binary using full path "/usr/sbin/getnforce". But, in case the binary exists Ansible tries to launch command using binary basename 'getenforce'.
So, it's better use full path.

```paste below

```


dj-wasabi:
Thanks! 👍

Issue #112, opened 2020-06-25
A zabbix_service module has no idempotency with Zabbix 5.0


sky-joker:
##### SUMMARY

The `changed` occur when executing the task of creating a new service twice.

##### ISSUE TYPE

- Bug Report

##### COMPONENT NAME

modules/zabbix_service.py

##### ANSIBLE VERSION

```
# ansible --version
ansible 2.11.0.dev0
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/collection.zabbix/PR/xxx/ansible/lib/ansible
  executable location = /root/collection.zabbix/PR/xxx/ansible/bin/ansible
  python version = 3.6.8 (default, Aug  7 2019, 17:28:10) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
```

##### CONFIGURATION

```
# ansible-config dump --only-changed
```

##### OS / ENVIRONMENT / Zabbix Version

* Zabbix Version 5.0

##### STEPS TO REPRODUCE

Execute the following playbook twice.

```yaml
- name: example
  hosts: localhost
  gather_facts: no
  vars:
    zabbix_server_url: http://127.0.0.1:8080
    zabbix_login_user: Admin
    zabbix_login_password: zabbix
  tasks:
    - name: "test - Create a new service"
      zabbix_service:
        server_url: "{{ zabbix_server_url }}"
        login_user: "{{ zabbix_login_user }}"
        login_password: "{{ zabbix_login_password }}"
        name: ExampleServiceForServiceModule
        sla: 99.99
        state: present
      register: create_service_idempotency_check_result
```

##### EXPECTED RESULTS

No `changed` occurs the second time.

##### ACTUAL RESULTS

The `changed` occurs twice in a row.

PR #116, opened 2020-06-28
Fix bug for zabbix_service has no idempotency and add integration tests

sky-joker:
This PR is fixes bug for zabbix_service has no idempotency with Zabbix 5.0.  
The cause is zero-padding was gone from `goodsla` in the got results from Zabbix 5.0.

**When Zabbix 4.4 or less**

```
            "goodsla": "99.9990",
```

**When Zabbix 5.0**

```
            "goodsla": "99.999",
```

So, fix to be in the same format when comparing.
And add integration tests for the zabbix_service module.

fixes: https://github.com/ansible-collections/community.zabbix/issues/112

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

modules/zabbix_service.py

##### ADDITIONAL INFORMATION

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Bug Fixes:
 #### Modules:
   - zabbix_action - now correctly selects mediatype for the (normal|recovery|update) operations with Zabbix 4.4 and newer. (PR [#90](https://github.com/ansible-collections/community.zabbix/pull/90))
+  - zabbix_service - fixed the zabbix_service has no idempotency with Zabbix 5.0 (PR [#116](https://github.com/ansible-collections/community.zabbix/pull/116))
 
 ## 0.2.0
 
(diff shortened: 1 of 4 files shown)


sky-joker:
I wonder, do these results make sense?
https://codecov.io/gh/ansible-collections/community.zabbix/compare/f96dab7346defd33f3a91bfa45ad07cd5800863d...729465962be46c360d9eda45c4f17db8a2049a26

D3DeFi:
> I wonder, do these results make sense?
> https://codecov.io/gh/ansible-collections/community.zabbix/compare/f96dab7346defd33f3a91bfa45ad07cd5800863d...729465962be46c360d9eda45c4f17db8a2049a26

I presume you are referring to a different state of codecov as is now shown. I was presented the same behavior before and also thought that it doesn't make any sense.

The thing is, that coverage is based on success of the tests. So if for example your changes fail on zabbix 4.4 and some of the modules will not get tested because of that, it reports that the coverage will be reduced. As you can see now, it makes much more sense since you force pushed.

sky-joker:
Thank you @D3DeFi

Issue #104, opened 2020-06-22
[enhancement][module] Implementing some module_utils for refactoring/centralizing things


rockaut:
##### SUMMARY
We've talked about this now quite often in other issues so I thought an issue might help to get it centralized or linked better.
It may be a benefit to have some centralized and reusable code to the the modules more streamlined and maybe even ease the maintainability of the code overall.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
all modules

##### ADDITIONAL INFORMATION
I've already played around with this a little bit - I will upload my code over at my forked repo so you first might have a look at it.
Basically I just thought about getting `plugins/module_utils/zapiwrapper.py` and `plugins/module_utils/zabbixbase.py` with each a class to derive of for now. Grafana is a good example and some other collections make it like this too.

This way the transition to the inherited classes would be easily manageable as the individual module classes can inherit the base classes but also do their own functions still and nothing would get compromised at all. Over time more code could get centralized. Another file like `plugins/module_utils/helper.py` could handle the various function currently in every module.

Imo, this approach might also ease the transition over from the old modules and to 5.0 wdyt?

rockaut:
Also this way one could create his own collection copy over one or more modules or module_utils and "redirect" things through altering `import` statements and with proper FQCN usage. With this, quick-fixes for specialized environments or bugs are good manageable even in enterprise environments - heck, one could even use multiple versions at one if wanted.

Thought this might then be even worth documenting or something for wiki?

rockaut:
Pushed to my fork under https://github.com/rockaut/community.zabbix/tree/refactor_test
Somewhat useless for now but should be a starting point.

For example, changing in zabbix_action.py:

```python

class Zapi(ZapiWrapper):
    """
    A simple wrapper over the Zabbix API
    """
    # removed __init__ here

class Action(ZabbixBase):
    """
    Restructures the user defined action data to fit the Zabbix API requirements
    """
    # removed __init__ here

    def _construct_parameters(self, **kwargs):
```

D3DeFi:
looks good to me and can be easily extended

> Thought this might then be even worth documenting or something for wiki?

Small steps at a time I would say :)

D3DeFi:
Do we keep this open since #107 has been merged already. Will you link followup PRs to this one as well?

rockaut:
Please keep it open for now until we have added the basic function we like to add.
Also I think it might work as a gatekeeper issue to link the other refactors coming.

[...]

PR #120, opened 2020-07-02
refactoring zabbix_discovery_rule module

tcraxs:
##### SUMMARY
refactoring zabbix_discovery_rule module.
* use module_utils classes and functions
* add documentation for return block
* adjust return values on success

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
* zabbix_discovery_rule

##### ADDITIONAL INFORMATION
* This PR uses same changes in `helpers.py` from #119
* Have to rebase one of the PRs after one was merged, but need the changes also here to get working tests

PR #123, opened 2020-07-05
refactor zabbix_group and zabbix_group_info

rockaut:
##### SUMMARY
Refactor zabbix_group to use module_utils.
See #104 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_group
zabbix_group_info

### DIFF
--- a/plugins/modules/zabbix_group.py
+++ b/plugins/modules/zabbix_group.py
@@ -41,7 +41,6 @@ options:
 extends_documentation_fragment:
 - community.zabbix.zabbix
 
-
 notes:
     - Too many concurrent updates to the same group may cause Zabbix to return errors, see examples for a workaround if needed.
 '''
--- a/plugins/modules/zabbix_group.py
+++ b/plugins/modules/zabbix_group.py
@@ -87,12 +87,11 @@ except ImportError:
 
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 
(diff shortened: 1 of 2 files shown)


D3DeFi:
> LGTM, but merging after #120, then this needs to be rebased

#120 has been merged. Rebase when you have a chance please and than this can be merged :)

Thank you for the contribution @rockaut and good job

D3DeFi:
actually whole `HAS_ZABBIX_API` is no longer needed as well. So try/except block for zabbix_api import can be removed from modules. Same for `missing_required_lib` import

rockaut:
Rebased and @D3DeFi I removed it from zabbix_group_info as it wasn't needed there. But in zabbix_group  there is still a `Already_Exists` exeption from the api so we need it.

I thought about movin the HostGroup Class to module_utils/base.py or an own file/class under modue_utils but I don't know if this would fit. That would be a step further which needs more discussion because it affects the other modules too.

Currently I think it's good.

D3DeFi:
> I thought about movin the HostGroup Class to module_utils/base.py or an own file/class under modue_utils but I don't know if this would fit. That would be a step further which needs more discussion because it affects the other modules too.


I’ve hit the same idea when working on the zabbix_host rework as lots of this stuff is really duplicated across the collection. For example - proxy, groups, host interfaces.

Either to dedicated module_utils files or, if it is even possible and not against some design principles, import from modules itself. Like “from plugins.modules.zabbix_group import ZabbixGroup”

PR #127, opened 2020-07-09
WIP: implement zabbix_credentials

### LINKED ISSUE #126: Implement a connection dictionary on modules
##### SUMMARY
Currently we need to write the common connection `server_url, login_user, login_password` on each tasks in plays. That's repeating. The AVI modules for example has a neat feature for this [see Ansible Docs](https://docs.ansible.com/ansible/latest/modules/avi_network_module.html#parameter-avi_credentials) / [github avi modules](https://github.com/ansible/ansible/tree/stable-2.9/lib/ansible/modules/network/avi) .

One the one hand they provide the parameters on each module like we do. But they also have a parameter `avi_credentials` which you can pass an dictionary with the connection parameters. So you just call this one parameter in the tasks and "store" a dictionary with the other parameters in the inventory. I like this - will do a PR.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
zabbix_connection or zabbix_credentials for each module

##### ADDITIONAL INFORMATION
Additional links to avi for examples:
[avi commit on credentials](https://github.com/ansible/ansible/commit/6d21cebd7d8b5ecc30c2d01d0eeb6e44d2d5fa06)
[avi helpers ](https://github.com/ansible/ansible/blob/0b8354751b3ecbdba1813030411701af1895bdfa/lib/ansible/module_utils/network/avi/avi_api.py)

rockaut:
##### SUMMARY
See #126 for more information
Fixes #126

##### ISSUE TYPE
- Feature Pull Request

##### ADDITIONAL INFORMATION
With this implementation you could for example:

```yaml
---

- hosts: localhost
  gather_facts: false
  collections:
    - community.zabbix
  vars:
    zabbix_creds:
      server_url: http://localhost:8080
      login_user: Admin
      login_password: zabbix
  tasks:
    - name: Create host groups - old style
      local_action:
        module: zabbix_group
        server_url: http://localhost:8080
        login_user: Admin
        login_password: zabbix
        state: "{{ state | default('present') }}"
        host_groups:
          - _JUSTATEST_

    - name: Create host groups - new style
      local_action:
        module: zabbix_group
        zabbix_credentials: "{{ zabbix_creds }}"
        server_url: http://someother:8080 # <- overrides zabbix_creds.server_url
        state: "{{ state | default('present') }}"
        host_groups:
          - _JUSTATEST2_

```

### DIFF
--- a/plugins/doc_fragments/zabbix.py
+++ b/plugins/doc_fragments/zabbix.py
@@ -9,41 +9,96 @@ class ModuleDocFragment(object):
     # Standard documentation fragment
     DOCUMENTATION = r'''
 options:
+    zabbix_credentials:
+        description:
+            - Zabbix connection details. A dictionary which can be used to store the connection details.
+            - direct module parameters overriding zabbix_credential parameters
+        suboptions:
+            server_url:
+                description:
+                    - URL of Zabbix server, with protocol (http or https).
+                required: true
+                type: str
+            login_user:
+                description:
+                    - Zabbix user name.
+                required: true
+                type: str
+            login_password:
+                description:
+                    - Zabbix user password.
+                required: true
+                type: str
+            http_login_user:
+                description:
+                    - Basic Auth login
+                type: str
+            http_login_password:
+                description:
+                    - Basic Auth password
+                type: str
+            timeout:
+                description:
+                    - The timeout of API request (seconds).
+                type: int
+                default: 10
+            validate_certs:
+                description:
+                    - If set to False, SSL certificates will not be validated. This should only be used on personally controlled sites using self-signed certificates.
+                type: bool
+                default: yes
+        type: dict
+        version_added: "0.3"
+        aliases: [ connection, zabbix_connection, credentials ]
     server_url:
         description:
             - URL of Zabbix server, with protocol (http or https).
               C(url) is an alias for C(server_url).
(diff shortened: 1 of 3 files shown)


rockaut:
The sanity tests are failing because they test the zabbix_withcreds.py which has for example parameter server_url required=False now. I thought we just need to change the included doc_fragment after we changed a module to use the new parameters.

What should I do here?

D3DeFi:
There is no way to tell documentation that X is required if Y is not specified and vice versa. You by default only specify it in the description of each option, then set both to `required=False` and do:

```python
required_one_of=[
    [ 'server_url', 'connection' ]
]
required_if=[
    ['server_url', True, ['login_password', 'login_user']]
]
```

src: [some blog](https://mobygeek.net/blog/2016/02/16/ansible-module-development-parameters/)

I don't know how you would do this in a separate function though. Normally these blocks are specified as a parameters to `AnsibleModule`, so alongside `argument_spec` and `supports_check_mode`. ref: [zabbix_action](https://github.com/ansible-collections/community.zabbix/blob/master/plugins/modules/zabbix_action.py)

rockaut:
Ok. I will remove the doc_fragment for now and incorporate it after all modules have it.
Also I will do required_if things.

rockaut:
Dropped for now see #126

PR #142, opened 2020-07-19
refactor screen

rockaut:
##### SUMMARY
implement module_utils
see #104

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_screen

### DIFF
--- a/plugins/modules/zabbix_screen.py
+++ b/plugins/modules/zabbix_screen.py
@@ -287,7 +287,7 @@ class Screen(ZabbixBase):
         if len(screen_item_list) > 0:
             if self._module.check_mode:
                 self._module.exit_json(changed=True)
-            ScreenItem.delete(self._zapi, screen_item_id_list)
+            ScreenItem.delete(self, screen_item_id_list)
             return True
         return False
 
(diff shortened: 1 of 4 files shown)


plugins/modules/zabbix_screen.py
```diff
@@ -355,65 +343,43 @@ def create_screen_items(self, screen_id, hosts, graph_name_list, width, height,
                                                           'colspan': 1, 'rowspan': 1,
                                                           'elements': 0, 'valign': 0, 'halign': 0,
                                                           'style': 0, 'dynamic': 0, 'sort_triggers': 0})
-        except Already_Exists:
-            pass
+        except Exception as ex:
+            if ex.__class__.__name__ == "Already_Exists":
+                pass
+            else:
+                raise ex
```
D3DeFi, line 350: This is not very fortunate approach I would say. But on the other hand, this whole module could use som more love in the future.

plugins/modules/zabbix_screen.py
```diff
@@ -302,7 +290,7 @@ def delete_screen_items(self, screen_id, screen_item_id_list):
                 self._zapi.screenitem.delete(screen_item_id_list)
                 return True
             return False
-        except ZabbixAPIException:
+        except Exception:
```
D3DeFi, line 293: This may be against "best practices"

plugins/modules/zabbix_screen.py
rockaut: I know ... I don't like it too much either. But this way we could dump the imports for zabbix-api.
After this I would look into how we can use the Wrapper class to get around this nasty.

plugins/modules/zabbix_screen.py
rockaut: You're right. I don't like the two approaches either. I'm still on vacation, keep it open I will see how I can get this into the Wrapper class or a helper.

D3DeFi: No worries, no rush! Thanks for all of your work tho :)

How bad of an idea is to just import into `module_utils.wrapper`, not use it, but then call it from modules like so:
`from ansible_collections.community.zabbix.plugins.module_utils.base import ZabbixBase, Already_Exists, ZabbixAPIException`?

But this will probably complain in linting or tox (since we will have that as well).

sky-joker:
I wonder if the reason there are so many change files is that they're not rebased well?

rockaut:
@sky-joker 

> I wonder if the reason there are so many change files is that commits are not rebased well?

You might be right. I rebased and squashed it anyways with latest push.

plugins/modules/zabbix_screen.py
rockaut: I've now implemented a class with static functions to wrap around the Already_Exists exception.
On the other hand I've still imported ZabbixAPIException from wrappers.py as you suggested... but to be true I don't understand why we `pass` the exception and don't exit with an error if the ScreenItem can't be deleted?

D3DeFi: I am not very familiar with this modules to honest and I haven't thought about it when reviewing as only thing I've considered was a diff itself. If you are up for it, this can be changed. Raising exceptions in this context is stupid anyways. This is one of the modules that would really need rework

rockaut:
ready

rockaut:
Didn't realize we don't have tests! Will write some and include it here.

rockaut:
Well... I have to look into why it's failing for some versions but not others now. *sigh*
I will find some time for it over the weekend.

D3DeFi:
@rockaut seems that versions <= 4.0 have `Zabbix server` host set to `disabled` and it doesn't match this select:
```python
host_list = self._zapi.host.get({'output': 'extend', 'selectGroups': 'groupid', 'groupids': group_ids, 'monitored_hosts': 1})
```

Also it seems that modules simply skips any non-existing graphs. Since `Zabbix internal queues` was added later, it had no changes to do.

Amazing job as always!

PR #150, opened 2020-07-26
Add Zabbix Server to ci

dj-wasabi:
##### SUMMARY

Another attempt to add the Zabbix Server to the CI process.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     ignore_paths:
       - 'roles/**'
+      - 'molecule/**'
 
 jobs:
   tox-linters:
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -1,0 +1,69 @@
+name: "community.zabbix.zabbix_server"
+on:
+  push:
+    paths:
+      - 'roles/zabbix_server/**'
+      - 'molecule/zabbix_server/**'
+      - '.github/workflows/server.yml'
+  pull_request:
+    paths:
+      - 'roles/zabbix_server/**'
+      - 'molecule/zabbix_server/**'
+      - '.github/workflows/server.yml'
+jobs:
+  molecule:
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        molecule_distro:
+          - container: mysql-centos
+            image: geerlingguy/docker-centos8-ansible:latest
+            group: mysql
+          - container: pgsql-centos
+            image: geerlingguy/docker-centos8-ansible:latest
+            group: postgresql
+          - container: mysql-ubuntu
+            image: geerlingguy/docker-ubuntu2004-ansible
+            group: mysql
+          - container: pgsql-ubuntu
+            image: geerlingguy/docker-ubuntu2004-ansible
+            group: postgresql
+            command: /sbin/init
+          - container: mysql-debian
+            image: geerlingguy/docker-debian9-ansible
+            group: mysql
+          - container: pgsql-debian
+            image: geerlingguy/docker-debian9-ansible
+            group: postgresql
+        collection_role:
+          - zabbix_server
+    steps:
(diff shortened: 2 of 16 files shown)


D3DeFi:
can you please push `- molecule/` under `ignore_paths` in .github/workflows/plugins.yml with this patch pls?

dj-wasabi:
> can you please push `- molecule/` under `ignore_paths` in .github/workflows/plugins.yml with this patch pls?

Yes, done that. Thank you! 👍

D3DeFi:
> > can you please push `- molecule/` under `ignore_paths` in .github/workflows/plugins.yml with this patch pls?
> 
> Yes, done that. Thank you! +1

oh but it still hits plugins CI due to docs/ and .github/workflow/server.yml. Roles should not care about result of those at all :( Not sure if adding more `ignore_paths` entries is a good idea.

Back in #110 it was introduced to .github/workflow/plugins.yml instead of explicit list of directories to test in `paths`. I believe that is some kind of ansible requirement that sanity tests are passing for the whole repo.

What if we split plugins to 2 files `plugins-sanity.yml`, `plugins-integrations.yml`? @sky-joker what do you think?

D3DeFi:
But I am really happy that you narrowed down the tests and they are working now! Awesome work @dj-wasabi ! :rocket:

dj-wasabi:
> But I am really happy that you narrowed down the tests and they are working now! Awesome work @dj-wasabi ! 🚀

Yes, i just stopped using the Ansible roles for both PosgreSQL and MySQL and using Docker images. They always work, and thus also as part of the CI. Once this PR is merged, adding the Proxy isn't much work but do have some issues with the Web..  but will solve that one as well.. 😄

sky-joker:
> What if we split plugins to 2 files plugins-sanity.yml, plugins-integrations.yml? @sky-joker what do you think?

How about limiting CI of this repo to running as follows?

```yaml
name: plugins
on:
# Run CI against all pushes (direct commits) and Pull Requests
# push: ensure we always run tests
# pull_requests: skip if PR only modifies "./roles/"
  push:
  pull_request:
    push:
      paths:
        - tests/**
        - plugins/modules/**
    #ignore_paths:
    #  - 'roles/**'
    #  - 'molecule/**'
(snip)
```

sky-joker:
I thought the molecule scenario was under the role directory, is it correct that it is at the top of the repository?

dj-wasabi:
> I thought the molecule scenario(or directory) was under the role directory, is it correct that it is at the top of the repository?

Yes correct. It is now in the <ROOT_DIR>. Other collections has the `molecule` present in the <ROOT_DIR>, so why not this one as well.

D3DeFi:
> For example, How about limiting CI of this repo to running as follows?

It was like that before and got changed in:
https://github.com/ansible-collections/community.zabbix/pull/105

I believe it was changed because Ansible require us to run sanity tests on whole collection. This is the reason why I consider splitting plugins CI file. It would allow us to have integration tests locked only on specific directories and run sanity on whole repo.

@gundalow can you please confirm that the change in #105 from paths to ignore_paths was intended?

D3DeFi:
Also I think that molecule will not work in any other directory. We had this problem with the previous try to start CI

sky-joker:
I believe that both sanity tests and integration tests are required if the module is modified.  
If only integration tests were changed, I think you should just run integration tests only, but I don't know if there is a particular need to separate the files.  
I'd like to check out @gundalow intentions.

gundalow:
ansible-test don't currently check anything for roles (apart from maybe yanllint), so feel free to limit ansible-test sanity/unit/integration to the non-roles directories.

dj-wasabi:
> I checked to work all tests of this CI even in my test environment :)
> Thank you @dj-wasabi

Nice! Thanks! :)

I am just glad to that there is something working. Continue with the rest! 👍

PR #164, opened 2020-07-29
DB TLS variables and StartPreprocessors

b2ronn:
There are no following variables in the current roles in templates for Zabbix server/proxy:
DBTLSconnect
DBTLScafile
DBTLScertfile
DBTLSkeyfile
DBTLScipher
DBTLScipher13
StartPreprocessors

and added the zabbix_repo_yum_schema variable so that the web schema can be selected.

### DIFF
--- a/docs/ZABBIX_AGENT_ROLE.md
+++ b/docs/ZABBIX_AGENT_ROLE.md
@@ -315,6 +315,8 @@ Keep in mind that using the Zabbix Agent in a Container requires changes to the 
 
 * `zabbix_agent_inventory_zabbix`: Adds Facts for a zabbix inventory
 
+* `zabbix_repo_yum_schema`: Option to change the web schema for the yum repository(http/https)
+
 ## IPMI variables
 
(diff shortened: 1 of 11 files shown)


dj-wasabi:
Hi @b2ronn 

Thank you! Can you also update the value on https://github.com/ansible-collections/community.zabbix/blob/b1fbb7947501eb98ace5de55c4c9511dd41bd782/roles/zabbix_web/defaults/main.yml#L48 ? 

And please add something into the `docs/*.md` about the property `zabbix_repo_yum_schema`, so people can know about it but also they can change it to the other value. Maybe we should but it default with `https` instead of `http`, so it will be "secure out-of-the-box".. 😄

dj-wasabi:
Hi @b2ronn 

Can you do a rebase from master? We merged a PR that contains some tests for the Zabbix Server role, if all is right we should see that this PR passes the tests. Thanks! 👍

b2ronn:
@dj-wasabi merged

dj-wasabi:
Thanks @b2ronn ! 👍 
Looks good!

PR #166, opened 2020-07-30
Add Zabbix Proxy to CI

dj-wasabi:
This PR contains the following changes:

* Removed old molecule dir for Zabbix (I thought that I had done that with previous PR);
* Added Molecule for the Zabbix Proxy;
* Copied logic for both MySQL as PgSQL from Server to Proxy, as they are almost the same;

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


dj-wasabi:
Thank you! 👍

PR #172, opened 2020-08-07
Updating the documents to same sort of format, updating where needed.

dj-wasabi:
##### SUMMARY

Made some updates to the documentation for the roles. Updated the TOC, added some more clarifications. Made sure that they all have some kind of a format/layout.

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/ZABBIX_AGENT_ROLE.md
+++ b/docs/ZABBIX_AGENT_ROLE.md
@@ -11,12 +11,15 @@ Table of Contents
   * [Issues](#issues)
 - [Role Variables](#role-variables)
   * [Main variables](#main-variables)
+    + [Overall Zabbix](#overall-zabbix)
+    + [SElinux](#selinux)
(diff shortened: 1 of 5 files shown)


dj-wasabi:
> Nicely polished, is apache no longer a dependency for zabbix_web?

Thanks!

I don't want to force people using a specific role, just because I want it so. Compared with Puppet, where I had to "force" people using puppet modules from a specific user so that they were able to work with the `wdijkerman/zabbix` module, this isn't the case with Ansible. Yes true, I had the `geerlingguy.apache` configured in the meta config (I had to start somewhere), but I want to get rid of that as well. People should use or be able to use their own Ansible role of choice for either Apache or Nginx (or other Reverse Proxy) and the `zabbix_web` should be able to work with that.

Issue #77, opened 2020-06-10
Document a way to migrate to collection from standalone roles and modules


D3DeFi:
##### SUMMARY
`docs/MIGRATION.md` or something similar required for users to ease them a pain with transferring to the collection.

##### ISSUE TYPE
- Documentation Report

rockaut:
That would be great but I think that is a whole lot of a task as there are so many issues with ansible-base and collections as such and pretty much like any collection I use/stumble across is having the same problems in this task.
So I think this won't go well until ansible-base hasn't stabilized it's collection implementation with 2.10 and 2.9 updates.

D3DeFi:
Indeed true, at this point it is more like providing a few steps for users to help with transition from:
`dj-wasabi/zabbix-$role -> community.zabbix.zabbix_$role` and `zabbix_module -> community.zabbix.zabbix_$module`. And I am a bit behind with this as we've already released 0.2.0

Everything else is just to unstable rn.

D3DeFi:
This is now obsolete, there was a upstream documentation for 2.10 on how to transition to collections

PR #173, opened 2020-08-07
Updating the variables to the rest of the roles

dj-wasabi:
##### SUMMARY

Some properties (Mostly of them were for the javagateway) need to have the same kind of "convention". Like a `zabbix_{comp}_version` was introduced, but were not applied to the proxy and javagateway. Also the java gateway, didn't had a `zabbix_` prefix.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/roles/zabbix_javagateway/defaults/main.yml
+++ b/roles/zabbix_javagateway/defaults/main.yml
@@ -1,16 +1,17 @@
 ---
 # defaults file for dj-wasabi.zabbix-javagateway
 
-zabbix_version: 5.0
-zabbix_repo: zabbix
-zabbix_package_state: present
+zabbix_javagateway_version: 5.0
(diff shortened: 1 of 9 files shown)


dj-wasabi:
Yes, i totally forgot. The only 4 javagateway related variables and the `package_state` is something that people need to change. But I am not sure if anyone is using the java gateway, I don't I ever remembered in getting a PR or so for this role.. 😃 

Should we aim for #77 to have it available with the 1.0.0 release?

D3DeFi:
Not sure with #77. If you are up for preparing a few steps related to roles about how to migrate from previous dj-wasabi.ansible-zabbix-XYZ to community.zabbix.zabbix_XYZ then please open a PR somewhere into the docs/ and we then think about how to smoothly link it from main README.

PR #215, opened 2020-08-27
Remove geerlingguy.apache dependency from zabbix_web

dlundgren:
##### SUMMARY
Removes the dependency on `geerlingguy.apache` that zabbix_web has defined.

Setting `zabbix_websrv: nginx` doesn't work unless you also require `geerlingguy.apache` on your own.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_web

### DIFF
--- a/roles/zabbix_web/meta/main.yml
+++ b/roles/zabbix_web/meta/main.yml
@@ -25,8 +25,4 @@ galaxy_info:
     - monitoring
     - zabbix
 
-dependencies:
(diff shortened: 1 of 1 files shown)


dj-wasabi:
Hi @dlundgren 

I will close this issue as there were some other issues while removing the dependency. I made additional changes to make that wor that are in PR #217 . So once PR #217 is merged, there will be no dependency anymore.

PR #224, opened 2020-09-15
apt packages cache_valid_time 1h

piotrkochan:
##### SUMMARY
This pr sets apt cache to 3600 seconds as it is very annoying to wait every time playbook is being run.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
role/zabbix_agent

##### ADDITIONAL INFORMATION
None


dj-wasabi:
Hi @piotrkochan 

Is there a way you can make this configurable? So people can decide for them self what the best value is?

piotrkochan:
I have looked at the agent role one more time and now I see that there is default variable `zabbix_apt_force_apt_get` which once set to false makes everything I wanted. Code changes not required.

PR #226, opened 2020-09-16
Fix to decode Unicode Escape of multibyte strings

sky-joker:
##### SUMMARY

This PR is to fix to decode Unicode Escape of multibyte strings in an importing template content.  
If using Python2, this module hasn't decoded Unicode Escape until now.  
Exporting template data with this module (or zabbix_template_info) and a copy module creates a file with multibyte strings encoded as a byte string (Unicode Escape).  
So, a garbled text occurs(byte strings) on Zabbix after to import using that.  
This issue only occurs in Python2.

fixes: https://github.com/ansible-collections/community.zabbix/issues/225

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

### DIFF
--- a/changelogs/fragments/226-zabbix_template.yaml
+++ b/changelogs/fragments/226-zabbix_template.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - zabbix_template - Fixed to decode Unicode Escape of multibyte strings in an importing template data(https://github.com/ansible-collections/community.zabbix/pull/226).
--- a/plugins/modules/zabbix_template.py
+++ b/plugins/modules/zabbix_template.py
@@ -308,6 +308,7 @@ template_xml:
 
 import json
 import traceback
+import codecs
 import xml.etree.ElementTree as ET
 
 from distutils.version import LooseVersion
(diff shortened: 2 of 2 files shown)


D3DeFi (approved):
To me, this logs good. However, I would still like a confirmation from related bug issue that it solves the problem.

D3DeFi:
@sky-joker can we rebase? Do you think we can merge without confirmation from the related issue? Would like to get this into 1.2.0 also

sky-joker:
@D3DeFi 

rebase done :)

D3DeFi:
After testing this myself with English/Non-English words and Python2.7/3.8 both with and without this patch I confirm it is working as inteded. Good job @sky-joker ! Very nice and clean patch

sky-joker:
Thank you @D3DeFi !

elrondvega:
This is breaking one of my templates. I have a json template with
`"key": "logrt[\"/file.log\",\"r.*t=([0-9.]+)\",,,skip,\\1]"`

removing the 
`template_content = codecs.decode(template_content, 'unicode-escape')`
line fixes it.

elrondvega:
Here is an example template that I'm trying to import that fails.

```
{
    "zabbix_export": {
        "version": "5.2",
        "date": "2021-01-13T17:38:21Z",
        "templates": [
            {
                "template": "test",
                "name": "test",
                "description": "test",
                "groups": [
                    {
                        "name": "test"
                    }
                ],
                "items": [
                    {
                        "name": "test",
                        "type": "ZABBIX_ACTIVE",
                        "key": "logrt[\"/test.*test.log\",\"test.*total=([0-9.]+)\",,,skip,\\1]"
                    }
                ]
            }
        ]
    }
}

```

PR #237, opened 2020-10-05
Fix not included tasks with tags

j8r:
##### SUMMARY

Always include files with tasks.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`community.zabbix.zabbix_agent`

##### ADDITIONAL INFORMATION

When running the role (through a playbook) with a tag like `--tags=config`, not much tasks will run.
For example, `Debian.yml` has tasks with tags, but the file will never be included so they are useless.
Even worse, file which include necessary variables won't be included too. 

The rationale is to always include files with tasks, then only included tasks which have the matching tags will ran.

### DIFF
--- a/roles/zabbix_agent/tasks/main.yml
+++ b/roles/zabbix_agent/tasks/main.yml
@@ -8,6 +8,8 @@
   when:
     - zabbix_agent2 is defined
     - zabbix_agent2
+  tags:
+    - always
 
 - name: "Fix facts for linuxmint - distribution release"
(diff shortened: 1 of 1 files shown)


D3DeFi (approved):
Looks good, thank you for your contribution! 👍

dj-wasabi (approved):

dj-wasabi:
Thanks! 👍

PR #254, opened 2020-11-02
Disable no_log warning for override_password in module zabbix_user

telegnom:
##### SUMMARY
The user module generates the warning: `Module did not set no_log for override_password` because part of the options name is `password`. Ansible has a built in checker for fields like this to prevent you from leaking passwords into log files or things like that.
But here the field is just a boolean and therefore no danger of leaking any sensitive information. So the checking mechanism can be turned off to get rid of this warning.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_user.py

### DIFF
--- a/changelogs/fragments/254-disable-no-log-warning.yml
+++ b/changelogs/fragments/254-disable-no-log-warning.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - zabbix_user - disable no_log warning for option override_password.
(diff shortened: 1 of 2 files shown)


sky-joker:
Thank you @telegnom for this patch.  
A changelog file needs when the module feature or options are modified.  
So can you please enter a changelog file?

https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs

telegnom:
Thanks for your help. And as this is my very first commit to an ansible related project I'm not too familiar with the rules, yet. ;)

sky-joker (approved):
@telegnom
LGTM
Thank you for your contribution :)

PR #255, opened 2020-11-02
Fully automate dev setup with Gitpod

rockaut:
Basic gitpod integration. Would enable to directly start an vscode like environment in gitpod for quick edits/PRs.
"Sadly" currently just for editing and maybe building/linting the code. Docker-Compose or starting other containers isn't supported currently so we can't start a zabbix-environment to test against.

- https://gitpod.io/
- https://www.gitpod.io/docs/config-gitpod-file/

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,9 @@
+FROM gitpod/workspace-full
+
+# Install custom tools, runtimes, etc.
+# For example "bastet", a command-line tetris clone:
+# RUN brew install bastet
+#
+# More information: https://www.gitpod.io/docs/config-docker/
+RUN pip install wheel && \
+  pip install ansible-base zabbix-api
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,6 @@
+image:
+  file: .gitpod.Dockerfile
+
+tasks:
+  - init: 'echo "TODO: Replace with init/build command"'
+    command: 'echo "TODO: Replace with command to start project"'


sky-joker:
Just out of curiosity, why is gitpod to need?
Is it because you want to use the web browser for development?
Sorry, I don't understand the usefulness of gitpod because I haven't used it.

rockaut:
First it would enable users to make easier PRs as there isn't even any installation needed yes. Then I also like the idea of collaborative coding which is also possible with it.

For us specific you are kind of right - there is not much benefit currently as gitpod doesn't support docker-compose usage.
But as it is supported to use own images we might create an "all-in-one" zabbix image (like the from me personally so much hated appliance (-; ). If that works you got an complete solution to code around without even touching local.

Finally with ansible-runner/-builder inbound gitpod might also be a good match to make testing easier.

sky-joker:
I see. Thank you for your explanation.  
I think it is okay to merge if the gitpod useful for this collection development.  
What do you think? @D3DeFi

D3DeFi:
Sorry, slipped my attention. I've thought that there is a practice for such files to end in .gitignore as one does usually not want to impose his/hers coding behaviors on others. I am not sure we all will be ever using this (e.g. I had multiple reasons to code offline in the past due to travel/limited date coverage/etc..)

As for 'all-in-one' I am strongly against, we have been there already with ansible-container, although extreme example, and it was too large and created a lots of problems. Small containers are fast and quickly disposable. 

Anyway, if you need this merged in repository in order to be able to code on Gitpod, we can merge it. If you can use Gitpod without this being in the repo, then lets not merge. we can move Gitpod as an idea for later, but for now I would worry more about actual plugins and roles rather than unified coding environments.

D3DeFi:
Please reopen in regards to my previous comment if necessary

PR #274, opened 2020-11-27
Added a new property 'zabbix_agent_dont_detect_ip'

### LINKED ISSUE #267: Introduce a variable to skip checks for public/private IP address in the zabbix-agent role
##### SUMMARY
As far as I can see the dependency on python-netaddr is only needed to set a fact, which I would like to avoid. I manually removed the code from tasks/Linux.yaml and it works perfectly without the additional package.

##### ISSUE TYPE
- I would suggest a new Ansible variable like **zabbix_agent_dont_detect_ip** as a boolean, which is by default False to not break the current code and then use that in the when clause inside of Linux.yaml for the detection code.

##### COMPONENT NAME
zabbix-agent

##### ADDITIONAL INFORMATION
It would remove the requirement to install the python-netaddr package which is only used to detect the IP address of the node for an Ansible fact.

```yaml

```

dj-wasabi:
##### SUMMARY

Added a new property 'zabbix_agent_dont_detect_ip'

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```
Fixes #267

### DIFF
--- a/changelogs/fragments/agent-not-detect-ips.yml
+++ b/changelogs/fragments/agent-not-detect-ips.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "Added a new property `zabbix_agent_dont_detect_ip` when set to true, it won't detect the ips and no need to install the python module `netaddr`."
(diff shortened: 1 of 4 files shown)


ronnybremer:
I will pull the changes and test it to give feedback. Thanks :)

ronnybremer:
Looks perfect to me, thank you for the quick fix. 👍

PR #316, opened 2021-01-26
Updated windows long version number

Park0:
I was testing the windows agent and it url generated was incorrect. 
Fixed the version number


dj-wasabi:
Thank you! 👍

Issue #314, opened 2021-01-19
unicode escape breaks templates


elrondvega:
##### SUMMARY

commit #226 breaks templates that have slash escaped characters.

removing the line:
```
template_content = codecs.decode(template_content, 'unicode-escape')
```
fixes the problem.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

zabbix_template

##### ANSIBLE VERSION

```paste below

```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT / Zabbix Version

zabbix 5.2

##### STEPS TO REPRODUCE

Here is a template that breaks.

```json

{
    "zabbix_export": {
        "version": "5.2",
        "date": "2021-01-13T17:38:21Z",
        "templates": [
            {
                "template": "test",
                "name": "test",
                "description": "test",
                "groups": [
                    {
                        "name": "test"
                    }
                ],
                "items": [
                    {
                        "name": "test",
                        "type": "ZABBIX_ACTIVE",
                        "key": "logrt[\"/test.*test.log\",\"test.*total=([0-9.]+)\",,,skip,\\1]"
                    }
                ]
            }
        ]
    }
}
```

##### ACTUAL RESULTS

```paste below

```

D3DeFi:
Thank you for summarizing this in one issue @elrondvega . Sorry to hear that we broke this logic and complicated things for you.

@sky-joker would you be able to please look at this matter further?

elrondvega:
changing the code type from "unicode-escape" to "raw-unicode-escape" also resolves my issue, however I don't know how this affects the intended code change.

D3DeFi:
Btw are you experiencing this problem only with python2?

sky-joker:
Thank you @elrondvega for reporting this issue.
As @D3DeFi written, can you please let me know your using environment?

* Python version
* Ansible version

I’ll check to see if the same error occurs.

sky-joker:
I confirmed that this error occurred even in my lab with Python2/3.  
Umm... sorry adding this line may have been a mistake.

https://github.com/ansible-collections/community.zabbix/blob/2f8c3a4b9a00dcf2fb8156c210d90c0ea45655af/plugins/modules/zabbix_template.py#L625

The cause of the issues is the slash to be removed from the template_content by doing unicode-escape.  
For example, please look at the following code.

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-
import codecs

[...]

PR #322, opened 2021-01-31
zabbix_template: Fix the issue for the unicode strings doesn’t decode in python2

sky-joker:
##### SUMMARY

I added the codecs processing for decoding unicode in Python2 to the zabbix_template module previously.  
https://github.com/ansible-collections/community.zabbix/pull/226

It turns out that this process also removes the necessary escape characters(slash) in Python2 and 3.  
So, this PR will fix that issue.

fixes: https://github.com/ansible-collections/community.zabbix/issues/314

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

tested on Zabbix 5.0

### DIFF
--- a/changelogs/fragments/322-zabbix_template.yml
+++ b/changelogs/fragments/322-zabbix_template.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zabbix_template - fixed the bug that it doesn't decode the Unicode in Python2 (https://github.com/ansible-collections/community.zabbix/pull/322).
--- a/plugins/modules/zabbix_template.py
+++ b/plugins/modules/zabbix_template.py
@@ -308,7 +308,7 @@ template_xml:
 
 import json
 import traceback
-import codecs
+import re
 import xml.etree.ElementTree as ET
 
 from distutils.version import LooseVersion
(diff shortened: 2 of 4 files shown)


D3DeFi (approved):
Looks ok to me

sky-joker:
Thank you @D3DeFi  for your review.
I removed WIP, so can you please look at this PR again?

D3DeFi:
@sky-joker I think that added integration test will not be able to detect the problem by itself as the bug will not prevent it from importing the template, rather would cause the string with `"\"` to end up in zabbix as `"\\"`. 

I think this would require exporting the string via zabbix_template_info and comparing values between original template and result from export to be sure.

sky-joker:
Thank you @D3DeFi for your feedback.
I added the test for comparing the Unicode!

D3DeFi (approved):
Amazing! :)

D3DeFi:
Thank you for the fix @sky-joker . Now, the question is: do we want to release 1.2.1 with this? I feel like we should some how do that

sky-joker:
Thank you @D3DeFi for merging!

@dj-wasabi 
Is there anything you want to include in v1.2.1 from the PRs that have not yet been merged?
If nothing, I think that we release v1.2.1 is ok.

D3DeFi:
Just for the completeness in this thread:

As mentioned on Gitter, we skip 1.2.1 and go for 2.0.0 directly

sky-joker:
Thank you @D3DeFi for the follow.

PR #350, opened 2021-03-17
fix: zabbix_maintenance - added visible_name

tuxmartin:
##### SUMMARY

If host have different names in zabbix (for example `hostname='server.example.net'`, `visible hostname='my test server'`), ansible module `zabbix_maintenance` crash, because it use [API](https://www.zabbix.com/documentation/5.0/manual/api/reference/host/object) property `name`  (visible name of the host) instead of `host` (technical name of the host).

I left the current state as default. So it's backwards compatible.
But it is now possible to select the type of name using `visible_name`.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_maintenance

### DIFF
--- a/plugins/modules/zabbix_maintenance.py
+++ b/plugins/modules/zabbix_maintenance.py
@@ -65,6 +65,7 @@ options:
             - I(visible_name=yes) to search by visible name,  I(visible_name=no) to search by technical name.
         type: bool
         default: 'yes'
+        version_added: '2.0.0'
 
 extends_documentation_fragment:
 - community.zabbix.zabbix
--- a/plugins/modules/zabbix_maintenance.py
+++ b/plugins/modules/zabbix_maintenance.py
@@ -59,9 +59,10 @@ options:
             - Type of maintenance. With data collection, or without.
         type: bool
         default: 'yes'
-    visible_name
+    visible_name:
         description:
-            - Type of zabbix host name to use. No: Technical name of the host. Yes: Visible name of the host.
+            - Type of zabbix host name to use for identifying hosts to include in the maintenance.
+            - I(visible_name=yes) to search by visible name,  I(visible_name=no) to search by technical name.
         type: bool
         default: 'yes'
 
--- a/plugins/modules/zabbix_maintenance.py
+++ b/plugins/modules/zabbix_maintenance.py
@@ -59,6 +59,11 @@ options:
             - Type of maintenance. With data collection, or without.
         type: bool
         default: 'yes'
+    visible_name
+        description:
+            - Type of zabbix host name to use. No: Technical name of the host. Yes: Visible name of the host.
+        type: bool
+        default: 'yes'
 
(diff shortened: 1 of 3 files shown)


plugins/modules/zabbix_maintenance.py
```diff
@@ -59,6 +59,11 @@
             - Type of maintenance. With data collection, or without.
         type: bool
         default: 'yes'
+    visible_name
+        description:
+            - Type of zabbix host name to use. No: Technical name of the host. Yes: Visible name of the host.
```
D3DeFi, line 64:
```suggestion
    visible_name:
        description:
            - Type of zabbix host name to use for identifying hosts to include in the maintenance.
            - I(visible_name=yes) to search by visible name,  I(visible_name=no) to search by technical name.
```

D3DeFi (commented):
Hey @tuxmartin . Thank you for this addition, just slight adjustments to the DOCUMENTATION block and it is good to go.

sky-joker (approved):
Thank you @tuxmartin for the contribution!
LGTM

D3DeFi:
Everything looks ok now, can you please do one more thing? We need a changelog snippet in `changelogs/fragments/350-zabbix_maintenance-visible_name.yaml` with the content:

```yaml
minor_changes:
- zabbix_maintenance - it is now possible to target hosts by their technical name if it differs from the visible name
```

^ if I've put it in correct order

plugins/modules/zabbix_maintenance.py
```diff
@@ -59,6 +59,12 @@
             - Type of maintenance. With data collection, or without.
         type: bool
         default: 'yes'
+    visible_name:
+        description:
+            - Type of zabbix host name to use for identifying hosts to include in the maintenance.
+            - I(visible_name=yes) to search by visible name,  I(visible_name=no) to search by technical name.
+        type: bool
+        default: 'yes'
```
sky-joker, line 67:
```suggestion
        default: 'yes'
       version_added: ''
```

Sorry, I have missed this.  
Can you please add the version_added of the version to be added?

@D3DeFi 
Am I understanding correctly that the next release version 2.0.0?

sky-joker (changes requested):
Can you please check my comment?

plugins/modules/zabbix_maintenance.py
tuxmartin: it's done

sky-joker (approved):
LGTM

@D3DeFi 
If the version is correct, can you please merge this PR at your timing?

tuxmartin:
@D3DeFi @sky-joker Thank you for merge. When will the next version be released?

D3DeFi:
@tuxmartin sorry for my very late replies. I will probably release 1.4.0 in a week or so and change this feature to that version as I still don't see a reason to release 2.0.0, although I would like to finally drop Zabbix 3.0

PR #352, opened 2021-03-19
Check database for existing schema, not done file

ChandlerSwift:
##### SUMMARY
This will handle the case where the database is recreated, as well as
the case where the zabbix_server instance is recreated and the database
is not.
##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_server role


dj-wasabi:
Thank you! This looks amazing, will apply it as well to the proxy one (And take a look at the postgres'es).. 👍

PR #345, opened 2021-03-13
fixed failing sanity tests when no_log was missing

D3DeFi:
##### SUMMARY
Fix for https://github.com/ansible-collections/overview/issues/45#issuecomment-797617217

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/zabbix_action.py
plugins/modules/zabbix_discovery_rule.py

PR #353, opened 2021-03-20
1.3.0prep

D3DeFi:
##### SUMMARY
1.3.0 mainly because of the #345

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,41 @@ community.zabbix Release Notes
 .. contents:: Topics
 
 
+v1.3.0
+======
+
+Release Summary
+---------------
+
+| Release date: 2021-03-20 | Last major release to support Zabbix server 3.X versions in plugins.
+
+Minor Changes
+-------------
+
+- zabbix_agent - added support for installations on arm64 systems (https://github.com/ansible-collections/community.zabbix/pull/320).
+- zabbix_proxy - now supports configuring StatsAllowedIP (https://github.com/ansible-collections/community.zabbix/pull/337).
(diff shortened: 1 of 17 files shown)


dj-wasabi:
No 2.0.0?

D3DeFi:
Nah, actually #345 requires 1.2.1, but I honestly don't know how to release it with just a security patch.

Only think I can think of is to checkout new branch from 1.2.0 latest commit a cherry-pick the security fixes to it, but then we would be stack with the branch forever.

Also not possible to release 1.2.1 if there are som many features prepared to be released in roles since 1.2.0 so I settled for the middle ground.

I still don't know what would 2.0.0 bring us except possibility to drop zabbix 3.x for modules.

dj-wasabi:
I always thought that with Ansible 3.0, we had to deliver a version 2.0.0. But maybe I am/was wrong about this.

D3DeFi:
> I always thought that with Ansible 3.0, we had to deliver a version 2.0.0. But maybe I am/was wrong about this.

Major versions of collections work only in one way. They are considered breaking releases and will not be included by default in current ansible version. There is no requirement to have major version ready for each new release of ansible.

Ansible 3.0.z is out already. That means if we release 2.0.0 of this collection it will not be included in any ansible 3.y.z release, only minor and bugfix releases for the major version that is in the ansible 3.0.x will be included with minor releases of ansible (that means any 1.y.z).

E.g. we do 2.0.0 now then it will get to ansible 4.0.0. We do 1.3.0 now it will get to both ansible 3.1.0 and 4.0.0 if we dont create collection 2.0.0 before release of 4.0.0.

D3DeFi:
Im am going ahead then and merge this since we are long overdue with security fixes release I guess

PR #373, opened 2021-05-12
Create zabbix_inventory.py

ttestscripting:
##### SUMMARY
Create a Zabbix inventory Plugin 
community.zabbix.zabbix_inventory

##### ISSUE TYPE
- New Inventory Module community.zabbix.zabbix_inventory

##### COMPONENT NAME
community.zabbix.zabbix_inventory

##### ADDITIONAL INFORMATION
Zabbix inventory plugin to create dynamic inventory from Zabbix.

### DIFF
--- a/plugins/inventory/zabbix_inventory.py
+++ b/plugins/inventory/zabbix_inventory.py
@@ -34,7 +34,7 @@ options:
         description: Proxy server to use for reaching zabbix API
         type: string
         default: ''
-    host_zapi_query: 
+    host_zapi_query:
         description:
             - API query for hosts - see zabbix documentation for more details U(https://www.zabbix.com/documentation/current/manual/api/reference/host/get)
         type: dict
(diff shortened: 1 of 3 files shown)


ttestscripting:
Not sure how to resolve the checks that are failing. Any help/advice on resolving the issues would be greatly appreciated, Both pulls I have submitted appear to be failing at the same point.

plugins/inventory/zabbix_inventory.py
```diff
@@ -0,0 +1,380 @@
...
+        os.environ['HTTPS_PROXY'] = proxy
+
+        server_url = self.get_option('server_url')
+        http_login_user = self.get_option('login_user')
+        http_login_password = self.get_option('login_password')
+        validate_certs = self.get_option('validate_certs')
+        timeout = self.get_option('timeout')
+        self._zapi = ZabbixAPI(server_url, timeout=timeout, user=http_login_user, passwd=http_login_password, validate_certs=validate_certs)
+        self.login()
+        self._zbx_api_version = self._zapi.api_version()[:5]
```
D3DeFi, line 317: Can you please also add `logout` handling? See this [line of code](https://github.com/ansible-collections/community.zabbix/blob/1fc3d457949a9e1f7857cdea94e462cb728fcd92/plugins/module_utils/wrappers.py#L56)

Maybe you would even be able to utilize whole wrapper?

plugins/inventory/zabbix_inventory.py
```diff
@@ -0,0 +1,380 @@
...
+      type: bool
+      default: yes
+    add_zabbix_groups:
+      description:
+       - If set to True, hosts will be added to groups based on their zabbix groups
+      type: bool
+      default: false
+extends_documentation_fragment:
+    - constructed
+    - inventory_cache
```
D3DeFi, line 223: Wasn't working like this? You wouldn't have to specify `*_login_*, server, timeout, validate_ssl`

```suggestion
extends_documentation_fragment:
    - constructed
    - inventory_cache
    - zabbix
```

D3DeFi (changes requested):
Thank you for this contribution @ttestscripting ! Plugin looks very neat! :rocket: 

Can you please also add:
- changelog fragment (see [this](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs) and [this example](https://github.com/ansible-collections/community.zabbix/pull/353/commits/fcff7028a1ef1f19f1e803dacbf31f6ecea6e308))
- mention to README's included content (https://github.com/ansible-collections/community.zabbix/blob/main/README.md#included-content)
- `version_added` ([doc](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-fields))

CI should be fixed in #374


I wouldn't require this, but for the inspiration if you find yourself wanting to implement some testing for this module, this is how other collections implement it :)
- [amazon.aws](https://github.com/ansible-collections/amazon.aws/tree/main/tests/integration/targets/inventory_aws_ec2)
- [community.digitalocean](https://github.com/ansible-collections/community.digitalocean/blob/main/tests/unit/plugins/inventory/test_digitalocean.py)
- [azure](https://github.com/ansible-collections/azure/tree/dev/tests/integration/targets/inventory_azure)

plugins/inventory/zabbix_inventory.py
ttestscripting: I tried to use existing wrapper but ran into a few issues where it was prompting for a module name, so for sake of time, I added in the login piece to the inventory plugin and atexit.register(self._zapi.logout) for the logout. I believe I have added in the required documentation I am unsure how to generate the ansible community docs as I new to contributing but I updated the readme to point to the zabbix inventory plugin. Please let me know if this looks correct

D3DeFi: fair enough :)

plugins/inventory/zabbix_inventory.py
D3DeFi: @ttestscripting just curious if this would work. Have you tried it?

D3DeFi (approved):
Ignore codecov test, I think this is good to go
@sky-joker ?

plugins/inventory/zabbix_inventory.py
ttestscripting: That will work as well I am good with either method :)

D3DeFi: On the other hand, that doc fragment copies state of ZabbixBase and since this plugin is not using that, it would be safer to also not use doc_fragment if we decide to change something in the future that will not be reflected in this plugin.

Maybe eventually we will rework our base class as there already was an initiative for that in the past and then we will be able to migrate this plugin to use that also without any hussle

sky-joker (approved):
Thanks, @ttestscripting for the contribution.
LGTM

PR #377, opened 2021-05-17
New Module zabbix_globalmacro

ttestscripting:
Created Global Macros Plugin

SUMMARY
Created the Global Macro Plugin.

ISSUE TYPE
New Module Pull Request
COMPONENT NAME
Zabbix Module to create and update global macros.

ADDITIONAL INFORMATION
New module based on the hostmacro module to create and update global macros. Secret Macros will always be updated since there is no way to validate the current value.


D3DeFi:
please feel free to skip Zabbix 3.*, I think tests for zabbix_action use something like
```
- block:
  when: zabbix_version is version('4.0', '>=')
    - tasks...
```

Zabbix 4+ needs to be supported though as that version is still supported.

D3DeFi (commented):
Once tests are passing I think this is good to go. @sky-joker can you please also give it a look?

ttestscripting:
Looks like Zabbix 4.0 and older does not support the Type parameter for global macros, I will have to tweak it a bit for the support. I should be able to get to it later this week, Also I may have to update the documentation 5.4  has a new type of 2 - Vault secret. Thanks again for all of your help.

README.md
```diff
@@ -39,6 +39,7 @@ Click on the name of a plugin or module to view that content's documentation:
     - [zabbix_action](https://docs.ansible.com/ansible/2.10/collections/community/zabbix/zabbix_action_module.html)
     - [zabbix_group_info](https://docs.ansible.com/ansible/2.10/collections/community/zabbix/zabbix_group_info_module.html)
     - [zabbix_discovery_rule](https://docs.ansible.com/ansible/2.10/collections/community/zabbix/zabbix_discovery_rule_module.html)
+	- [zabbix_globalmacro](https://docs.ansible.com/ansible/2.10/collections/community/zabbix/zabbix_globalmacro_module.html)
```
sky-joker, line 42:
```suggestion
    - [zabbix_globalmacro](https://docs.ansible.com/ansible/2.10/collections/community/zabbix/zabbix_globalmacro_module.html)
```

tests/integration/targets/test_zabbix_globalmacro/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+- name: test - attempt to update global macro with string value
+  zabbix_globalmacro:
+    server_url: "{{ zabbix_server_url }}"
+    login_user: "{{ zabbix_login_user }}"
+    login_password: "{{ zabbix_login_password }}"
+    macro_name: zbxgmacro_test01
+    macro_value: abc
+    macro_type: 0
+  register: zbxgmacro_update
+  ignore_errors: True
```
sky-joker, line 85: I see that there is the ignore_erros here only, why is it necessary?

plugins/modules/zabbix_globalmacro.py
```diff
@@ -0,0 +1,222 @@
...
+  local_action:
+    module: community.zabbix.zabbix_globalmacro
+    server_url: http://monitor.example.com
+    login_user: username
+    login_password: password
+    macro_name: "{$EXAMPLE.MACRO}"
+    state: absent
+'''
+
+
```
sky-joker, line 102:
```suggestion
RETURN=r"""
"""
```
Is there any information that can be included in the RETURN block?
If any can be included in the return value, I think it is useful information for the module users.
What do you think?

sky-joker (changes requested):
Thanks, @ttestscripting for the contribution!
Rest LGTM

Could you please look at my comment?
And please also fix the integration errors.

D3DeFi:
> 
> 
> Looks like Zabbix 4.0 and older does not support the Type parameter for global macros, I will have to tweak it a bit for the support. I should be able to get to it later this week, Also I may have to update the documentation 5.4 has a new type of 2 - Vault secret. Thanks again for all of your help.

@ttestscripting you can use this as an inspiration

plugins/modules/zabbix_globalmacro.py
```diff
@@ -0,0 +1,258 @@
...
+        type: str
+    macro_type:
+        description:
+            - Type of the global macro Text or Secret Text.
+            - Required if I(state=present).
+            - 0 Text
+            - 1 Secret Text Works only with Zabbix >= 5.0 and will default to Text in lower versions
+            - 2 Vault secret Works only with Zabbix >= 5.2 and will default to Text in lower versions
+        type: str
+        choices: ['0', '1', '2']
```
D3DeFi, line 45: forgot one more thing, these should ideally be text values `['text', 'secret', 'vault']` or something similar and you need to do conversion to integers in the module. Its is sort of Ansible way of doing things that users should not consult API documentation when working with modules (although not doable 100% of a time)

ttestscripting: sure let me resolve that quickly, I will probably need to re-write the tests quickly as well

D3DeFi (approved):
Awesome! :rocket: :heart: 
Thank you for this plugin and for reacting to our reviews

@sky-joker I think we are ready to merge this. If you agree as well

changelogs/fragments/377-create_globalmacros_module.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- zabbix_globalmacros - it is now possible to create global macros using this module
```
sky-joker, line 2:
```suggestion
- zabbix_globalmacros - it is now possible to create global macros using this module (https://github.com/ansible-collections/community.zabbix/pull/377).
```

sky-joker: Finally, could you please add the PR URL?

D3DeFi:
yay! :)

D3DeFi:
With this one in, I will work on 1.4.0 release during weekend

Issue #380, opened 2021-05-21
zabbix_user should not overwrite user defined settings 


weehal:
##### SUMMARY
Every time when `zabbix_user`  task is called it overwrites user custom settings in zabbix - that makes them unusable. 

Don't update the user settings:
`lang, theme, refresh, rows_per_page, after_login_url`  etc. when differs from default or provide an override setting like passwd has.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
zabbix_user

##### ADDITIONAL INFORMATION
 it's similar to #261

D3DeFi:
Thank you for opening this @weehal . Yes, this is definitely a bug since API doesn't require us to provide those values:
https://www.zabbix.com/documentation/current/manual/api/reference/user/object#user

So we should default them to `None` and don't send them via request if user doesn't specify them.

sky-joker:
Thanks, @weehal for reporting the issue.  
I'll work on making a patch to resolve the issue.

sky-joker:
@weehal
I made a patch to resolve this issue.  
So could you please look at https://github.com/ansible-collections/community.zabbix/pull/382?  

Unfortunately, there are some default values that I can't remove because backward compatibility will be broken.  
But, those will be removed in the future.

PR #382, opened 2021-05-25
zabbix_user: Remove part of the default values from the parameters of the module.

sky-joker:
##### SUMMARY

This PR will remove some default values because the users want to change a configuration only if specified a parameter.
But, I'll leave part of the default value because a backward-compatible is broken if refresh and rows_per_page remove.  
(The integration test didn't be passed if refresh and row_per_page the default value remove.)  
In about those, I added the version to be removed into the documentation.

fixes: https://github.com/ansible-collections/community.zabbix/issues/380
fixes: https://github.com/ansible-collections/community.zabbix/issues/384

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

changelogs/fragments/382-zabbix_user.yml
plugins/module_utils/helpers.py
plugins/modules/zabbix_user.py
tests/integration/targets/test_zabbix_user/tasks/for_zabbix_50_lower.yml
tests/integration/targets/test_zabbix_user/tasks/for_zabbix_52_higher.yml

##### ADDITIONAL INFORMATION

### DIFF
--- a/changelogs/fragments/382-zabbix_user.yml
+++ b/changelogs/fragments/382-zabbix_user.yml
@@ -1,2 +1,2 @@
-minor_changes
+minor_changes:
   - zabbix_user - removed some default values because a configuration should be changed only if specified a parameter (https://github.com/ansible-collections/community.zabbix/pull/382).
--- a/changelogs/fragments/382-zabbix_user.yml
+++ b/changelogs/fragments/382-zabbix_user.yml
@@ -1,0 +1,2 @@
+minor_changes
+  - zabbix_user - removed some default values because a configuration should be changed only if specified a parameter (https://github.com/ansible-collections/community.zabbix/pull/382).
--- a/changelogs/fragments/382-zabbix_user.yml
+++ b/changelogs/fragments/382-zabbix_user.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - zabbix_user - removed some default values because a configuration should be changed only if specified a parameter (https://github.com/ansible-collections/community.zabbix/pull/382).
+  - zabbix_user - removed some of the default values because a configuration should be changed only if specified as a parameter (https://github.com/ansible-collections/community.zabbix/pull/382).
--- a/plugins/modules/zabbix_user.py
+++ b/plugins/modules/zabbix_user.py
@@ -612,9 +612,6 @@ def main():
         required_if=[
             ['state', 'present', ['usrgrps', 'refresh', 'rows_per_page', 'type']]
         ],
-        mutually_exclusive=[
-            ('autologin', 'autologout')
-        ],
         supports_check_mode=True
     )
 
--- a/plugins/modules/zabbix_user.py
+++ b/plugins/modules/zabbix_user.py
@@ -98,14 +98,10 @@ options:
     refresh:
         description:
             - Automatic refresh period in seconds.
-            - The default value will be removed at the version 2.0.0.
-        default: '30'
         type: str
     rows_per_page:
         description:
             - Amount of object rows to show per page.
-            - The default value will be removed at the version 2.0.0.
-        default: '50'
         type: str
     after_login_url:
         description:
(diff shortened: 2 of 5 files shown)


D3DeFi (approved):
Looks good to me, maybe we can even add some tests that will try this?

sky-joker:
@D3DeFi 
I removed WIP of the PR, so could you please look at this one again when you have time?

plugins/modules/zabbix_user.py
```diff
@@ -600,15 +608,15 @@ def main():
...
+        timezone=dict(type='str'),
+        role_name=dict(type='str'),
+        type=dict(type='str', choices=['Zabbix user', 'Zabbix admin', 'Zabbix super admin']),
         state=dict(type='str', default="present", choices=['present', 'absent'])
     ))
     module = AnsibleModule(
         argument_spec=argument_spec,
         required_if=[
-            ['state', 'present', ['usrgrps']]
+            ['state', 'present', ['usrgrps', 'refresh', 'rows_per_page']]
```
D3DeFi, line 619: Is it necessary to require these parameters with `state=present`? I think that Zabbix API itself will default them as they don't seem to be required with `user.create` API call

plugins/modules/zabbix_user.py
```diff
@@ -93,28 +89,27 @@
...
     autologout:
         description:
             - User session life time in seconds. If set to 0, the session will never expire.
             - If enable autologout, cannot enable autologin.
-        default: '0'
         type: str
     refresh:
         description:
             - Automatic refresh period in seconds.
+            - The default value will be removed at the version 2.0.0.
```
D3DeFi, line 101: Although I understand why you wanted to do it like this, I think that we can simply omit default value for `refresh` and `rows_per_page` as module defaults it to exactly the same values as Zabbix API itself, thus if we simply pass `None` via API call no backwards compatibility should be broken and it will simplify a few things for us (a.k.a not needing to change this further with 2.0.0)

changelogs/fragments/382-zabbix_user.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - zabbix_user - removed some default values because a configuration should be changed only if specified a parameter (https://github.com/ansible-collections/community.zabbix/pull/382).
```
D3DeFi, line 2:
```suggestion
  - zabbix_user - removed some of the default values because a configuration should be changed only if specified as a parameter (https://github.com/ansible-collections/community.zabbix/pull/382).
```

D3DeFi (changes requested):
@sky-joker thank you for this patch. Looks good, but please see my comments and let me know what you think especially about `refresh` and `rows_per_page`

plugins/modules/zabbix_user.py
sky-joker: The refresh and rows_per_page must be required when creating a new zabbix user.  
But, maybe those are unnecessary when updating it.  
As an alternative, I can be modified the following like. (I don't test yet, but...)

https://github.com/ansible-collections/community.zabbix/pull/382/files#diff-e484d4bad2263c86ac36f50e86e030cdfece1aae86e45755e64303286dd7a01bR440-R445

What do you think?

plugins/modules/zabbix_user.py
sky-joker: I think that this is the same issue as the above.  
So please see the above my comment.

plugins/modules/zabbix_user.py
sky-joker: For example, the role_name hadn't required to specify(we could have omitted it) because it set the default value.  
The parameter couldn't have required before in the zabbix_user module.  
But if the default value will be removed, it will be required when creating a new user.  
This is a temporary process to help that.

sky-joker: By the way, the existing integration test is failed if the default values are removed only.  
I think that this will be broken backward-compatible.  
I think that we should be fixing the existing integration test isn't failed until releasing version 2.0.0.
Sorry, If I don’t correctly understand, please tell me.

sky-joker: Ah, I get it.  
Sorry, that was my misunderstanding.  
If those params are None, I could create a new user by removing None from the request data with the helper_normalize_data function.  
I'll work on fixing the module based on your feedback.

D3DeFi: cool, thanks! :)

sky-joker:
Thanks, @D3DeFi for merging!

PR #376, opened 2021-05-17
Role - Zabbix Agent - Windows

ironbishop:
A couple additions to the Windows role.

fix #372 

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

PR #395, opened 2021-06-07
Boolean when is wrong

lingfish:
##### SUMMARY
I think this is meant to not be "not".

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
community.zabbix

### DIFF
--- a/roles/zabbix_agent/tasks/Windows.yml
+++ b/roles/zabbix_agent/tasks/Windows.yml
@@ -102,7 +102,7 @@
   with_items:
     - "{{ zabbix_agent_win_include }}"
   when:
-    - ('.' not in zabbix_agent_win_include)
+    - ('.' in zabbix_agent_win_include)
 
 - name: "Windows | Place TLS-PSK file"
   win_copy:


dj-wasabi:
@lingfish 

To be honest, I don't know? (Not a Windows User so I don't know what that 'dot' does/means...)

lingfish:
Hi, it was #376 by @ironbishop that added it.  Best that I can tell, they were trying to only run that task if `zabbix_agent_win_include` ended up being a string that looked like it had a dot in it, as in `zabbix_agentd.d`, following the standard of "dot d" include dirs.

ironbishop:
True, i didn't account for the '.d' folder. But this check is meant to skip the folder creation when you want to include a single file, not to a folder. Maybe something like `'.conf' not in zabbix_agent_win_include` would be better?

dj-wasabi:
So how do we continue? Shall I merge it, or will @lingfish update the PR with comment from @ironbishop ?

ironbishop:
i can't commit to his branch, but I can create my own if he's unable to go on with the change

dj-wasabi:
> i can't commit to his branch, but I can create my own if he's unable to go on with the change

Ok! 👍 
Will merge the PR. Thanks @lingfish and @ironbishop

Issue #399, opened 2021-06-17
Create a new host using agent2 use wrong port for interface create/update


qnox81:
##### SUMMARY
when using `zabbix_agent2: true` interfaces variable is setup incorrectly using `zabbix_agent_listenport` when it should use `zabbix_agent2_listenport` instead

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`roles/zabbix_agent/defaults/main.yml`

##### ANSIBLE VERSION
irrevelant as its variable definition issue

##### CONFIGURATION
irrevelant as its variable definition issue

##### OS / ENVIRONMENT / Zabbix Version
tested on Zabbix 5.4 but its valid on all Zabbix machines 

##### STEPS TO REPRODUCE
1. set `zabbix_agent2` to `true`
2. set `zabbix_agent2_listenport` to non standard port, like 10052
3. run role

##### EXPECTED RESULTS
host created/updated on Zabbix has port 10052

##### ACTUAL RESULTS
host created/updated on Zabbix has port 10050

PR #400, opened 2021-06-17
fix: always use zabbix_agent_listenport for creating/updating Zabbix host via API

qnox81:
…host even when agent2 is used

##### SUMMARY
use `zabbix_agent2_listenport` if `zabbix_agent2` is used for create/update host in Zabbix via API

Fixes: #399 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`roles/zabbix_agent/defaults/main.yml`


dj-wasabi:
Thanks! 👍

PR #403, opened 2021-06-23
set zabbix_agent_src_reinstall to false by default

fourstepper:
based on the discussion on gitter @dj-wasabi


dj-wasabi:
Thanks! 👍

PR #406, opened 2021-06-24
Fix zabbix_user module not working with Zabbix 5.4

### LINKED ISSUE #398: Zabbix 5.4: user.alias has been renamed to user.username
In Zabbix 5.4 the `user.alias` field no longer exists, querying for it does not work (the filter cannot be applied, causing all users to be returned). This leads to weird errors where the `zabbix_user` task will try to update the admin account when adding a new user.

See https://support.zabbix.com/browse/ZBXNEXT-6474

To fix this, both the legacy approach, as well as the 5.4 approach need to be supported.

href:
With 5.4, 'alias' has been renamed to 'username'. With this change it is
possible to use both parameters. The right one will be sent to the Zabbix
server transparently, depending on the server's version.

Resolves #398.

### DIFF
--- a/plugins/modules/zabbix_user.py
+++ b/plugins/modules/zabbix_user.py
@@ -23,6 +23,7 @@ options:
         description:
             - Name of the user alias in Zabbix.
             - alias is the unique identifier used and cannot be updated using this module.
+        aliases: [ username ]
         required: true
         type: str
     name:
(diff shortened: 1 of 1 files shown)


D3DeFi (approved):
Thank you for this patch. Pipeline for zabbix 5.4 is not enabled in our repo yet as it would fail, but running tests locally:
```
zabbix_version=5.4 docker-compose up -d
ansible-test integration test_zabbix_user
```

results in tests passing.

PR #407, opened 2021-06-24
Fix Zabbix 5.4 template imports failing

href:
Applications are no longer supported, everything is solved through tags now.

### DIFF
--- a/plugins/modules/zabbix_template.py
+++ b/plugins/modules/zabbix_template.py
@@ -622,6 +622,10 @@ class Template(ZabbixBase):
             if LooseVersion(self._zbx_api_version) >= LooseVersion('5.2'):
                 update_rules["templateDashboards"] = update_rules.pop("templateScreens")
 
+            # Zabbix 5.4 no longer supports applications
+            if LooseVersion(self._zbx_api_version) >= LooseVersion('5.4'):
+                update_rules.pop('applications', None)
+
             # The loaded unicode slash of multibyte as a string is escaped when parsing JSON by json.loads in Python2.
             # So, it is imported in the unicode string into Zabbix.
             # The following processing is removing the unnecessary slash in escaped for decoding correctly to the multibyte string.


D3DeFi:
Thank you for this patch. However, it seems there is also something else broken as `updateExisting` seems to be ignored for the Zbx 5.4 and module fails idempotency checks.

Currently, we don't have CI enabled for 5.4, but if run locally via docker:
```bash
zabbix_version=5.4 docker-compose up -d
ansible-test integration test_zabbix_template
```
It fails on the following step:
```
TASK [test_zabbix_template : Import Zabbix template from JSON file with matching values (idempotency check).] **********************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ********_api.Already_Exists: ('Error -32602: Invalid params., Template "ExampleHost" already exists. while sending {"jsonrpc": "2.0", "method": "configuration.import", "params": {"format": "json", "source": "{\\"********_export\\": {\\"date\\": \\"2020-05-29T15:22:11Z\\", \\"templates\\": [{\\"template\\": \\"ExampleHost\\", \\"name\\": \\"ExampleHost\\", \\"description\\": \\"\\", \\"groups\\": [{\\"name\\": \\"Linux servers\\"}, {\\"name\\": \\"Templates\\"}], \\"macros\\": [{\\"macro\\": \\"{$EXAMPLE_MACRO1}\\", \\"value\\": \\"1000\\"}, {\\"macro\\": \\"{$EXAMPLE_MACRO2}\\", \\"value\\": \\"text\\"}], \\"templates\\": [{\\"name\\": \\"FTP Service\\"}, {\\"name\\": \\"Zabbix Proxy\\"}]}], \\"version\\": \\"3.0\\", \\"groups\\": [{\\"name\\": \\"Linux servers\\"}, {\\"name\\": \\"Templates\\"}]}}", "rules": {"discoveryRules": {"createMissing": true, "updateExisting": true, "deleteMissing": true}, "graphs": {"createMissing": true, "updateExisting": true, "deleteMissing": true}, "groups": {"createMissing": true}, "httptests": {"createMissing": true, "updateExisting": true, "deleteMissing": true}, "items": {"createMissing": true, "updateExisting": true, "deleteMissing": true}, "templates": {"createMissing": true, "updateExisting": true}, "templateLinkage": {"createMissing": true, "deleteMissing": true}, "triggers": {"createMissing": true, "updateExisting": true, "deleteMissing": true}, "valueMaps": {"createMissing": true, "updateExisting": true}, "templateDashboards": {"createMissing": true, "updateExisting": true, "deleteMissing": true}}}, "auth": "a8ba01a7bd18a8db7965c2975e68c336", "id": 4}', -32602)
```

So in order to fix this problem we need to dig even deeper. 

Anyway, I think we can merge this patch and solve at least 50% of problems with zbx template rather than having it bugged 100%. Do you agree @sky-joker ?

D3DeFi:
hmm, this seems to work correctly tho when executed multiple times:
```python
from zabbix_api import ZabbixAPI

zbx = ZabbixAPI('http://localhost:8080')
zbx.login('Admin', 'zabbix')

rules = {
    'templates': {
        'createMissing': True,
        'updateExisting': True
    }
}

source = '''<?xml version="1.0" encoding="UTF-8"?>
<zabbix_export>
  <version>5.0</version>
  <templates>
    <template>
      <template>ExampleTemplate</template>
      <groups>
        <group>
          <name>Linux servers</name>
        </group>
      </groups>
    </template>
  </templates>
</zabbix_export>'''

zbx.configuration.import_(dict(format='xml', source=source, rules=rules))

zbx.logout()
```

plugins/modules/zabbix_template.py
```diff
@@ -622,6 +622,10 @@ def import_template(self, template_content, template_type='json'):
             if LooseVersion(self._zbx_api_version) >= LooseVersion('5.2'):
                 update_rules["templateDashboards"] = update_rules.pop("templateScreens")
 
+            # Zabbix 5.4 no longer supports applications
```
sky-joker, line 625: Thanks, @href for the patch!
Could you please add a changelog?

sky-joker:
@D3DeFi
I agree, please merge after adding a changelog.

D3DeFi:
I was able to verify that attempting to import the same file via Zabbix GUI also results in the same error. 

The file is https://github.com/ansible-collections/community.zabbix/blob/main/tests/integration/targets/test_zabbix_template/files/template1_52_higher.json

image

I believe there is nothing wrong with such json, or is it? I think this should be open as a bug for Zabbix

sky-joker:
@D3DeFi 
I tried whether the same error happens, but I didn't confirm in my environment.  
I used zabbix container version 5.4.1.

image

I wonder that why doesn't the error happen(environment dependent?).

D3DeFi:
@sky-joker it must be that we are doing something wrong in the zabbix_template. I am running against 5.4.0 (ubuntu-5.4-latest from our docker-compose) file with the following steps:

1. `ansible-test integration test_zabbix_template` <- fails
2. log in manually and attempt to import same template as described in the picture <- fails with template ExampleHost already exists

Other test is:

1. Log in manually and attempt to import the template
2. Do it once again -> result is 'No change'

I suspect there was a secret change of how API works (again..). I think this is how it works in the second case (thus not failing) in sort-of pseudocode:
```python
template = zbx.template.get()
cmpres = zbx.configuration.importcompare() if template
zbx.configuration.import() if cmpres.different
```

D3DeFi:
Now this is very interesting. If I FIRST upload template manually via `Import` button in Zabbix GUI and then try to print template diff I get:
```bash
[]
```

Buut, if I run `ansible-test integration test_zabbix_template` and then try to print template diff:
```bash
{
  "templates": {
    "added": [
      {
        "after": {
          "template": "ExampleHost",
          "name": "ExampleHost",
          "macros": [
            {
              "macro": "{$EXAMPLE_MACRO1}",
              "value": "1000"
            },
            {
              "macro": "{$EXAMPLE_MACRO2}",
              "value": "text"
            }
          ],
          "templates": [
            {
              "name": "FTP Service"
            },
            {
              "name": "Zabbix Proxy"
            }
          ],
          "uuid": "c1a80829556941e4a1ff307b39130b5a"
        }
      }
    ]
  }
}

```

This is my current test script:
```pythonfrom zabbix_api import ZabbixAPI
import json

zbx = ZabbixAPI('http://localhost:8080')
zbx.login('Admin', 'zabbix')


update_rules = {
    'discoveryRules': {
        'createMissing': True,
        'updateExisting': True,
        'deleteMissing': True
    },
    'graphs': {
        'createMissing': True,
        'updateExisting': True,
        'deleteMissing': True
    },
    'groups': {
        'createMissing': True
    },
    'httptests': {
        'createMissing': True,
        'updateExisting': True,
        'deleteMissing': True
    },
    'items': {
        'createMissing': True,
        'updateExisting': True,
        'deleteMissing': True
    },
    'templates': {
        'createMissing': True,
        'updateExisting': True
    },
    'templateLinkage': {
        'createMissing': True,
        'deleteMissing': True
    },
    'templateDashboards': {
        'createMissing': True,
        'updateExisting': True,
        'deleteMissing': True
    },
    'triggers': {
        'createMissing': True,
        'updateExisting': True,
        'deleteMissing': True
    },
    'valueMaps': {
        'createMissing': True,
        'updateExisting': True
    }
}

with open('tests/integration/targets/test_zabbix_template/files/template1_52_higher.json', 'r') as f:
    source_json = json.load(f)

x = zbx.configuration.importcompare(dict(format='json', source=json.dumps(source_json), rules=update_rules))
print(json.dumps(x, indent=2))

zbx.logout()
```

Now I am lost :)

D3DeFi:
Anyway, I will open new issue and merge this. Changelog will be added when I will prepare 1.4.0 release

sky-joker:
@D3DeFi 

Thanks, for the details.  
I confirmed that the same error happens.  
Indeed, that's strange behavior.

> Anyway, I will open new issue and merge this. Changelog will be added when I will prepare 1.4.0 release

ok, thanks in advance.

PR #419, opened 2021-07-06
Fix booleans in agent install

isuftin:
##### SUMMARY
The  incoming `zabbix_agent_listeninterface` variable was being used to check against `ansible_all_ipv4_addresses` which doesn't make sense.  Updated to check that the value for `zabbix_agent_listeninterface` looks at the list in the `ansible_interfaces` variable instead.

Also update how Ansible is looking at the `zabbix_agent_listeninterface` variable to verify that it exists and is not blank.

`zabbix_agent_listeninterface | bool` returns False even if it's defined.

`(zabbix_agent_listeninterface)` is the proper check here.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Zabbix Agent

### DIFF
--- a/roles/zabbix_agent/tasks/Linux.yml
+++ b/roles/zabbix_agent/tasks/Linux.yml
@@ -37,7 +37,7 @@
     msg: "The specified network interface does not exist"
   when:
     - (zabbix_agent_listeninterface)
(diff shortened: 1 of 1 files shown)


isuftin:
I made a PR from the wrong feature branch - Will create new PR

isuftin:
I've reopened this PR

PR #420, opened 2021-07-06
zabbix_agent_listeninterface validity check fix

isuftin:
SUMMARY
The incoming `zabbix_agent_listeninterface` variable was being used to check against `ansible_all_ipv4_addresses` which doesn't make sense. Updated to check that the value for `zabbix_agent_listeninterface` looks at the list in the `ansible_interfaces` variable instead.

Also update how Ansible is looking at the `zabbix_agent_listeninterface` variable to verify that it exists and is not blank.

`zabbix_agent_listeninterface | bool` returns False even if it's defined.

`(zabbix_agent_listeninterface)` is the proper check here.

ISSUE TYPE
Bugfix Pull Request
COMPONENT NAME
Zabbix Agent

### DIFF
--- a/roles/zabbix_agent/tasks/Linux.yml
+++ b/roles/zabbix_agent/tasks/Linux.yml
@@ -36,8 +36,8 @@
   fail:
     msg: "The specified network interface does not exist"
   when:
-    - zabbix_agent_listeninterface | bool
-    - (zabbix_agent_listeninterface not in ansible_all_ipv4_addresses)
+    - (zabbix_agent_listeninterface)
+    - (zabbix_agent_listeninterface not in ansible_interfaces)
   tags:
     - zabbix-agent
     - config


isuftin:
This PR is failing during CI but the issue seems unrelated.

dj-wasabi:
Hi @isuftin 

Sorry for late reply, can you do a rebase? I believe the issue has been fixed lately.

D3DeFi:
tried to rebase on my own:

```
dropping e6ebbc2156d92ab8c49b9de6b558291dda0921c5 fix: incoming interface variable checks against proper ansible variable, use pythonic boolean check for string variable -- patch contents already upstream
Successfully rebased and updated refs/heads/fix-erroneous-interface-check.
```

D3DeFi:
Oh actually it was the next one - #421

Issue #425, opened 2021-07-14
php-fpm socket not accesible by nginx 


baszoetekouw:
##### SUMMARY
In the php-fpm config, the fpm-processes are configured to run as the same user who owns the socket (either `zabbix_php_fpm_conf_user` or `zabbix_web_conf_web_user`).  Typically, that won't work, because the webserver (nginx) runs as a different user (`www-data` on Debian/Ubuntu).

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix-web

##### ANSIBLE VERSION

```paste below
ansible 2.10.5
  config file = /home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg
  configured module search path = ['/home/bas/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /bin/ansible
  python version = 3.9.5 (default, May 11 2021, 08:20:37) [GCC 10.3.0]

```

##### CONFIGURATION

```
ANSIBLE_FORCE_COLOR(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = True
ANSIBLE_NOCOWS(env: ANSIBLE_NOCOWS) = True
ANSIBLE_PIPELINING(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = True
ANSIBLE_SSH_ARGS(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = -o ControlMaster=auto -o ControlPersist=3600s
CACHE_PLUGIN(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = yaml
CACHE_PLUGIN_CONNECTION(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = .ansible/facts
DEFAULT_BECOME(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = True
DEFAULT_CALLBACK_WHITELIST(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = ['profile_tasks']
DEFAULT_FORKS(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = 25
DEFAULT_GATHERING(/home/bas/Documents/Repositories/SCZ-deploy/ansible.cfg) = smart
[...]

PR #426, opened 2021-07-14
Make php socket accessible for webserver group

baszoetekouw:
##### SUMMARY
Make php socket accessible for webserver group

In the current setup the user/group who owns the php-fpm socket is always identical to the user that is used to run the php-fpm processes (either zabbix_php_fpm_conf_user or zabbix_web_conf_web_user).

This produces an broken setup, as nginx/apache are running as different users (`www-data` on Debian), so they are not able to access the socket.

Fixes: #425 

I'm not entirely sure if this patch is really the nicest way to handle this, as now the user is no longer able to override the owner of the socket from a group var.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix-web


dj-wasabi:
Thanks bro! 👍

Issue #334, opened 2021-02-23
TimescaleDB support for Zabbix Server


KarlikB:
##### SUMMARY
Current zabbix_server role does not support to install and configure TimescaleDB extension for Postgres as described in https://www.zabbix.com/documentation/current/manual/appendix/install/timescaledb

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
Zabbix Server PostgreSQL DB backend

##### ADDITIONAL INFORMATION
I consider this proposed feature very helpful, the TimescaleDB is time-series extension on PostgreSQL and is able to significantly increase historic storage performance, simplifies housekeeping process and using compression it decreases used diskspace
The proposed feature will require add possibility to install PostgreSQL from its own repo (distro variants are usually too old) and add possibility to define installed version of the Postgres due to TimescaleDB compatibility matrix.
The TimescaleDB support has been added to Zabbix Server in version 5.0

klando:
this issue can be closed probably.

PR #428, opened 2021-07-16
TimescaleDB support for Zabbix Server role

SimBou:
##### SUMMARY
Modification of zabbix_server role to allow support of TimescaleDB.
It requires tha timescaledb is already installed (as PostgreSQL)

HINT: Include "TimescaleDB support for Zabbix Server #334"

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
TimescaleDB support for Zabbix Server

##### ADDITIONAL INFORMATION
You can use this ansible role (https://github.com/SimBou/ansible-role-timescaledb.git) to install TimescaleDB


dj-wasabi:
Thank you!

Can you either update the documentation that this only is available yet for the PostGreSQL, or apply the change also for MySQL? Otherwise we might get new issues created by people that it won't work for them, because they are using MySQL as database.

Thanks! 👍

dj-wasabi:
> Doc updated to explain that it works only for postresql

Thank you! 👍 
Will merge the PR when the last CI successfully ends.

PR #437, opened 2021-08-08
fix CI tests for module_defaults and zabbix_poller services

D3DeFi:
##### SUMMARY
I expect this to maybe fail for older zabbix version with zabbix poller trigger name. Pushing so I can sort what versions are affected

Edit: ready to go now

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/integration


D3DeFi:
they only changed name of triggers in 5.0...

PR #438, opened 2021-08-12
Add ability to set gpgcheck for yum repos

dingus9:
##### SUMMARY
Add the ability and property(zabbix_repo_yum_gpgcheck) to control gpg key checking on the default yum repo entries

This feature should allow users to more easily toggle gpg key checking with the default yum repo values

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
improvement to role

### DIFF
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -21,7 +21,7 @@ zabbix_agent_dont_detect_ip: False
 zabbix_agent_allow_key: []
 zabbix_agent_deny_key: []
 
-# Selinux related vars 
+# Selinux related vars
 selinux_allow_zabbix_run_sudo: False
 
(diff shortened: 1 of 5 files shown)


dj-wasabi:
Hi @dingus9 

Thank you for making time to create a PR and provide a improvement. Can I ask you if you can apply the same change to the other roles as well, so we have a consistent way of configuring repo's? Other people that has both the Agent and Server/Proxy on the same node will have no idempotency, as each role is changing the contents of the repo files.

Thank you!

dingus9:
Sure I'll look into that change on the others also

dingus9:
Does that look about right? Not sure if I got them all!

dj-wasabi:
Hi @dingus9 

Yes, looks very good! Thank you! 👍

PR #444, opened 2021-08-22
WIP: implement httpapi

rockaut:
> this is a work in progress and in early stages - PR already opened to get feedback and maybe contributions already

##### SUMMARY
This PR is implementing zabbix api requests over ansible "native" httpapi plugin structure. This would enable to strip the dependency of zabbix-api python module.

For example and starter there will be two new modules added to show and test the httpapi plugin.

##### ISSUE TYPE
- Feature Pull Request
- New Module Pull Request

##### COMPONENT NAME
httpapi
api_info

### DIFF
--- a/plugins/httpapi/api.py
+++ b/plugins/httpapi/api.py
@@ -1,0 +1,119 @@
+# (c) 2021, Markus Fischbacher (fischbacher.markus@gmail.com)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# Quick Link to Zabbix API docs: https://www.zabbix.com/documentation/current/manual/api
+
+from __future__ import absolute_import, division, print_function
+
+from ansible.plugins import AnsiblePlugin
+
+__metaclass__ = type
+
+DOCUMENTATION = """
+---
+author: fischbacher.markus@gmail.com
+httpapi : zabbix
+short_description: HttpApi Plugin for Zabbix
+description:
+  - This HttpApi plugin provides methods to connect to Zabbix over their HTTP(S)-based api.
+version_added: "1.0"
+options:
+  zabbix_token:
+    type: str
+    description:
+      - Specifies the api token path of the FTD device
+    vars:
+      - name: ansible_httpapi_zabbix_token
+"""
+
+import json
+
+from ansible.module_utils.basic import to_text
+from ansible.errors import AnsibleAction, AnsibleConnectionFailure
+from ansible.module_utils.six.moves.urllib.error import HTTPError
+from ansible.plugins.httpapi import HttpApiBase
+from ansible.module_utils.connection import ConnectionError
+
+
+BASE_HEADERS = {"Content-Type": "application/json"}
+
+
+import debugpy
+
+class HttpApi(HttpApiBase):
+
+    def set_become(self, become_context):
+      """As this is an http rpc call there is no elevation available
+      """
+      pass
+
+    def update_auth(self, response, response_text):
+      return None
+
+    def login(self, username, password):
+      self.connection._auth = {}
+
+    # def logout(self):
+    #   logout_path = '/my/logout/path'
+    #   self.send_request(None, path=logout_path)
+
+    #   # Clean up tokens
+    #   self.connection._auth = None
+
+    def send_request(self, request_method="POST", path="/api_jsonrpc.php", payload=None):
+        payload = json.dumps(payload) if payload else '{}'
+
+        import debugpy
+        debugpy.listen(5678)
+        print("Waiting for debugger attach")
+        debugpy.wait_for_client()
+        debugpy.breakpoint()
+        print('break on this line')
+
+
+        try:
+            self._display_request(request_method, path)
+            response, response_data = self.connection.send(
+                path,
+                payload,
+                method=request_method,
+                headers=BASE_HEADERS
+            )
+            value = to_text(response_data.getvalue())
+
+            try:
+              json_data = json.loads(value) if value else {}
+              if "result" in json_data:
+                json_data = json_data["result"]
+            # JSONDecodeError only available on Python 3.5+
+            except ValueError:
+              raise ConnectionError("Invalid JSON response: %s" % value)
+
+            return response.getcode(), json_data
+        except AnsibleConnectionFailure as e:
+            self.connection.queue_message("vvv", "AnsibleConnectionFailure: %s" % e)
+            if to_text("Could not connect to") in to_text(e):
+                raise
+            if to_text("401") in to_text(e):
+                return 401, "Authentication failure"
+            else:
+                return 404, "Object not found"
+        except Exception as e:
+            raise e
+
+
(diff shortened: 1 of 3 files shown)


D3DeFi:
judging by the implementation, would we need to keep some mapping of `ZabbixApiRequest.method() <-> API /resource` ?

rockaut:
I'm playing around if I make it more a manager style class or an Base Class. All other implementations are more of a single class style.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

rockaut:
> judging by the implementation, would we need to keep some mapping of `ZabbixApiRequest.method() <-> API /resource` ?
> 
> ```python
>     def get_api_version(self):
>       payload = self.payload_builder("apiinfo.version")
>       response = self.post(payload=payload)
>       return response
> ```
> 
> Some kind of translator between API and Ansible. One neat thing about zabbix-api is that it is always backwards/forward compatible with all of the Zabbix version, since it doesn't implement any calls statically. Everything used as zbx.X.Y is passed to `post('X.Y')`

I just now get what you mean. No I of course implement the api calls like zabbix-api does. It was just the test implementation and the specific handling of apiinfo.version.

I finally made good progress as I realized the netcommon.httpapi plugin doesn't plays well with zabbix handling of auth key in payload instead of http header. So in my last commit I just "borrowed" the netcommon.httpapi and redoing the auth handling ... and now it works. This would also enable us to do more or less an drop in replacement as I see it.

Still some work to do but in my first tests now I have already reduced the api calls significantly as now for each play the login/logout and apiinfo.version is only called ONCE!
Before: 1 play with 10 tasks = at least 30 api calls (api version, login, logout) + all really needed calls
Now: same playbook = 3 calls + the needed calls.
- login
- apiversion
- 10 x host.info (for example)
- logout

And also this will enable us instantly to already work with API Tokens for auth through the httpapi option settings.

rockaut:
Heja guys,

I have the initial implementation done for the httpapi and http connection plugin. I've made an api_info module initially but I removed it from this MR for now so that everything keeps clean. This MR doesn't touch any modules for now.

I would love if we can get this merged and proceed with the open questions around moving the modules to the httpapi in a separate issue - or more precisely in multiple issues.

There are lot's of open questions - the main thing is, that I don't see a chance to implement this WITHOUT breaking backwards compatibility. I'm currently looking into if we can make the transition smooth for users but I have not figured out a way.

Basic usage - for end users currently:

```yaml
---
# hosts.yml

all:
  children:
    api:
      hosts:
        apilocal:
          ansible_host: host.docker.internal                            # the host to connect to ( f.e.: zabbix.mydomain.lan )
          ansible_httpapi_port: 80
          ansible_httpapi_use_ssl: false
          ansible_httpapi_validate_certs: false
          ansible_network_os: community.zabbix.jsonrpc
          ansible_connection: community.zabbix.httpapi
          ansible_user: Admin
          ansible_httpapi_pass: zabbix                                    # this should be a vault secret of course or ...
          # ansible_httpapi_session_key: slaksdljasdlfj           # ... instead of password use a session key (zabbix token)

```

```yaml
---
# example playbook
# api_info module IS NOT included in this MR!

- hosts: api
  collections:
    - community.zabbix
  gather_facts: false
  tasks:
    - community.zabbix.api_info:
      register: zbx_api_info
    - debug:
        var: zbx_api_info

```

As you can see the amount of configuration and the useage is very simple thanks to ansibles httpapi implementation. Additionally we already get Token auth included with httpapi.

The basic usage for modules is also straight forward.

```python

from distutils.version import LooseVersion
from ansible.module_utils.basic import AnsibleModule

from ansible_collections.community.zabbix.plugins.module_utils.api_request import ZabbixApiRequest

def main():
    """Main ansible module function
    """

    argument_spec = dict()
    module = AnsibleModule(
        argument_spec=argument_spec,
        supports_check_mode=True
    )

    api_request = ZabbixApiRequest(module)
    v = api_request.api_version()

    module.exit_json(changed=False, response=v)

if __name__ == '__main__':
    main()

```

A module would just have to use ZabbixApiRequest class. This would be the drop in replacement for ZabbixApi class from zabbix-api.

For example with zabbix-api:

```python
host = self._zapi.host.get({
                'output': 'extend',
                'selectGroups': 'extend',
                'selectParentTemplates': ['name'],
                'hostids': hostinterface['hostid'],
                'selectInventory': host_inventory,
                'selectTags': 'extend',
                'selectMacros': 'extend'
            })
```

would be:

```python
api_request = ZabbixApiRequest(module)
host = api_request.host.get({
                'output': 'extend',
                'selectGroups': 'extend',
                'selectParentTemplates': ['name'],
                'hostids': hostinterface['hostid'],
                'selectInventory': host_inventory,
                'selectTags': 'extend',
                'selectMacros': 'extend'
            })
```

I think thats what you meant before @D3DeFi ?

There are open strings, questions and issues still:
* I haven't tried migrating another module for now (just the api_info)
* In the current implementation I've included a separate connection plugin which is basically just a "copy" of ansible.netcommon.httpapi . This is needed for now because of the way zabbix does authentication. The upside of this is, that we don't have ANY dependencies (neither on ansible than on python)
* I've only tested it on python3 currently
* Migration of modules ... basically it's drop in but given the config and playbook changes needed it would be a hard break. I'm currently looking into if we can "initialize" the httpapi with the old module options and thus no breaking change would be introduced

But as said above already those questions I would like to clarify in other issues then. Maybe we could also use the "Projects" functionality of github for this too?

rockaut:
Wow... I don't know how but I managed to get all tests pass :D

dj-wasabi:
> Wow... I don't know how but I managed to get all tests pass :D

And the tests are still there? 😄 Or did you silently changed it to `pass` 😉 

But this look good. I have requested review from @D3DeFi  and then we can merge it. From a roadmap p.o.v. we should make - or apply - the next changes you suggest to 2.0.0. If we have a backwards compatibility issue then we can mention this in the upgrade document.

Before we release 2.0.0, we will deploy an 1.6.0 though which contains Zabbix 6.0. (And some other small features or fix issues that people provide with PR's)

D3DeFi:
we had lengthy chat about this on gitter, I think it is good, maybe we cannot deprecate right away, but we can work on having some kind of switch that will allow users to have this new behavior via `new_connection: true` or something

@rockaut can you please extract key points of from our comments in this PR and make a new issue for the adoption process?

Maybe we can even create project for this on the repo - @dj-wasabi ?

rockaut:
@D3DeFi  Of course! I will implement an PoC. I will do it over on a separate GitHub repo for you guys to just clone and run it. Also I will open some Issues to split and discuss the points we made in the chat.

@dj-wasabi yeah an project would be cool to track the progress and plan issues ahead of future releases.

rockaut:
Off Topic side note: #444 cool issue uuid for this PR :D

dj-wasabi:
Yes, will create a new project for this where can have an overview of this implementation. 👍

dj-wasabi:
Created a project 'Plugins HTTPAPI'. So when someone creates an issue (or PR), we can select this project.

PR #456, opened 2021-09-12
zabbix_hostmacro/ add support for macro type secret, vault secret

### LINKED ISSUE #410: zabbix_hostmacro module does not support macro type (text/secret/vault)
Hello, following  [this comment](https://github.com/ansible-collections/community.zabbix/issues/392#issuecomment-868473496)

Zabbix 5.0 introduced secret text macro : https://www.zabbix.com/documentation/5.0/manual/config/macros/user_macros
And Zabbix 5.2 also introduced secret Vault macro : https://www.zabbix.com/documentation/5.2/manual/config/macros/user_macros

Macro types text/secrets/vault should be supported by zabbix_hostmacro module.

thanks

icez:
##### SUMMARY
Zabbix 5.0 and above have introduced a secret macro which allow user to store secret more securely. Here's the patch to allow zabbix_hostmacro module to set type of the macro accordingly.

Fixes #410

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_hostmacro

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/zabbix_hostmacro.py
+++ b/plugins/modules/zabbix_hostmacro.py
@@ -36,6 +36,15 @@ options:
             - Value of the host macro.
             - Required if I(state=present).
         type: str
+    macro_type:
+        type: int
+        description:
+            - Type of the host macro.
+            - 0 = Text macro (default)
+            - 1 = Secret macro (Works only with Zabbix >= 5.0)
+            - 2 = Vault secret (Works only with Zabbix >= 5.2)
+        choices: [0, 1, 2]
+        default: '0'
     state:
         description:
             - State of the macro.
@@ -123,24 +132,28 @@ class HostMacro(ZabbixBase):
             self._module.fail_json(msg="Failed to get host macro %s: %s" % (macro_name, e))
 
     # create host macro
-    def create_host_macro(self, macro_name, macro_value, host_id):
+    def create_host_macro(self, macro_name, macro_value, macro_type, host_id):
         try:
             if self._module.check_mode:
                 self._module.exit_json(changed=True)
-            self._zapi.usermacro.create({'hostid': host_id, 'macro': macro_name, 'value': macro_value})
+            self._zapi.usermacro.create({'hostid': host_id, 'macro': macro_name, 'value': macro_value, 'type': macro_type})
             self._module.exit_json(changed=True, result="Successfully added host macro %s" % macro_name)
         except Exception as e:
             self._module.fail_json(msg="Failed to create host macro %s: %s" % (macro_name, e))
 
     # update host macro
-    def update_host_macro(self, host_macro_obj, macro_name, macro_value):
+    def update_host_macro(self, host_macro_obj, macro_name, macro_value, macro_type):
         host_macro_id = host_macro_obj['hostmacroid']
-        if host_macro_obj['macro'] == macro_name and host_macro_obj['value'] == macro_value:
+        try:
+            host_macro_obj['type']
+        except IndexError:
+            host_macro_obj['type'] = 0
+        if host_macro_obj['macro'] == macro_name and host_macro_obj['value'] == macro_value and host_macro_obj['type'] == macro_type:
             self._module.exit_json(changed=False, result="Host macro %s already up to date" % macro_name)
         try:
             if self._module.check_mode:
                 self._module.exit_json(changed=True)
-            self._zapi.usermacro.update({'hostmacroid': host_macro_id, 'value': macro_value})
+            self._zapi.usermacro.update({'hostmacroid': host_macro_id, 'value': macro_value, 'type': macro_type})
             self._module.exit_json(changed=True, result="Successfully updated host macro %s" % macro_name)
         except Exception as e:
             self._module.fail_json(msg="Failed to update host macro %s: %s" % (macro_name, e))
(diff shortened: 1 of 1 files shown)


D3DeFi (commented):
Thank you for the patch. This looks good, but I would like to make one requests:

We are trying to migrate away from the numerical values where possible. This module should have `type=[ text, secret, vault ]` instead of `type=[ 0 1 2]`. Can you please implement that? Change would be required in DOCUMENTATION, argument_spec and then some hard-coded mapping around line 218 that will be converting str to int.

Same was done in https://github.com/ansible-collections/community.zabbix/blob/main/plugins/modules/zabbix_host.py#L317, unfortunately at the time there were no vault macros.

icez:
Thank you for pointing out the example. I've updated the code to represent that as per your suggested. 

Please check if there's anything I'm missing.

D3DeFi (approved):

D3DeFi:
> 
> 
> Thank you for pointing out the example. I've updated the code to represent that as per your suggested.
> 
> Please check if there's anything I'm missing.

Looks good now, will let CI pass and then merge :)

D3DeFi:
Oh, one more thing. Would you be willing to add some tests to the tests/integration/targets/test_zabbix_hostmacro? Also please add changelog fragment to the changelog/fragments/456-zabbix-hostmacro-secret-vault.yml with content like:
```
minor_changes:
- zabbix_hostmacro - now supports creating macros of type secret and vault
```

D3DeFi (changes requested):
Seems these changes are breaking other Zabbix versions. Please have a look at how you can test those locally before submitting a patch:

https://github.com/ansible-collections/community.zabbix/blob/main/CONTRIBUTING.md#testing-zabbix-modules

my guess is that problem may be somewhere in update_macro conditions as things seems to get kinda complicated there

plugins/modules/zabbix_hostmacro.py
```diff
@@ -123,24 +134,36 @@ def get_host_macro(self, macro_name, host_id):
...
     # update host macro
-    def update_host_macro(self, host_macro_obj, macro_name, macro_value):
+    def update_host_macro(self, host_macro_obj, macro_name, macro_value, macro_type):
         host_macro_id = host_macro_obj['hostmacroid']
-        if host_macro_obj['macro'] == macro_name and host_macro_obj['value'] == macro_value:
-            self._module.exit_json(changed=False, result="Host macro %s already up to date" % macro_name)
+        if host_macro_obj['macro'] == macro_name:
+            if LooseVersion(self._zbx_api_version) >= LooseVersion('5.0'):
+                # no change only when macro type == 0. when type = 1 or 2 zabbix will not output value of it.
+                if host_macro_obj['type'] == 0 and macro_type == 0 and host_macro_obj['value'] == macro_value:
```
miettal, line 155: It seems CI failing indicate "wrong decetion of changing".
I guess type of host_macro_obj['type'] is string, so this condition will be false always.

```
TASK [test_zabbix_hostmacro : assert that nothing has been changed] ************
fatal: [testhost]: FAILED! => {
    "assertion": "not zbxhmacro_existing is changed",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

icez: @miettal Thank you for help me pointing out. I've been quite busy with my work recently and should be able to submit a fix in a few days.

D3DeFi: There is also something wrong with integration testing with Zbx 5.0+:
https://github.com/ansible-collections/community.zabbix/pull/456/checks?check_run_id=3590366728#step:7:2649

Seems like module is not idempotent once macro `type` stirs into the mix

D3DeFi: You can even test locally if you move you collection to ~/ansible_collections/community/zabbix and then:
```bash
export zabbix_version=5.0
docker-compose up -d
ansible-test integration -v test_zabbix_hostmacro
docker-compose down
```
which is waaay quicker then relying on Github Actions.

D3DeFi:
> Hello,
> 
> I've made a fix for idempotent on host macro update. But there's one case when macro type is 'secret' or 'vault' and 'force' is yes, which will always cause ansible to update the macro as the zabbix API doesn't return the macro value of it so that ansible don't know the previous variable and will always set the value for it. Is it fine for this case?

Yes, this is acceptable scenario. We will never have an option to be idempotent with secret/vault macros as Zabbix simply exposes them as write-only. Feel free to skip the test for secret/vault macro idempotency

D3DeFi:
Thank you :)

andreicornea:
Hello! I found a bug regarding this feature while using community.zabbix version 1.5.1 and Zabbix version 5.0.19. It looks like the first time you are creating a global usermacro as a secret it looks fine, the secret is actually a secret as it is displayed with "******" but if you rerun the task with the same configurations it turns out to be changed to plain text instead of the "******". Anyone else encountered this issue? Shall I open a separate issue ticket?

BGmot:
Hi @andreicornea,
please confirm the issue exists in the latest community.zabbix version (we can't fix anything for 1.5.1) and if the problem exists then open a new issue.

PR #457, opened 2021-09-13
win zabbix2 agent var re-use, template log path

errcmd:
##### SUMMARY

Add re-use of zabbix_win_install_dir var in other vars ( exclude bulk vars override in case changing installation dir )
Add if/else statement depends on os for agent2 log path 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zabbix_agent

##### ADDITIONAL INFORMATION

```paste below

```


dj-wasabi:
Thanks! 👍

PR #477, opened 2021-10-08
feat: Update roles to use collections for Ansible 2.10

john-p-potter:
##### SUMMARY
Most of the roles do not function properly when using Ansible 2.10 due to the transition of modules into separate collections.  For example, `seboolean` is not part of the `ansible.posix` collection, and this module is used in nearly every role.  Likewise, there are modules such as `mysql_` and `win_` that are now par of `community.mysql/windows` collections.  

I have updated the roles to call modules from their respective collections and updated the documentation for each role (and the overall package) to note what is required.

Additionally, I made a minor fix to the documentation of the `zabbix_template` module, correcting the example for importing a .xml template.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_agent role
zabbix_proxy role
zabbix_server role
zabbix_web role
zabbix_template module

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,14 @@ For the majority of modules, however, you can get away with just:
 ```bash
 pip install zabbix-api
 ```
+#### Ansible 2.10 and higher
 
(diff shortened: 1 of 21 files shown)


dj-wasabi:
I like this PR. Thank you!
But I think merging this PR would make Ansible 2.10 the minimum version that can be used with this collection/role right? Are there still people "out there" running lower versions of Ansible?

PR #480, opened 2021-10-14
fix: Add no_log to PSK encryption

john-p-potter:
##### SUMMARY

The no_log attribute can be used to prevent output being sent to stdout or stderr as Ansible does by default.  By setting no_log to log nothing below ansible_verbosity level 3, we allow anyone who wants to see that output still get it with -vvv or -vvvv, but prevent it from being displayed in the case of an error.  

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_agent role

##### ADDITIONAL INFORMATION

```paste below
-v BEFORE:
TASK [zabbix_agent : AutoPSK | Save existing TLS PSK secret] *****************************************************************
ok: [host] => {"ansible_facts": {"zabbix_agent_tlspsk_read": "aAFb4bE09958eF13A26FfE0d4Ebb5E38eE2FDDc7cEdafd07acC1C5df49e2AA56"}, "changed": false}

-v AFTER:
TASK [zabbix_agent : AutoPSK | Save existing TLS PSK secret] *****************************************************************
ok: [host] => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}

default (0) verbosity with a fail BEFORE:
TASK [zabbix_agent : AutoPSK | Read existing TLS PSK identity file] **********************************************************************************************************************************
fatal: [host]: FAILED! => {"changed": false, "content": "NTEuNzkuOTAuMTIzXzFkMDA=", "encoding": "base64", "failed_when_result": true, "source": "/etc/zabbix/tls_psk_auto.identity"}

default (0) verbosity with a fail AFTER:
TASK [zabbix_agent : AutoPSK | Read existing TLS PSK identity file] **********************************************************************************************************************************
fatal: [host]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}
```

### DIFF
--- a/roles/zabbix_agent/tasks/tlspsk_auto.yml
+++ b/roles/zabbix_agent/tasks/tlspsk_auto.yml
@@ -11,21 +11,25 @@
   register: zabbix_agent_tlspsk_base64
   become: yes
   when: zabbix_agent_tlspskcheck.stat.exists
+  no_log: "{{ ansible_verbosity < 3 }}"
 
 - name: AutoPSK | Save existing TLS PSK secret
   set_fact:
     zabbix_agent_tlspsk_read: "{{ zabbix_agent_tlspsk_base64['content'] | b64decode | trim }}"
   when: zabbix_agent_tlspskcheck.stat.exists
+  no_log: "{{ ansible_verbosity < 3 }}"
 
 - name: AutoPSK | Use existing TLS PSK secret
   set_fact:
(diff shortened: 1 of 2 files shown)


gundalow:
recheck

dj-wasabi:
Thanks @john-p-potter  for creating the PR and thank you @gundalow for rerunning the failing check. 👍

Issue #460, opened 2021-09-17
Updating PSK is broken with Zabbix 5.4.2


manuelschneider:
##### SUMMARY

Since Zabbix 5.4.0 PSK fields are writeonly (also in the API), see

https://github.com/ansible-collections/community.zabbix/blob/2f1898f36ba2c8751e84c2892aa04fd003e52f6b/plugins/modules/zabbix_host.py#L834

That means, the key isn't returned, so you cannot check it for changes - it is empty. In your check if something has changed at
  https://github.com/ansible-collections/community.zabbix/blob/2f1898f36ba2c8751e84c2892aa04fd003e52f6b/plugins/modules/zabbix_host.py#L834
you're not checking if the psk *should* exist, but doesn't. That means, changing the PSK key won't be detected as neccessary change.

I'd recommend to also check if tls_psk is not NONE and not 'tls_psk' in host. Maybe you wanna check if this is also required for other properties.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix_host

PR #481, opened 2021-10-19
fix: zabbix 5.4 psk

tuxmartin:
FIX: https://github.com/ansible-collections/community.zabbix/issues/460

### DIFF
--- a/plugins/modules/zabbix_host.py
+++ b/plugins/modules/zabbix_host.py
@@ -836,7 +836,7 @@ class Host(ZabbixBase):
             if tls_psk is not None and 'tls_psk' in host:
                 if host['tls_psk'] != tls_psk:
                     return True
-￼
+
         if tls_issuer is not None and 'tls_issuer' in host:
             if host['tls_issuer'] != tls_issuer:
                 return True


dj-wasabi:
@D3DeFi No blockers from my side, but would be nice I guess to have #502 and #505  merged as well. As an FYI, in the `upgrade.md` file there is a mention of 1.4.1, that should then also be updated with value 1.5.0 (As there is no 1.4.1 release).

D3DeFi:
> @D3DeFi No blockers from my side, but would be nice I guess to have #502 and #505 merged as well. As an FYI, in the `upgrade.md` file there is a mention of 1.4.1, that should then also be updated with value 1.5.0 (As there is no 1.4.1 release).

Sure, no problem. will do preps in the meantime and wait for the PRs :)

dj-wasabi:
> > @D3DeFi No blockers from my side, but would be nice I guess to have #502 and #505 merged as well. As an FYI, in the `upgrade.md` file there is a mention of 1.4.1, that should then also be updated with value 1.5.0 (As there is no 1.4.1 release).
> 
> Sure, no problem. will do preps in the meantime and wait for the PRs :)

Thanks bro! 👍

PR #489, opened 2021-10-20
Set the PostgreSQL version to use for testing

klando:
##### SUMMARY

PG13 is defined here instead of nothing, which in turn uses latest,
which is PG14 now, and PG14 does SHA instead of MD5 by default.

Thus the issue.

NOTE: the collection won't work with PostgreSQL 14 as is. Fixes are
required.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

molecule

##### ADDITIONAL INFORMATION

If it passes regression tests as expected, it should allow other PRs to re-run tests....

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@
 version: '3.7'
 services:
   zabbix-db:
-    image: postgres:latest
+    image: postgres:13
     environment:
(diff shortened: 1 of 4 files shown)


klando:
thank you for having handled that so quickly! you'll get more work to come :p

PR #491, opened 2021-10-20
Review of zabbix config templates (not those related to http)

klando:
##### SUMMARY

Update variable name and use the same decision:

each variable name use in the configuration templates is named from
`config_filename + _ + config_parameter_name|lower`

Old variables not removed in order to not produce regression.

##### ISSUE TYPE

- Imrovement  Pull Request

##### COMPONENT NAME

role zabbix _agent and zabbix_proxy

##### ADDITIONAL INFORMATION

I have defined the new variable from the old variables (`newvar: "{{ old_var }}"`). I hope it's enough to not have to bump major version ?

Looks like I managed to push all doc UPGRADE in a single commit, in the other PR about API review.

### DIFF
--- a/docs/UPGRADE.md
+++ b/docs/UPGRADE.md
@@ -85,7 +85,7 @@ Example of using the role in this collection::
 - hosts: zabbix-proxy
   roles:
     - role: community.zabbix.zabbix_proxy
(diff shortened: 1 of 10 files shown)


dj-wasabi:
Thank you! 👍

PR #492, opened 2021-10-20
Review parameters used to connect to API

klando:
##### SUMMARY

The variables names were a little bit mixed...
Define that all parameters to connect to API are prefixed by
`zabbix_api_` followed by the API parameter name (except for *_http_* where it looks really useless).

##### ISSUE TYPE

- Improvment Pull Request

##### COMPONENT NAME

roles zabbix_agent, zabbix_proxy and zabbix_web

##### ADDITIONAL INFORMATION

The behavior should remain intact.


dj-wasabi:
I like it, also the "deprecated" messages! 👍 
Was thinking about making some rework changes as well, will find some time for that so we can soon create a 2.0.0.

Thanks!

PR #495, opened 2021-10-22
Apply the same logic of renaming parameters

klando:
##### SUMMARY

I renamed some more parameters used when working with API.

##### ISSUE TYPE

- Improvment Pull Request

##### COMPONENT NAME

roles zabbix_agent and zabbix_proxy

##### ADDITIONAL INFORMATION

Working more with the collection, I wonder why `zabbix_agent` role is not named `zabbix_host`... It looks like it's based on the packages name, but using the concepts of the webUI (host instead of agent) might be interesting.


dj-wasabi:
Thanks! 👍

PR #517, opened 2021-11-03
Repository GPG checking

### LINKED ISSUE #511: Find better way to handle signing keys for Zabbix
##### SUMMARY

Each role has a `zabbix.yml` file in the `vars/` directory containing the version of Zabbix, the name of the OS and the signing key. It not maintainance friendly to update the file with new OS's with signing keys and maintain them over de 5 roles. Something smarter should either be used, or ....

##### ISSUE TYPE
- Feature Idea

dopice:
##### SUMMARY
As being mentioned in #511, for us to handle the signing keys for the packages is a pain. This is a proposal how it can be solved. I've also included some changes for how we add the repository to the server.

Fixes #511

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_server

##### ADDITIONAL INFORMATION
For Debian based distribution we add all keys instead of single out one per distribution.


dopice:
@dj-wasabi This is my initial proposal. I will do some modifications for EL based distributions as well so we can enable GPG checking by default.

dopice:
And now with EL as well

dopice:
Seems like I don't have to create any PR towards molecule! 😄

dj-wasabi:
Looks good to me. 👍 

I think we should rename `zabbix_repo_yum` to something similar as the deb`zabbix_server_repo_deb_url`, something like `zabbix_server_repo_rpm_url`.
We should also mention something in the documenation, I think it at least in the upgrade document that some of the properties are deprecated.

dopice:
Glad you like it! Sure, I will do some cleanup and implement the same for the other modules as well. Same thing goes with the documentation.
I suppose we also should note that by default it installs the latest _supported_ version. In case of CentOS 7, it's 5.0. Do you think that should be a part of this PR or should this be discussed more if it's a functionality we really want? If yes, then I suppose we should do the same for Debian based distributions as well.

dj-wasabi:
I think for now we should just focus on the gpg's. Let write an issue here in Github where we describe the functionality with regards to C7 and 5.0. Then we can also let people to weigh in their thoughts about this, I don't want to "hide" certain stuff that might be confusing (Why is 5.0 installed when I expect 5.4)..

dopice:
Let's see if this works... I've lost my test machines so haven't tested this locally. Suppose there will be some issues..!

evilhamsterman:
If you are working on revising key management, I would also recommend removing the usage of apt-key. That has been deprecated by Debian/Ubuntu due to the key being implicitly trusted for all repos. Even the Ansible Module state it should not longer be used.

The recommended way now is to download the keys to `/usr/share/keyrings` then the repo line can point directly to the key file specific to that repo E.G. `deb [signed-by=/usr/share/keyrings/deriv-archive-keyring.gpg] https://deriv.example.net/debian/ stable main` 

You can see more info here https://wiki.debian.org/DebianRepository/UseThirdParty

PR #518, opened 2021-11-05
Use templateid property for templates

mu1f407:
Fixes "Error -32500: Application error., No permissions to referred object or it does not exist!" error on Zabbix 5.0 when linking another template.

Currently this module is sending: `"templates": ["12345"]` which result in the above error on Zabbix 5.0. It worked on version 4.0. I didn't found any documented change in API, but it was already required before to send templateid property according to the documentation: https://www.zabbix.com/documentation/5.0/manual/api/reference/template/create

This commit changes request that it is sending: `"templates": [{"templateid": "12345"}]` which is correct way according to the documentation. This works fine on Zabbix 5.0.

Thanks.

### DIFF
--- a/plugins/modules/zabbix_template.py
+++ b/plugins/modules/zabbix_template.py
@@ -352,7 +352,7 @@ class Template(ZabbixBase):
                 continue
             else:
                 template_id = template_list[0]['templateid']
-                template_ids.append(template_id)
+                template_ids.append({'templateid': template_id})
         return template_ids
 
     def add_template(self, template_name, group_ids, link_template_ids, macros):


mu1f407:
I found this change breaks some other things. Closing for now. Needs to be done in a different way.

D3DeFi:
@mu1f407 thank you for spotting this. I wonder why our tests dont fail. I also wonder why our modules do even work :D This seems to be required even from zabbix 4.0, which we long support, but apparently we don't pass like `templats: [templateid: value]`

I will at least open issue based on this PR

PR #522, opened 2021-11-11
Feature: maintenance start time, default name

ironbishop:
##### SUMMARY
* added `start_time` parameter to create scheduled maintenances
* defined a default `name` for the maintenance

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/plugins/modules/zabbix_maintenance.py
+++ b/plugins/modules/zabbix_maintenance.py
@@ -312,7 +312,7 @@ def main():
         zabbix_host = "host"
 
     if not name:
-        name = f"Created at {now.strftime('%Y-%m-%d %H:%M')}"
+        name = "Created at " + now.strftime('%Y-%m-%d %H:%M')
 
     changed = False
 
@@ -323,7 +323,7 @@ def main():
         if start_time:
             try:
                 time_obj = datetime.datetime.strptime(start_time, '%Y-%m-%d %H:%M')
-            except:
+            except (Exception, ValueError) as e:
                 module.fail_json(msg="Failed to parse start_time: expected format is YYYY-mm-dd HH:MM")
         else:
             time_obj = now.replace(second=0)
--- a/plugins/modules/zabbix_maintenance.py
+++ b/plugins/modules/zabbix_maintenance.py
@@ -44,6 +44,12 @@ options:
             - Length of maintenance window in minutes.
         default: 10
         type: int
+    start_time:
+        description:
+            - Time the maintenance starts in format YYYY-mm-dd HH:MM
+            - If None the default start time is now()
+        default: None
+        type: str
     name:
         description:
             - Unique name of maintenance window.
@@ -74,7 +80,7 @@ extends_documentation_fragment:
 notes:
     - Useful for setting hosts in maintenance mode before big update,
       and removing maintenance window after update.
-    - Module creates maintenance window from now() to now() + minutes,
+    - Module creates maintenance window from start_time to start_time + minutes,
       so if Zabbix server's time and host's time are not synchronized,
       you will get strange results.
     - Install required module with 'pip install zabbix-api' command.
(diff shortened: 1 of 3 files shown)


plugins/modules/zabbix_maintenance.py
```diff
@@ -301,14 +311,24 @@ def main():
...
     if state == "present":
         if not host_names and not host_groups:
             module.fail_json(msg="At least one host_name or host_group must be defined for each created maintenance.")
 
-        now = datetime.datetime.now().replace(second=0)
-        start_time = time.mktime(now.timetuple())
+        if start_time:
+            try:
+                time_obj = datetime.datetime.strptime(start_time, '%Y-%m-%d %H:%M')
+            except:
```
D3DeFi, line 326: tox is not happy with except without explicit Exception definition. In this case I think this should be `ValueError`

plugins/modules/zabbix_maintenance.py
```diff
@@ -301,14 +311,24 @@ def main():
     else:
         zabbix_host = "host"
 
+    if not name:
+        name = f"Created at {now.strftime('%Y-%m-%d %H:%M')}"
```
D3DeFi, line 315: unfortunately, f-strings are supported starting form 3.6+, we need to also still support python2.7 until ansible 2.12 is deprecated:
```
 ERROR: Found 1 compile issue(s) on python 2.7 which need to be resolved:
ERROR: plugins/modules/zabbix_maintenance.py:315:61: SyntaxError: name = f"Created at {now.strftime('%Y-%m-%d %H:%M')}"
```

D3DeFi (changes requested):
Thank you for this patch, except the Python2.7 errors this looks good and I will be happy to merge this once those problems are fixed.

Can you please also add an changelog fragment to the changelogs/fragments?
```
minor_changes:
- zabbix_maintenance - now allows to set start time in the future via `start_time` parameter
```

plugins/modules/zabbix_maintenance.py
```diff
@@ -301,14 +311,24 @@ def main():
...
     if state == "present":
         if not host_names and not host_groups:
             module.fail_json(msg="At least one host_name or host_group must be defined for each created maintenance.")
 
-        now = datetime.datetime.now().replace(second=0)
-        start_time = time.mktime(now.timetuple())
+        if start_time:
+            try:
+                time_obj = datetime.datetime.strptime(start_time, '%Y-%m-%d %H:%M')
+            except (Exception, ValueError) as e:
```
D3DeFi, line 326:
```suggestion
            except (Exception, ValueError):
```

unused `e` is troubling tox linters

plugins/modules/zabbix_maintenance.py
```diff
@@ -44,11 +44,18 @@
...
+            - If None the default start_time is now()
+        default: None
+        type: str
     name:
         description:
             - Unique name of maintenance window.
-        required: true
+        required: false
         type: str
+        default: 'Created at now.strftime('%Y-%m-%d %H:%M')'
```
D3DeFi, line 58: Not sure what causes `ansible-doc -t module community.zabbix.zabbix_maintenance` fail, but I would guess that this may be a problem. Probably expects basic keywords like true/false or choices from other options.

D3DeFi (changes requested):
Neither `CHANGELOG.rst` or `changelogs/changelog.yml` can be adjusted by hand. 

Please move the changelog snippet to the changelog/fragments/ sub-directory. For example like one (replacing `bugfixes` for `minor_changes`) - https://github.com/ansible-collections/community.zabbix/blob/main/changelogs/fragments/523-templateid-property.yml

pyrodie18:
Closed due to lack of corrections. Please feel free to make the required changes and create a new PR.

PR #523, opened 2021-11-11
Use templateid property for linking of templates

### LINKED ISSUE #519: Wrong usage of template.create/update API calls?
##### SUMMARY
See #518

Apparently, we link templates in a wrong way?

```
"templates": ["12345"] which result in the above error on Zabbix 5.0. It worked on version 4.0. I didn't found any documented change in API, but it was already required before to send templateid property according to the documentation: https://www.zabbix.com/documentation/5.0/manual/api/reference/template/create

This commit changes request that it is sending: "templates": [{"templateid": "12345"}] 
```

I wonder why our zabbix_template module even works in actual state, since doc for 4.0 also notes using `templates: [templateid: val]`

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

mu1f407:
##### SUMMARY
Fixes "Error -32500: Application error., No permissions to referred object or it does not exist!" error on Zabbix 5.0+ when linking another template.

Currently this module is sending: `"templates": ["12345"]` which result in the above error during template.create API call on Zabbix 5.0 and newer. It worked correctly on version 4.0. There is none documented change in API, but it was already required before to send `templateid` property according to the documentation: https://www.zabbix.com/documentation/4.0/manual/api/reference/template/create

This PR changes request that it is sending: `"templates": [{"templateid": "12345"}]` which is correct way according to the documentation. This works fine on Zabbix 5.0.

It is implemented in a way that `get_template_ids` function returns list of dicts. Configuration dump and template delete still needs only list of IDs, so it is transformed to plain list of numbers for these.

A new test for template creating with linked template is included. This test currently fails on 5.0+ without changes in this PR.

Successfully tested against versions 4.0, 5.0 and 5.2. On 5.4 the template module tests currently fails due to another issue.

Fixes #519

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/523-templateid-property.yml
+++ b/changelogs/fragments/523-templateid-property.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - template - use templateid property when linking templates for ``template.create``
+    and ``template.update`` API calls.
--- a/plugins/modules/zabbix_template.py
+++ b/plugins/modules/zabbix_template.py
@@ -352,7 +352,7 @@ class Template(ZabbixBase):
                 continue
             else:
                 template_id = template_list[0]['templateid']
-                template_ids.append(template_id)
+                template_ids.append({'templateid': template_id})
         return template_ids
 
     def add_template(self, template_name, group_ids, link_template_ids, macros):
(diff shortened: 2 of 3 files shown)


D3DeFi (approved):
Awesome work @mu1f407 ! Thank you very much for this fix :)

PR #528, opened 2021-11-15
proposal for #515

### LINKED ISSUE #515: Role Zabbix Agent (linux): do not create PSK directory and file on false-ish key file value
##### SUMMARY

At the moment, the only way to not create the passive linux agent PSK key file is to have variable zabbix_agent2_tlspskfile left undefined.

This is sometimes a difficult requirement when inventory is complex.

My case: I'd like to set a default value in group_vars/all.yml and override this in other group_vars, however having the default value in group_vars gives me no way to reset it to undefined (as far as my ansible knowledge goes).

P.S.:
The templating of the conf file already has the logic to ignore defined yet fals'ish value

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

zabbix_agent (role)

##### ADDITIONAL INFORMATION

```yaml

```

shk3bq4d:
##### SUMMARY

Fixes #515

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
zabbix_agent role

### DIFF
--- a/roles/zabbix_agent/tasks/Linux.yml
+++ b/roles/zabbix_agent/tasks/Linux.yml
@@ -134,6 +134,7 @@
   become: true
   when:
     - zabbix_agent_tlspskfile is defined
+    - zabbix_agent_tlspskfile
     - not zabbix_agent2
 
 - name: "Create directory for PSK file if not exist (zabbix-agent2)"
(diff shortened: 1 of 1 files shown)


dj-wasabi:
Thanks! 👍

Issue #542, opened 2021-11-27
zabbix_user module requires a password on "disabled" authentication types


ragingpastry:
##### SUMMARY
Currently the zabbix_user Ansible module will require a password to be set on any new user that is a member of a usergroup using any authentication type other than "LDAP". I believe it should be required for "internal" authentication types only

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix_user module

##### ANSIBLE VERSION
N/A

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT / Zabbix Version
N/A

##### STEPS TO REPRODUCE

```yaml
- name: Create ReadOnly UserGroup
   community.zabbix.zabbix_usergroup:
       server_url: "{{ vars_server_url }}"
        login_user: Admin
        login_password: "{{ vars_zabbix_password }}"
        name: ReadOnly
        gui_access: disable
        rights: "{{ hostgroup_rights }}"
        validate_certs: False
        state: present

      - name: Create zabbix users
        community.zabbix.zabbix_user:
          server_url: "{{ vars_server_url }}"
          login_user: Admin
          login_password: "{{ vars_zabbix_password }}"
          alias: "opsgenie"
          name: "opsgenie"
          validate_certs: False
          usrgrps: "ReadOnly"
          state: present
          type: "Zabbix user"
```

##### EXPECTED RESULTS
The user can be created without a password because the authentication type for the usergroup is set to "disabled"

##### ACTUAL RESULTS
The password is required because "disabled" is not "LDAP" authentication type

PR #543, opened 2021-11-27
fix: zabbix_user require password only on internal

ragingpastry:
Modifies the logic in the zabbix_user module to require a password
only when GUI access is set to internal. A user in zabbix only
requires a password to be set when the authentication type is set
to internal

##### SUMMARY
Fixes: #542

Previously the zabbix_user module was enforcing a password on all authentication types except for LDAP. Zabbix however only requires a password to be set on a user when the authentication type is internal.

This change ensures that a user created with a usergroup that uses "internal" authentication has a password. Otherwise allow the user to be created without a password

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
zabbix_user module

##### ADDITIONAL INFORMATION

If somebody sees a better way to enable this please let me know!`

### DIFF
--- a/plugins/modules/zabbix_user.py
+++ b/plugins/modules/zabbix_user.py
@@ -306,13 +306,13 @@ class User(ZabbixBase):
         res = self._zapi.usergroup.get(params)
         if res:
             ids = [{'usrgrpid': g['usrgrpid']} for g in res]
-            # User can be created password-less only when all groups are LDAP
-            # Verify there are no groups with different access methods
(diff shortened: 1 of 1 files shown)


ragingpastry:
Strange...... Will look into why coverage dropped later tonight.

D3DeFi (approved):
Good catch, I guess this  somehow came to be when there were fewer authentication choices available :)

Thank you! :)

PR #546, opened 2021-11-29
Avoid using rich rules

### LINKED ISSUE #545: Zabbix Agent firewalld avoid rich rules
##### SUMMARY

Currently the Zabbix Agent Ansible uses rich rules to configure FirewalD. While the zone name is configurable, this only really works if the zone has a dedicated interface. Rich rules are basically a hack and do not allow FirewallD to be used as intended. Furthermore the current code can only use existing zones.

Instead of using rich rules, the code should configure the FirewallD zone if absent and then use the zones to filter by source and protocol. 

For example:
```yaml
- firewalld:
    zone: "{{ zabbix_agent_firewalld_zone }}"
    service: "zabbix-agent"
    state: enabled
    permanent: yes
    immediate: yes
  when: zabbix_agent_firewalld_enable

-  firewalld:
    zone: "{{ zabbix_agent_firewalld_zone }}"
    source: "{{ zabbix_agent_firewalld_source }}"
    state: enabled
    permanent: yes
    immediate: yes
  when: zabbix_agent_firewalld_enable
```

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

zabbix-agent role

##### ADDITIONAL INFORMATION

When a server uses FirewallD as intended, there are multiple zones separated by source IP. In such a case, rich rules are not usable for any zone other than "public".

Instead you want the firewall configured something like this:
```
monitoring (active)
  target: default
  icmp-block-inversion: no
  interfaces:
  sources: 192.168.1.2
  services: zabbix-agent
  ports:
  protocols:
  forward: no
  masquerade: no
  forward-ports:
  source-ports:
  icmp-blocks:
  rich rules:
```
...so that it can coexist with other similar zones.

I will provide a pull request with the code I ended up using in my wrapper, however I've done not extensive checking if it plays nice with other functionality.

juresaht2:
##### SUMMARY

Use FirewallD as intended instead of hacking with rich rules.

Fixes #545 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ansible-agent role

##### ADDITIONAL INFORMATION

The existing functionality should remain unchanged, except for better coexistence with other FirewallD zones.


dj-wasabi:
As mentioned in #545 I would not like to remove the firewalld functionality and thus closing this PR. But still a thank you for your work.

PR #570, opened 2021-12-12
Fix autopsk

### LINKED ISSUE #564: AutoPSK broken for existing PSK
##### SUMMARY

AutoPSK was broken in PR #537.  `tlspsk_auto_common.yml` has the wrong variable being checked on lines 31 and 43.  `zabbix_agent2_tlspskidentity_check.stat.exists` should be `zabbix_agent_tlspskidentity_check.stat.exists`. 
- Bug Report

##### COMPONENT NAME

Zabbix Agent Role
##### ANSIBLE VERSION

```paste below
Any
```

Linux

##### STEPS TO REPRODUCE

Run the Zabbix Agent role on a system with an existing PSK.

##### EXPECTED RESULTS

It uses the existing PSK

##### ACTUAL RESULTS

It detects the existing PSK, but when it gets to the `AutoPSK | Use existing TLS PSK identity` stage, it fails as it is checking for the wrong variable.

john-p-potter:
##### SUMMARY

Fixed the autopsk tasks for Zabbix Agent.  `tlspsk_auto_common.yml` had the wrong variable being checked on lines 31 and 43. `zabbix_agent2_tlspskidentity_check.stat.exists` should be `zabbix_agent_tlspskidentity_check.stat.exists`.
Fixes #564 
Added testing of autopsk for the role.  Reconfigured the Workflow a bit for molecule testing to add a lot more tests instead of just running one test - now every distro is with the agent (default scenario), agent2, autopsk with agent, and autopsk with agent2.  
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Zabbix Agent Role

##### ADDITIONAL INFORMATION

Thanks to @ragingpastry for help with getting the testing updated.

```paste below
Prior to this fix, the role fails when run against a system with an existing PSK.  It detects the existing PSK, but when it gets to the AutoPSK | Use existing TLS PSK identity stage, it fails as it was checking for the wrong variable.
```


dj-wasabi (approved):
lgtm, thanks 👍

PR #601, opened 2022-01-04
fix: corrects invalid usergroup query

### LINKED ISSUE #578: zabbix_user no longer working with default authentication
##### SUMMARY
Patch #542 broke creating users that have authentication defaulting to the global setting (if that global setting is internal authentication).

```
Possible values:
0 - (default) use the system default authentication method;
1 - use internal authentication;
2 - use LDAP authentication;
3 - disable access to the frontend.
```

image

I have module failing on:

```yaml
failed: [zabbix -> localhost] (item=user1) => changed=false 
  msg: 'Failed to create user tjd: (''Error -32602: Invalid params., Incorrect value for field "passwd": cannot be empty. while sending {"jsonrpc": "2.0", "method": "user.create", "params": {"alias": "user1", "usrgrps": [{"usrgrpid": "15"}], "theme": "dark-theme", "type": "1"}, "auth": "XYZ", "id": 4}'', -32602)'
```

So for me the snippet from the #542 
```python
internal_auth = bool([g for g in res if g['gui_access'] == '1'])
```

results in `False` as `gui_access == '0'` when defaulting the authentication system-wide to internal.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
plugins/modules/zabbix_user

ragingpastry:
##### SUMMARY

fixes #578 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_user.py

##### ADDITIONAL INFORMATION
Refer to the issue for context

### DIFF
--- a/plugins/modules/zabbix_user.py
+++ b/plugins/modules/zabbix_user.py
@@ -297,6 +297,15 @@ class User(ZabbixBase):
 
         return 'alias'
 
+    def get_default_authentication(self):
+        auth = self._zapi.authentication.get({'output': 'extend'})
+        if auth.authentication_type == 0:
+            return "internal"
+        elif auth.authentication_type == 1:
+            return "LDAP"
+        else:
+            self._module.fail_json(msg="Failed to query authentication type. Unknown authentication type %s" % auth)
+
     def get_usergroups_by_name(self, usrgrps):
         params = {
             'output': ['usrgrpid', 'name', 'gui_access'],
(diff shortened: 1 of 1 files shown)


D3DeFi (approved):
This looks good to me. Nice patch :) Will not merge since it is still in WIP

ragingpastry:
I still need to look at the code coverage misses and I'd like to test it in a real environment (at the very least locally)

ragingpastry:
It might be better to get this merged in so that people can start using it. I can devote some time over this weekend to get this set up with proper LDAP tests so that we can make sure we don't break in the future.

Thoughts @D3DeFi ?

D3DeFi:
> It might be better to get this merged in so that people can start using it. I can devote some time over this weekend to get this set up with proper LDAP tests so that we can make sure we don't break in the future.
> 
> Thoughts @D3DeFi ?

I agree, feel free to merge it when you are ready :)

PR #624, opened 2022-02-04
Support for multiple agent config Alias entries

Henderson42:
##### SUMMARY
Required setup of multiple aliases, existing code gives an incorrect syntax that prevents the agent from starting. Edited this code locally to allow it to input multiple aliases in supported syntax.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_agent template

##### ADDITIONAL INFORMATION
Original template:
Can't define multiple alias values in yaml, passing an array passes the array into zabbix_agentd.conf wholesale which prevents the agent from starting due to invalid syntax.

Proposed changed to template:
Simple fix applied to zabbix_agent2.conf.j2 and zabbix_agentd.conf.j2 allows you to specify an array of aliases you'd like to include in your configuration.

##### USAGE
`zabbix_agent_zabbix_alias: ["alias1[*]:alias","alias2.info[*]:alias.info[*]","alias3.info[*]:alias.info[*]"]`

### DIFF
--- a/roles/zabbix_agent/templates/zabbix_agent2.conf.j2
+++ b/roles/zabbix_agent/templates/zabbix_agent2.conf.j2
@@ -312,7 +312,9 @@ PersistentBufferFile={{ zabbix_agent2_persistentbufferfile }}
 # Range:
 # Default:
 {% if zabbix_agent2_zabbix_alias is defined and zabbix_agent2_zabbix_alias %}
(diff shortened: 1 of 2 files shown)


Henderson42:
Hadn't noticed that this was fixed since I fixed it locally. My bad.

PR #626, opened 2022-02-07
Fix package name [Ready to merge]

ChoiZ:
Update package name "mysql-client" by "default-mysql-client"
Or maybe we can update by "mariadb-client" also.

##### SUMMARY

Update the package mysql-client because it's failed on debian 11. This package is not found.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

zabbix_server tasks debian

### DIFF
--- a/roles/zabbix_server/tasks/Debian.yml
+++ b/roles/zabbix_server/tasks/Debian.yml
@@ -206,7 +206,7 @@
 - name: "Debian | Install Mysql Client package"
   apt:
     name:
-      - mysql-client
+      - default-mysql-client
       - "{{ zabbix_python_prefix }}-mysqldb"
     state: present
   environment:


ragingpastry:
Hmm I wonder why this isn't caught in CI...

ragingpastry:
I think we need to change up the logic here a bit. It looks like this was originally coded without debian11 in mind. 

On buster we specifically install `mariadb-client` instead of `mysql-client`, then for all releases that aren't buster we attempt to install `mysql-client`. This works for Debian 9, but not for 11. It also looks like Debian 9 has `default-mysql-client` available as well.

I am not too sure what the differences in the packages are here, or why we install `mariadb-client` on buster specifically. 

Perhaps we could change all of these to `default-mysql-client`?

dj-wasabi:
Hi @ChoiZ 

I would suggest to go for it. We can then verify with the tests if all goes right and then merge it. 👍

ChoiZ:
Someone can merge it please? :100:

dj-wasabi:
Thanks!

(Sorry, my bad I could have merged it yesterday (to much on my mind)!)

ChoiZ:
> Thanks!
> 
> (Sorry, my bad I could have merged it yesterday (to much on my mind)!)

Thank you!

Issue #627, opened 2022-02-07
Add module for autoregistration


masa-orca:
##### SUMMARY

May I add new module which manage autoregistration?
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

- zabbix_autoregistration

##### ADDITIONAL INFORMATION

I have tried to create a module to manage [autoregistration](https://www.zabbix.com/documentation/5.0/en/manual/api/reference/autoregistration/object).
However, I forgot asking you that you are welcoming to this contribute or not.
My commit is [here](https://github.com/masa-orca/community.zabbix/commit/6242e000fd4e7a95acacb791c978d374b1da5f03). If you have time, please give me advices!!

```yaml

```

D3DeFi:
Hi @masa-orca, this is definitely very nice and clean solution. I can see autoregistration being added to the zabbix somewhere around 4.4, so we should have added it way sooner. There are also other API endpoints like authentication, setting, etc.. that may also be a good module contributions.

+1 from me, feel free to open PR

@ragingpastry @rockaut what do you guys think?

ragingpastry:
Definitely open up a PR!

rockaut:
Wait, have you had a look at the action module? This should be used for auto registration.

D3DeFi:
My first though exactly, then I did more digging and noticed that is related to the settings -> autoregistration, not actions -> autoregistration, so it makes sense to have modules that will configure zabbix settings. I am only worried if it will not create some confussion among users. Bc if I would be searching for configuring autoregistration action in the future, my first bet would be zabbix_autoregistration module an not zabbix_action

Any ideas there?

rockaut:
Oh, hm. Then we should maybe do something like zabbix_settings_autoreg? I'm on vacation but i will have a deeper look at it tonight.

masa-orca:
I apologize for confusing you.
Zabbix provided some api's for updating settings(e.g. authentication). In addition, from version 6.0, Zabbix adds [setting](https://www.zabbix.com/documentation/devel/en/manual/api/reference/settings/object) api. However, it is not for authentication and autoregistration but for GUI settings. Therefore, I think we need to divide autoregistration and auth and some setting modules.
If we use TLS with PSK for connection between server and agents, we need to use this module and update zabbix_agent.conf before creating actions for autoregistration.

D3DeFi:
[...]

PR #630, opened 2022-02-13
Add autoregistration module

masa-orca:
##### SUMMARY

- Add zabbix_autoregistration module (Update only).
- Add tests about zabbix_autoregistration module.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

zabbix_autoregistration

##### ADDITIONAL INFORMATION

I create PR for  a module which mentioned at a [issue](https://github.com/ansible-collections/community.zabbix/issues/627).
I am not good at English. Please commnet if I need to change words and grammars!


rockaut:
Please see my last comment on the issue. Sorry for late response.

This function is a part of zabbix action already.

Please have a look at this and feel free to ask us if you need help.

rockaut:
Sorry,hit the wrong button

plugins/modules/zabbix_autoregistration.py
```diff
@@ -0,0 +1,182 @@
...
+            - Type of allowed incoming connections for autoregistration.
+            - Choose from C(unsecure), C(tls_with_psk) or both.
+        type: list
+        elements: str
+        required: true
+    tls_psk_identity:
+        description:
+            - TLS connection use this PSK identity string.
+            - The PSK identity string will be transmitted unencrypted over the network. Therefore, you should not put sensitive information.
+            - It is required if you set C(tls_with_psk) to I(tls_accept) and current tls_accept is C(unsecure).
```
D3DeFi, line 39:
```suggestion
            - TLS connection uses this PSK identity string.
            - The PSK identity string will be transmitted unencrypted over the network. Therefore, you should not put any  sensitive information here.
            - This setting requires I(tls_accept=tls_with_psk) if current value of I(tls_accept) is C(unsecure).
```

plugins/modules/zabbix_autoregistration.py
```diff
@@ -0,0 +1,182 @@
...
+        type: str
+    tls_psk:
+        description:
+            - TLS connection use this PSK value.
+            - It is required if you set C(tls_with_psk) to I(tls_accept) and current tls_accept is C(unsecure).
```
D3DeFi, line 44:
```suggestion
            - TLS connection uses this PSK value.
            - This setting requires I(tls_accept=tls_with_psk) if current value of I(tls_accept) is C(unsecure)
```

Does the 2nd line mean that `tls_accept=tls_with_psk` is only required when you updating from `tls_accept=unsecure`, but can be omitted when re-running multiple times without module doing any changes itself?

plugins/modules/zabbix_autoregistration.py
```diff
@@ -0,0 +1,182 @@
...
+    login_password: secret
+    tls_accept: unsecure
+'''
+
+RETURN = '''
+msg:
+    description: The result of the operation
+    returned: success
+    type: str
+    sample: 'Successfully update autoregistration'
```
D3DeFi, line 81:
```suggestion
    sample: 'Successfully updated global autoregistration setting'
```

plugins/modules/zabbix_autoregistration.py
```diff
@@ -0,0 +1,182 @@
...
+
+import ansible_collections.community.zabbix.plugins.module_utils.helpers as zabbix_utils
+
+
+class Autoregistration(ZabbixBase):
+    def __init__(self, module, zbx=None, zapi_wrapper=None):
+        super(Autoregistration, self).__init__(module, zbx, zapi_wrapper)
+        self.existing_data = None
+        if LooseVersion(self._zbx_api_version) < LooseVersion('4.4.0'):
+            module.fail_json(msg="This module unsuport Zabbix %s" % self._zbx_api_version)
```
D3DeFi, line 98:
```suggestion
            module.fail_json(msg="This module doesn't support Zabbix versions lower than %s" % self._zbx_api_version)
```

plugins/modules/zabbix_autoregistration.py
```diff
@@ -0,0 +1,182 @@
...
+        type: str
+    tls_psk:
+        description:
+            - TLS connection use this PSK value.
+            - It is required if you set C(tls_with_psk) to I(tls_accept) and current tls_accept is C(unsecure).
+        type: str
+
+notes:
+    - Only Zabbix >= 4.4 is supported.
+    - This module returns state of changed is true when you set values to I(tls_psk_identity) and I(tls_psk).
```
D3DeFi, line 49:
```suggestion
    - This module returns changed=true when any value is set in I(tls_psk_identity) or I(tls_psk) as Zabbix API will not return any sensitive information back for module to compare.
    - Please note that this module configures B(global Zabbix Server settings). If you want to create autoregistration action so your hosts can automatically add themselves to the monitoring have a look at M(community.zabbix.zabbix_action).
```

D3DeFi (changes requested):
Thank you for this contribution! :+1: :rocket: 

I took a liberty and corrected some of the grammar. Please see my suggestions. 

Also please add YAML drop-in file to changelog/fragments/630-zabbix-autoreg.yml with content like:
```
minor_changes:
- new module zabbix_autoregistration for configuring global autoregistration settings in Zabbix Server's Settings section of GUI
```
Otherwise very nice contribution

plugins/modules/zabbix_autoregistration.py
masa-orca: I tried it. However, I could not come up with best way to get last task's arguments.
In addition, I think these values must be not logged.

masa-orca:
@D3DeFi Thank you for your commens! and I have reflected your comments. Would you check my new commit?

D3DeFi (approved):
Very nice contribution :) +1 for merge.

@rockaut please decide which name we wsnt for the module and if this is fine with you, feel free to merge

rockaut:
I haven't tested it personally - but it looks great for me. Thanks you @masa-orca and sorry for me screwing it up with the typo. +1 from me too.

dj-wasabi:
Will merge it, as it seems that both @D3DeFi and @rockaut  agree. 😄

D3DeFi:
Hmm, @masa-orca care to open follow-up PR for renaming? :D

dj-wasabi:
Ok, my bad. I will try to do that somewhere in the next few days. No need for @masa-orca to do that.

PR #639, opened 2022-02-21
Fix zabbix_globalmacro module

masa-orca:
##### SUMMARY

- zabbix_globalmacro  module changes type to text from secret and vault. Fix them.
- The value parameter must be not logged for the value which's type is secret.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_globalmacro


D3DeFi (approved):
Looks good to me. LGTM

PR #683, opened 2022-05-02
Fixes for Zabbix 6.0

BGmot:
##### SUMMARY
Many modules currently fail with Zabbix 6.0.
This PR makes all modules correctly work with 6.0 with backward compatibility to previous Zabbix versions.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/zabbix_action.py
plugins/modules/zabbix_maintenance.py
plugins/modules/zabbix_mediatype.py
plugins/modules/zabbix_proxy.py
plugins/modules/zabbix_service.py

### DIFF
--- a/plugins/module_utils/helpers.py
+++ b/plugins/module_utils/helpers.py
@@ -129,7 +129,7 @@ def helper_compare_lists(l1, l2, diff_dict):
         return diff_dict
     for i, item in enumerate(l1):
         if isinstance(item, dict):
-            for i2, item2 in enumerate(l2):
+            for item2 in l2:
                 diff_dict2 = {}
                 diff_dict2 = helper_compare_dictionaries(item, item2, diff_dict2)
                 if len(diff_dict2) == 0:
--- a/plugins/module_utils/helpers.py
+++ b/plugins/module_utils/helpers.py
@@ -129,8 +129,13 @@ def helper_compare_lists(l1, l2, diff_dict):
         return diff_dict
     for i, item in enumerate(l1):
         if isinstance(item, dict):
-            diff_dict.insert(i, {})
-            diff_dict[i] = helper_compare_dictionaries(item, l2[i], diff_dict[i])
+            for i2, item2 in enumerate(l2):
+                diff_dict2 = {}
+                diff_dict2 = helper_compare_dictionaries(item, item2, diff_dict2)
+                if len(diff_dict2) == 0:
+                    break
+            if len(diff_dict2) != 0:
+                diff_dict.insert(i, item)
         else:
             if item != l2[i]:
                 diff_dict.append(item)
--- a/tests/integration/targets/test_zabbix_action/tasks/main.yml
+++ b/tests/integration/targets/test_zabbix_action/tasks/main.yml
@@ -335,6 +335,13 @@
 
   - assert:
       that: zbxaction_conditions_reorder.changed is sameas True
+    when:
+      - zabbix_version is version('6.0', '<')
+
+  - assert:
+      that: zbxaction_conditions_reorder.changed is sameas False
+    when:
+      - zabbix_version is version('6.0', '>=')
 
   - name: test - update action with multiple conditions by removing one condition
     zabbix_action:
--- a/tests/integration/targets/test_zabbix_action/tasks/main.yml
+++ b/tests/integration/targets/test_zabbix_action/tasks/main.yml
@@ -334,14 +334,7 @@
     register: zbxaction_conditions_reorder
 
   - assert:
-      that: zbxaction_conditions_reorder.changed is sameas True
-    when:
-      - zabbix_version is version('6.0', '<')
-
-  - assert:
       that: zbxaction_conditions_reorder.changed is sameas False
-    when:
-      - zabbix_version is version('6.0', '>=')
 
   - name: test - update action with multiple conditions by removing one condition
     zabbix_action:
@@ -1261,10 +1254,6 @@
           - Admin
   ignore_errors: true
   register: zbxaction_application
-
-- assert:
-    that: zbxaction_delete.changed is sameas True
-  when: zabbix_version is version('5.4', '<')
 
 - assert:
     that:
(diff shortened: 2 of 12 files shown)


dj-wasabi:
Looks good @BGmot 👍 

As long as the `[WIP]` is in the title, I won't do anything btw.. So when you think you are happy to have it merged, please remove the `[WIP]`.. Thank you for all the work you do! 👍

BGmot:
Removed [WIP]. Thanks for your help!

D3DeFi:
I would like to read through this first tomorrow so I better understand all of the changes and then we can merge

plugins/module_utils/helpers.py
```diff
@@ -129,8 +129,13 @@ def helper_compare_lists(l1, l2, diff_dict):
...
         if isinstance(item, dict):
-            diff_dict.insert(i, {})
-            diff_dict[i] = helper_compare_dictionaries(item, l2[i], diff_dict[i])
+            for i2, item2 in enumerate(l2):
+                diff_dict2 = {}
+                diff_dict2 = helper_compare_dictionaries(item, item2, diff_dict2)
+                if len(diff_dict2) == 0:
+                    break
+            if len(diff_dict2) != 0:
+                diff_dict.insert(i, item)
```
D3DeFi, line 138: I am having trouble understanding this correctly. Maybe my brain is not working optimally this morning :rofl: 

So previously, if we found dict as an element of list, we compared it with the item in the list2 at the same position, guessing that it would be a dict as well.

The new logic will, for each item that is a dict in list1, iterate whole dict2 and compare the item1 (dict) with everything in list2 no matter what that object would be (number/string for example). If it finds a match `len(diff_dict2) == 0` then it doesn't continue anymore. Wouldn't this require something like this?

```suggestion
            for i2, item2 in enumerate(l2):
                diff_dict2 = {}
                if isinstance(item2, dict):
                    diff_dict2 = helper_compare_dictionaries(item, item2, diff_dict2)
                    if len(diff_dict2) == 0:
                        break
            if len(diff_dict2) != 0:
                diff_dict.insert(i, item)
```

Also the enumeration index `i2` seems to be unused at all and can be omitted.

tests/integration/targets/test_zabbix_action/tasks/main.yml
```diff
@@ -334,7 +334,7 @@
     register: zbxaction_conditions_reorder
 
   - assert:
-      that: zbxaction_conditions_reorder.changed is sameas True
+      that: zbxaction_conditions_reorder.changed is sameas False
```
D3DeFi, line 337: Lol I wonder how this was not failing tests...

tests/integration/targets/test_zabbix_user/tasks/for_zabbix_54_higher.yml
```diff
@@ -10,7 +10,7 @@
     surname: test
     usrgrps:
       - Guests
-    passwd: password
+    passwd: G$jd_79!jw
```
D3DeFi, line 13: Not sure if you wanted to commit this (password)?

Also lets not commit change of `lang: ` parameter or lets unify it at `en_US`. I don't see reason for this change

D3DeFi (changes requested):
What an amazing patch this is! Really good job @BGmot , thank you! 

Can you please check my comments and help me understand some of the logic I pin-pointed? I think this is almost ready to be merged. And there are even tests! <3

Can you please also drop short changelog fragments to changelogs/fragments/683-zbx60.yml? https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment

plugins/module_utils/helpers.py
BGmot: Oh I spent couple of night with this problem. Imagine you work on a module locally without pushing anything to GitHub  and running integration test on this module over and over again, then at some point tests that were passing start to fail. Obviously I did not destroy Zabbix containers to speed up testing just manually deleting the resouce in Zabbix UI before re-running the test. I worked on zabbix_action.py and the test `- name: test - create new action with multiple conditions (reorder)` suddenly failed. I discovered that it was not returning re-ordered list anymore thus assertion `that: zbxaction_conditions_reorder.changed is sameas True` started to fail. Re-create containers and everything starts working again. I did not know what to do and was not sure how Python treats an order in lists/dictionaries and went to packet capture and guess what - after approximately 12-13 runs the order of elements in the list returned by Zabbix changes! I guess it is some kind of DB optimization optimization (plan? I am not a DB expert). I looked more closely at what we are trying to do and what we are expecting and it turns out that indeed the order of the elements in the list does not mater (at least I have not found any such case in this collection).
So the code is doing:
```
for every element in list l1
  iterate through all elements in list l2
    if any element in list l2 equals to element in list l1 then break this iteration and move to the next element in list l1
```
I am happy to add your `if isinstance(item2, dict):` but it might lead to unexpected results - if we compared a dict from l1 to let's say str in l2 (which should never happen!) we would not produce any diff and I'd rather see Python's exeption than silently ignoring this difference. Let me know.

And yes, I will definitely remote index i2.

tests/integration/targets/test_zabbix_action/tasks/main.yml
BGmot: Please see my previous comments. On fresh containers and first run it will not fail but later it will start failing. And we are talking about using this collection in real life so if order does not matter than the resource should not be indicated as 'changed' -)

tests/integration/targets/test_zabbix_user/tasks/for_zabbix_54_higher.yml
BGmot: This is just random password in a test and we need to replace default password with something complex as Zabbix 6.0 does not accept 'bad' passwords any longer (unless you change security settings) so to make tests work with Zabbix 6.0 we need to have som 'secure' password.

D3DeFi (approved):
Then I am ok with merging :) 

Or will you be pushing removal of `i2`?

BGmot:
I will be pushing removal of i2 and updating CHANGELOG but it's going to be later, stupid work distracts me from what I love doing -(

D3DeFi (approved):
Awesome! :) Once again, thank you very much for this amazing patch!

BGmot:
> Once again, thank you very much for this amazing patch!
Anytime. Thanks for your your guidance.

PR #689, opened 2022-05-12
Zabbix script module

### LINKED ISSUE #634: New module zabbix_scripts
##### SUMMARY
Since 5.4 Zabbix has a central scripts "repository" under Administration > Scripts with different types (actions, hosts, etc.) and accordingly also an API module for that.

It is crucial that we implement a new Ansible Module for this too. As without this module we can not provide support for Zabbix 5.4 and higher at all. For example: we need to create such a central script to reference it as an operation in a trigger action. Without this no trigger action with an script operation can be created. And my guess it, that this is only one example already.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
zabbix_scripts

##### ADDITIONAL INFORMATION
See https://www.zabbix.com/documentation/5.0/en/manual/api/reference/script

BGmot:
##### SUMMARY
A module added implementing Zabbix Scripts creation/modification.
Fixes #634 

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
zabbix_script

### DIFF
--- a/changelogs/fragments/685-zabbix-script.yml
+++ b/changelogs/fragments/685-zabbix-script.yml
@@ -1,4 +1,3 @@
 ---
 minor_changes:
-  - zabbix_script module added
-  - (https://github.com/ansible-collections/community.zabbix/issues/634)
+  - zabbix_script module added (https://github.com/ansible-collections/community.zabbix/issues/634)
--- a/changelogs/fragments/685-zabbix-script.yml
+++ b/changelogs/fragments/685-zabbix-script.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - zabbix_script module added
+  - (https://github.com/ansible-collections/community.zabbix/issues/634)
--- a/plugins/modules/zabbix_script.py
+++ b/plugins/modules/zabbix_script.py
@@ -10,7 +10,7 @@ __metaclass__ = type
 
 DOCUMENTATION = '''
 module: zabbix_script
-short_description: Gather information about Zabbix proxy
+short_description: Create/update/delete Zabbix scripts
 version_added: 1.7.0
 author:
     - Evgeny Yurchenko (@BGmot)
--- a/plugins/modules/zabbix_script.py
+++ b/plugins/modules/zabbix_script.py
@@ -11,7 +11,7 @@ __metaclass__ = type
 DOCUMENTATION = '''
 module: zabbix_script
 short_description: Gather information about Zabbix proxy
-version_added: 1.6.0
+version_added: 1.7.0
 author:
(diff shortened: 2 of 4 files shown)


changelogs/fragments/685-zabbix-script.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
+  - zabbix_script module added
+  - (https://github.com/ansible-collections/community.zabbix/issues/634)
```
D3DeFi, line 4:
```suggestion
  - zabbix_script module added (https://github.com/ansible-collections/community.zabbix/issues/634)
```

has to be single line, otherwise changelog format will get wonky and require manual fixes

plugins/modules/zabbix_script.py
```diff
@@ -0,0 +1,416 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+module: zabbix_script
+short_description: Gather information about Zabbix proxy
+version_added: 1.6.0
```
D3DeFi, line 14: This has to be next feature release
```suggestion
version_added: 1.7.0
```

plugins/modules/zabbix_script.py
```diff
@@ -0,0 +1,416 @@
...
+# Copyright: (c) 2022, BGmot
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+module: zabbix_script
+short_description: Gather information about Zabbix proxy
```
D3DeFi, line 13: typo?

D3DeFi (changes requested):
Few minor tweaks, otherwise very quality PR! :) Thanks a lot!

changelogs/fragments/685-zabbix-script.yml
BGmot: does file name matter ? 685-zabbix-script.yml should be 689-zabbix-script.yml I think.

D3DeFi: it doesnt

plugins/modules/zabbix_script.py
BGmot: yes, sorry, fixing.

D3DeFi (approved):
will merge once CI finishes. Thank you very much for this addition! :)

Issue #705, opened 2022-05-31
Zabbix Agent installs with --check flag


piotrkochan:
##### SUMMARY

Ansible role zabbix_agent installs Zabbix when --check flag is provided. This is very dangerous as --check is used usually to determine what the playbook will do and it definitely should not make changes to the system.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
https://github.com/ansible-collections/community.zabbix/tree/2287bc4165a03e2ae8a2ed94ec1088428b29dad9/roles/zabbix_agent

##### ANSIBLE VERSION
not revelant

##### CONFIGURATION
not revelant

##### OS / ENVIRONMENT / Zabbix Version

Ubuntu/Debian

##### STEPS TO REPRODUCE

```yaml
- name: Zabbix Agent Playbook
  hosts: zabbix_agent
  become: true
  tasks:
    - name: Ensure host is monitored by Zabbix
      community.zabbix.zabbix_host:
```
ansible-playbook -i ... playbook.yml --check

##### EXPECTED RESULTS

Playbook result without actual changes to the system

##### ACTUAL RESULTS

My friends panicked and cried

##### POSSIBLE FIX

Remove `check_mode: false` from line 200 and 224 

https://github.com/ansible-collections/community.zabbix/blob/2287bc4165/roles/zabbix_agent/tasks/Debian.yml#L200

pyrodie18:
Resolved with https://github.com/ansible-collections/community.zabbix/pull/706....closing

PR #706, opened 2022-05-31
Bugfix 705/check mode debian

piotrkochan:
#705


roles/zabbix_agent/tasks/Debian.yml
```diff
@@ -199,8 +199,7 @@
   when: ansible_distribution in ['Ubuntu', 'Debian']
   register: zabbix_agent_package_installed
   until: zabbix_agent_package_installed is succeeded
-  become: true
-  check_mode: false
+  become: yes
```
brnl, line 202: Why did you change this to `yes` while `true`|`false` is used everywhere in this project? 🤔

brnl:
Note: The issue also exists on line 227: [`roles/zabbix_agent/tasks/Debian.yml`: 227 "Debian | Installing zabbix-{sender,get}"](https://github.com/ansible-collections/community.zabbix/blob/2287bc4165/roles/zabbix_agent/tasks/Debian.yml#L227)

D3DeFi (approved):
@piotrkochan thanks for the patch!

@brnl thank you for helping with review :)

PR #709, opened 2022-06-10
When configuring firewalld, apply new configuration immediately

nathou:
##### SUMMARY

When configuring firewalld, apply new configuration immediately.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zabbix_agent

##### ADDITIONAL INFORMATION

Without this, you need to manually reload firewalld or add a handler elsewhere.

### DIFF
--- a/changelogs/fragments/firewalld_immediate.yaml
+++ b/changelogs/fragments/firewalld_immediate.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - zabbix_agent - when configuring firewalld, make sure the new rule is applied immediately
--- a/roles/zabbix_agent/tasks/firewall.yml
+++ b/roles/zabbix_agent/tasks/firewall.yml
@@ -30,6 +30,7 @@
     rich_rule: 'rule family="ipv4" source address="{{ zabbix_agent_firewalld_source }}" port protocol="tcp" port="{{ zabbix_agent_listenport }}" accept'
     zone: "{{ zabbix_agent_firewalld_zone }}"
     permanent: true
+    immediate: true
     state: enabled
   become: true
   when:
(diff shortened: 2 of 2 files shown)


pyrodie18:
Thanks for the contribution @nathou .

@D3DeFi  would we consider this a major or minor change?  I personally would say minor but am not set in stone either way.

D3DeFi:
minor, anyone that is using `zabbix_agent_firewall*` should expect that configured changes also get applied. I would even say this is bugfix as it should worked like this from start. But change it in the changelog as minor_changes please.

Otherwise legitimate patch, thank you!

pyrodie18 (changes requested):
Please update the changelog fragment to a minor_change.

nathou:
I just did set the change to "minor".

I fact I realized after creating this pull request (did not have much time to share that, sorry) that there already was a `firewalld-reload` handler, but it only executes when ansible_facts.services exists (has been changed in c011538c92aa31fd425d6c9ca1909c4fe7c14c38)
```
- name: firewalld-reload
  command: "firewall-cmd --reload"
  become: true
  when:
    - ansible_facts.services["firewalld"] is defined
    - ansible_facts.services["firewalld"].state == "running"
```
Those facts are only gathered when [removing zabbix_agent](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/tasks/remove.yml), hence the handler is never executed when **installing** zabbix_agent.

Anyway, it seems to me that relying on the firewalld module features is safer.

D3DeFi (approved):
looks good

@pyrodie18 feel free to merge once CI passes if you are satisfied with the introduced changes since last review

pyrodie18 (approved):
Great work and thank you....merging

PR #710, opened 2022-06-10
Make sure PostgreSQL password variable is used consistently across all plays.

CompPhy:
##### SUMMARY
This PR fixes the usage zabbix_server_dbpassword variable for consistency.  Currently, the playbook fails on a fresh database init because the variable isn't used consistently across all plays.

Instead of fixing all the usages of zabbix_server_dbpassword, I instead created a new fact at the top of the playbook, which is based on the zabbix_server_dbpassword variable.  This fact is then used consistently across all the remaining plays.  This should allow for modifying usage of zabbix_server_dbpassword in the future, while maintaining consistent usage.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_server - PostgreSQL

### DIFF
--- a/changelogs/fragments/postgresql_fixpasswordvariable.yaml
+++ b/changelogs/fragments/postgresql_fixpasswordvariable.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - zabbix_server - PostgreSQL - Make sure PostgreSQL password variable is used consistently across all plays.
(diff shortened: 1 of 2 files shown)


CompPhy:
This is actually more complex them originally thought.   I'm still investigating if this PR is correct, and will close for now.

D3DeFi:
Thank you for looking into the matter though! :) Very much appreciated

PR #711, opened 2022-06-11
Minimize agent configs, pointing to online documentation instead

brnl:
##### SUMMARY

I found the used configuration templates are incomplete. To mitigate the troubles we have supporting so many Zabbix versions I thought it would be good to remove the comments and just point to the online documentation for all the configured parameters.

The URL to the correct documentation is dynamicly populated by the used OS family and Zabbix Agent version:
```jinja2
# https://www.zabbix.com/documentation/{{ zabbix_version }}/en/manual/appendix/config/zabbix_agent2{{ "_win" if zabbix_agent_os_family != "Windows" else "" }}
```

**Possible enhancements**
If this idea catches on, I think it would also be great to add comments on what role variable is responsible for the configuration item in the config file. This could help users debug their configurations.

**I'm still a noob at contributing**
Feel free to ditch this PR if it's not helping. But please guide me a bit to how it could have been better. I'm always trying to learn!

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`roles/zabbix_agent`

##### ADDITIONAL INFORMATION

Example generated `/etc/zabbix/zabbix_agent2.conf`
```paste below
#
# Ansible managed
#
# This is a configuration file for Zabbix agent 2
# To get more information about Zabbix, visit http://www.zabbix.com

# This configuration file is "minimalized", which means all the original comments 
# are removed. The full documentation for your zabbix agent2 can be found here:
# https://www.zabbix.com/documentation/5.0/en/manual/appendix/config/zabbix_agent2

PidFile=/var/run/zabbix/zabbix_agent2.pid
LogType=file
LogFile=/var/log/zabbix/zabbix_agent2.log
LogFileSize=100
DebugLevel=3
Server=192.168.0.2
ListenPort=10050
StatusPort=9999
ServerActive=192.168.0.2
Hostname=zabbix-agent1
RefreshActiveChecks=120
BufferSend=5
BufferSize=100
PersistentBufferPeriod=1h
Timeout=3
Include=/etc/zabbix/zabbix_agent2.d
UnsafeUserParameters=0
ControlSocket=/tmp/agent.sock
TLSConnect=psk
TLSAccept=psk
TLSPSKIdentity=zabbix-agent1
TLSPSKFile=/etc/zabbix/zabbix_agent_pskfile.psk
```


mu1f407:
Hi @brnl, interesting suggestion. But I think for some people it is opinionated as some people could prefer native way, i.e. have configuration files same as when Zabbix is installed from official packages.

IMHO this feature would be nice addition with these changes:
- Have it consistent across all config files from all roles. Not only for agent.
- Have it configurable (opt-in).
- The implementation could be probably simpler - e.g. to have templates with original comments and when user wants it minimized, then just generate the template using template lookup and apply some clever regexp which would leave only real configuration and would strip comments and empty lines.

But this is mostly on decision of the maintainers.

D3DeFi:
> Hi @brnl, interesting suggestion. But I think for some people it could be much opinionated as they could prefer native way, i.e. have configuration files same as when Zabbix is installed from official packages.
> 
> IMHO this feature would be nice addition with these changes:
> 
>     * Have it consistent across all config files from all roles. Not only for agent.
> 
>     * Have it configurable (opt-in).
> 
>     * The implementation could be probably simpler - e.g. to have templates with original comments and when user wants it minimized, then just generate the template using template lookup and apply some clever regexp which would leave only real configuration and would strip comments and empty lines.
> 
> 
> But this is mostly on decision of the maintainers.

I am also on the 'minimal config side' when it comes to configuration management tools, as it doesn't make sense to distribute tons of comments to the managed hosts that shouldn't most of the time be accesses manually anyways.

On the other hand I completely understand argument advocating user's choice and agree with this being opt-in. For the implementation specifics, not sure if having each comment block embedded in `{% if comments -%}{%- endif %}` would be better or if we want to risk it with the regular expressions (although something like `^\s*#+\s+` may be enough just for this).

brnl:
The issue I see with the commented configs is that they are version-specific. Unless we commit to add comments only for the role-managed variables, of course. But that's kinda why I've added the version specific documentation link on top of the configs. The user can then read the actual accurate documentation for their specific Zabbix version.

```
# This configuration file is "minimalized", which means all the original comments 
# are removed. The full documentation for your zabbix agent2 can be found here:
# https://www.zabbix.com/documentation/{{ zabbix_version }}/en/manual/appendix/config/zabbix_agent2{{ "_win" if zabbix_agent_os_family == "Windows" else "" }}
```

> Have it consistent across all config files from all roles. Not only for agent.

Sure thin! I'll add the other two as well, if we come to a conclusion in this debate. :+1:

D3DeFi:
sorry to leave this hanging for so long. Since we have a couple of new maintainers we can have a vote! :)

Unfortunately, as dj-wasabi had to step down as a maintainer of roles, we have to decide how to stir this proposition by @brnl.

@BGmot @lzadjsf @pyrodie18 any ideas/preferences/observations?

I vote for removing comments, solely because I am minimalist and always do this in my own roles/etc...

pyrodie18:
I think I am on the minimalist train here personally.  I think the comment at the top with a link to the documentation is sufficient and maintainable in the long term vs trying to keep the comments current and allowing an option for including them or not.

BGmot:
I agree that it is better to go with no comments (would merge this PR) but it would be really nice to have it consistent across all config files from all roles.

brnl:
Thanks for the feedback, all. I'll add the other configurations this evening! I have them ready so I'll just have to commit.

BGmot:
No. Thanks to you @brnl !

D3DeFi:
Ok, I think that there is a consensus towards minimalism. Once this PR includes removal of comments for other roles, we can merge. Thank you all for participating.

Maybe an interesting feature request would be to allow users to provide their own agent conf (whole file) as an alternative to providing all of the variables for the role.

brnl:
I think I got them all now. The zabbix_web role config templates are fine for me.

pyrodie18 (changes requested):
Looks good.  Please include a change fragment and I'm happy.

D3DeFi:
https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment

roles/zabbix_agent/templates/zabbix_agentd.conf.j2
```diff
@@ -1,221 +1,63 @@
...
 
-############ general parameters #################
+# This configuration file is "minimalized", which means all the original comments 
+# are removed. The full documentation for your Zabbix Agent can be found here:
+# https://www.zabbix.com/documentation/{{ zabbix_version }}/en/manual/appendix/config/zabbix_agentd{{ "_win" if zabbix_agent_os_family == "Windows" else "" }}
 
-### option: pidfile
-#       name of pid file.
-#
+{{ ansible_managed | comment }}
```
D3DeFi, line 9: Duplicate ansible_managed block? One present on line 1

D3DeFi (commented):
Also good from my side. Once changleog is present, we can merge

roles/zabbix_agent/templates/zabbix_agentd.conf.j2
brnl: Yup that was a mistake. Fixed, thanks!

brnl:
I'm struggling with the changelog fragment. Feel free to change/critique/reject if it's not good. :-)

pyrodie18 (approved):
Looks good to me.  Great work and thanks for the contribution.

PR #730, opened 2022-07-07
FIX: Changed Invalid parameter "/output" value from "groupid" to "ext…

lzadjsf:
…end".

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/zabbix_host.py
+++ b/plugins/modules/zabbix_host.py
@@ -615,11 +615,11 @@ class Host(ZabbixBase):
     # get group ids by group names
     def get_group_ids_by_group_names(self, group_names):
         if self.check_host_group_exist(group_names):
-            return self._zapi.hostgroup.get({'output': 'groupid', 'filter': {'name': group_names}})
+            return self._zapi.hostgroup.get({'output': 'extend', 'filter': {'name': group_names}})
 
     # get host groups ids by host id
     def get_group_ids_by_host_id(self, host_id):
-        return self._zapi.hostgroup.get({'output': 'groupid', 'hostids': host_id})
+        return self._zapi.hostgroup.get({'output': 'extend', 'hostids': host_id})
 
     # get host templates by host id
     def get_host_templates_by_host_id(self, host_id):


D3DeFi:
Thank you! :)

BGmot:
I would create a ZBX ticket about this failure, doesn't make any sense to me why `groupid` is suddenly not allowed.

mu1f407:
@BGmot
I think the issue is that `groupid` need to be in a list. Quoting from [documentation](https://www.zabbix.com/documentation/current/en/manual/api/reference_commentary):

> Can be defined as an **array of property names** to return only specific properties, or as one of the predefined values:
> extend - returns all object properties;

So I believe this should work: `'output': ['groupid']`

BGmot:
Or don't define `output` at all, it's by default `extend` -)

D3DeFi:
Yeah ok, maybe I should not have merged so hastily, but then another question pops up to my mind. Why was this working on <6.2 version and was it even working correctly?

BGmot:
I think in < 6.2 parameter `output` was ignored and all the fields were always returned and now it is validated and actually in use to produce only requested output. This is the diff between 6.2.0 and 6.0.6
```
--- a/ui/include/classes/api/services/CHostGroup.php
+++ b/ui/include/classes/api/services/CHostGroup.php
<snip>
+       public function get(array $options) {
<snip>
+               $output_fields = ['groupid', 'name', 'flags', 'uuid'];
<snip>
+               $api_input_rules = ['type' => API_OBJECT, 'fields' => [
<snip>
+                       'output' =>                                                             ['type' => API_OUTPUT, 'in' => implode(',', $output_fields), 'default' => API_OUTPUT_EXTEND],

                if ($result) {
                        $result = $this->addRelatedObjects($options, $result);
-               }
+                       $result = $this->unsetExtraFields($result, ['groupid'], $options['output']);
 <snip>
                }
```

No big deal for this merge but I think I'll run all the modules tests against 6.2 and will create a massive issue to fix all failures against 6.2 (I suspect there is many more).

PR #731, opened 2022-07-09
zabbix_template_info: Add yaml and none formats

mu1f407:
##### SUMMARY
Adds 3 features into zabbix_template_info module:
1. `format: yaml` - export of template in YAML format (supported in Zabbix >= 5.2)
2. `format: none` - do not retrieve template export at all
3. `template_id` - return value of ID of the template (which we retrieve anyway in the code but was not returned from module)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_template_info

##### ADDITIONAL INFORMATION FOR `format: yaml`:
I chose to implement `omit_date` without parsing and dumping YAML as it changed the resulting format a lot. I think it is better to keep it in line with native Zabbix output.

##### ADDITIONAL INFORMATION FOR `format: none`:
It is useful when you need only to test template presence/absence. In that case we can skip one possibly huge API request/response. We can then force task to not fail and check result to determine if template exists:
```
- name: retrieve information about template
  community.zabbix.zabbix_template_info:
    ...login parameters...
    template_name: Some template
    format: none
  failed_when: false
  register: template_result

- name: do something when template exists
    ...some module data...
  when: template_result.template_id is defined
```

##### ADDITIONAL INFORMATION FOR `template_id`:
Needed for template existence check as seen in previous example when template export is not returned and task is set to not fail. For consistency I added it to all results as we have the value anyway and it is not returned in the template export itself.

### DIFF
--- a/changelogs/fragments/731-template-info-formats.yml
+++ b/changelogs/fragments/731-template-info-formats.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - zabbix_template_info - add yaml and none formats
+  - zabbix_template_info - add template_id return value
--- a/plugins/modules/zabbix_template_info.py
+++ b/plugins/modules/zabbix_template_info.py
@@ -27,7 +27,8 @@ options:
     format:
         description:
             - Format to use when dumping template.
-        choices: ['json', 'xml']
+            - C(yaml) works only with Zabbix >= 5.2.
+        choices: ['json', 'xml', 'yaml', 'none']
         default: json
         type: str
     omit_date:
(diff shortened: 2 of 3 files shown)


D3DeFi:
This is awesome addition! Looking forward to merging this. Seems 6.0 pipelines are failing around this step:
https://github.com/ansible-collections/community.zabbix/runs/7265163510?check_suite_focus=true#step:7:5313

mu1f407:
@D3DeFi Everything is fixed now. Thank you.

BGmot:
@D3DeFi why codecov tests are suddenly green?

D3DeFi:
@BGmot since failed tests don't progress further after failure, but rather try to repeat same steps with increased verbosity and mostly fail on the same tasks, codecov interprets this as worse coverage from tests (same as if this PR would remove all of those tests after failed task). 

Once the failure is fixed and tests are run again, codecov corrects its report.

@mu1f407 thank you for this addition!

PR #733, opened 2022-07-10
Add zabbix_api_info module and tests

masa-orca:
##### SUMMARY
Add apiinfo_facts module

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

community.zabbix.zabbix_apiinfo_facts

##### ADDITIONAL INFORMATION
- Add apiinfo_facts module and tests.
- This returns only version info now


plugins/modules/zabbix_apiinfo_facts.py
```diff
@@ -0,0 +1,83 @@
...
+            self._module.exit_json(apiinfo=apiinfo)
+        except Exception as e:
+            self._module.fail_json(msg="Failed to get apiinfo version: %s" % e)
+
+
+def main():
+    argument_spec = zabbix_utils.zabbix_common_argument_spec()
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True
```
D3DeFi, line 75: I would go for `False` here. You after all, do not use if `module.check_mode` anywhere and for this kind of module it is not even needed. 

```suggestion
        supports_check_mode=False
```

D3DeFi (changes requested):
Nice module addition. Thank you! :)

D3DeFi (approved):
Looks good to me :)

D3DeFi:
I may have miss-lead you with a check mode :disappointed: 

https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general.html#creating-an-info-or-a-facts-module

it seems they must to support check mode? Probably a reason why this fails.

I am so sorry

mu1f407:
I was just reading through documentation yesterday and noticed the requirements for info/facts modules. Beside check mode, fact modules must return `ansible_facts`. Module documentation (`RETURN`) should also reflect that. Please see [ipinfoio_facts](/ansible-collections/community.general/blob/main/plugins/modules/net_tools/ipinfoio_facts.py) as an example.

BGmot:
To be honest I'd rename this module to either zabbix_api_info or zabbix_api_facts.

masa-orca:
Hi, Reviewers.
Thank you for your reviewing. I appologize for taking much times to modify.
I read [creating-an-info-or-a-facts-module](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general.html#creating-an-info-or-a-facts-module) (@D3DeFi mentioned). 
The document says 
> Modules that query/return general information (and not ansible_facts) should be named _info. General information is non-host specific information, for example information on online/cloud services (you can access different accounts for the same online service from the same host), or information on VMs and containers accessible from the machine, or information on individual files or programs.

Therefor, I think it is `_info` module.

> To be honest I'd rename this module to either zabbix_api_info or zabbix_api_facts.

I agree it. I renamed module to `zabbix_api_info `.

BGmot:
Thanks a lot @masa-orca !

plugins/modules/zabbix_api_info.py
```diff
@@ -0,0 +1,108 @@
...
+short_description: Retrieve Zabbix API info
+
+description:
+   - This module allows you to retrieve Zabbix api info.
+
+author:
+    - ONODERA Masaru(@masa-orca)
+
+requirements:
+    - "python >= 2.6"
```
BGmot, line 24: Since we are adding this module just now would you please update?
```suggestion
    - "python >= 3.9"
```

plugins/modules/zabbix_api_info.py
```diff
@@ -0,0 +1,108 @@
...
+
+version_added: 1.8.0
```
BGmot, line 26: Since we are adding this module just now would you please update?
```suggestion
version_added: 2.1.0
```

plugins/modules/zabbix_api_info.py
```diff
@@ -0,0 +1,108 @@
...
+RETURN = '''
+api:
+    description: Summaries of Zabbix API info
+    returned: success
+    type: dict
+    contains:
+        version:
+            description: API version
+            type: str
+            sample: 4.0.42
```
BGmot, line 68: Maybe a bit fresher example?
```suggestion
            sample: 6.0.18
```

BGmot (changes requested):

masa-orca:
@BGmot Thank you for your reviwing! I updated documents.

BGmot (approved):

masa-orca:
I rebased main branch and added this module to action_groups.

PR #744, opened 2022-07-17
zabbix_mediatype: Add support for time units in attempt_interval

### LINKED ISSUE #721: zabbix_mediatype: attempt_interval not allowing time suffixes
##### SUMMARY
The [documentation of Zabbix 6](https://www.zabbix.com/documentation/6.0/en/manual/config/notifications/media) states for the parameter `attempt_interval` that input with time suffixes (e.g. 10s, 5m, 2h) is supported. However, when using the parameter `attempt_interval` in module `zabbix_mediatype` an error appears, stating that only integer is allowed. The feature request is to support time suffixes in parameter `attempt_interval` as well.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
zabbix_mediatype

##### ADDITIONAL INFORMATION
Current error message, when using time suffix, e.g. 10s,  in module `zabbix_mediatye`.

````
argument attempt_interval is of type <class 'str'> and we were unable to convert to int: <class 'str'> cannot be converted to an int
````

mu1f407:
##### SUMMARY
Adds support for time unit suffixes for `attempt_interval` parameter in zabbix_mediatype module. Now it accepts both, only number for seconds or unit suffixed number. Time suffixes in this parameter are supported since the parameter appeared in API, so there is no need to make if/else based on version.

Resolves #721

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_mediatype

### DIFF
--- a/changelogs/fragments/744-mediatype-attemptinterval-timeunits.yml
+++ b/changelogs/fragments/744-mediatype-attemptinterval-timeunits.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - zabbix_mediatype - added support for time units in ``attempt_interval`` parameter
--- a/plugins/modules/zabbix_mediatype.py
+++ b/plugins/modules/zabbix_mediatype.py
@@ -77,12 +77,12 @@ options:
             - Works only with Zabbix versions 3.4 or newer.
         default: 3
     attempt_interval:
-        type: 'int'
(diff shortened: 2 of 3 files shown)


D3DeFi (approved):
looks good to me! :)

PR #770, opened 2022-08-01
zabbix_agent: Add zabbix_agent(2)_include_pattern variable

### LINKED ISSUE #748: zabbix_agent_include var should be able to contain a glob pattern
##### SUMMARY

Zabbix-agent can use a glob pattern in the `Include` key. By default this is something like `/etc/zabbix/zabbix.agent.d/*.conf`. When you change this, it will try to create a directory called `*.conf`. 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ansible_agent

##### ANSIBLE VERSION

```paste below
ansible [core 2.13.1]
  ansible python module location = /usr/local/Cellar/ansible/6.1.0/libexec/lib/python3.10/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.10.5 (main, Jun 23 2022, 17:15:25) [Clang 13.1.6 (clang-1316.0.21.2.5)]
  jinja version = 3.1.2
  libyaml = True
```

##### CONFIGURATION

```paste below
DEFAULT_HOST_LIST([path]/ansible.cfg) = ['[path]/inventory']
DEFAULT_ROLES_PATH([path]/ansible.cfg) = ['[path]/roles']
DEPRECATION_WARNINGS([path]/ansible.cfg) = False
```

##### STEPS TO REPRODUCE

Change the `zabbix_agent_include` variable of the role to `/etc/zabbix/zabbix.agent.d/*.conf` and run the ansible_agent playbook.

##### EXPECTED RESULTS

The directory `/etc/zabbix/zabbix.agent.d/*.conf/` should not be created.

##### ACTUAL RESULTS

The directory `/etc/zabbix/zabbix.agent.d/*.conf/` is created and looks stupid. The template however uses the correct glob pattern.

mu1f407:
##### SUMMARY
Adds possibility to set `zabbix_agent(2)_include_pattern` for zabbix_agent role. Allows to use custom file pattern in Include configuration option.

~~Adds possibility to set `zabbix_agent(2)_include_dir` for zabbix_agent role. Allows to use file pattern in the `zabbix_agent(2)_include`. It defaults to the latter to not break existing role's configs.~~

Resolves #748

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_agent

##### EXAMPLE
```yaml
zabbix_agent_include_pattern: "*.conf"
```

### DIFF
--- a/changelogs/fragments/770-agent-include-pattern.yml
+++ b/changelogs/fragments/770-agent-include-pattern.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - zabbix_agent - add possiblity to set include file pattern using ``zabbix_agent(2)_include_pattern`` variable.
--- a/docs/ZABBIX_AGENT_ROLE.md
+++ b/docs/ZABBIX_AGENT_ROLE.md
@@ -211,6 +211,7 @@ Otherwise it just for the Zabbix Agent or for the Zabbix Agent 2.
 * `zabbix_agent(2)_zabbix_alias`: sets an alias for parameter. it can be useful to substitute long and complex parameter name with a smaller and simpler one. Can be both a string as an list.
 * `zabbix_agent(2)_timeout`: spend no more than timeout seconds on processing
 * `zabbix_agent(2)_include`: you may include individual files or all files in a directory in the configuration file.
+* `zabbix_agent(2)_include_pattern`: Optional file pattern used for included files.
 * `zabbix_agent(2)_include_mode`: The mode for the directory mentioned above.
 * `zabbix_agent(2)_unsafeuserparameters`: allow all characters to be passed in arguments to user-defined parameters.
 * `zabbix_agent_loadmodulepath`: Full path to location of agent modules.
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -141,6 +141,7 @@ zabbix_agent_allowroot: 0
 zabbix_agent_zabbix_alias:
 zabbix_agent_timeout: 3
 zabbix_agent_include: /etc/zabbix/zabbix_agentd.d
+zabbix_agent_include_pattern:
 zabbix_agent_include_mode: "0750"
 zabbix_agent_unsafeuserparameters: 0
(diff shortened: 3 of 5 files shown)


D3DeFi (approved):
So If I am understanding this correctly. if user wants to use `_include: x/y/z/*` then they need to also provide `_include_dir: x/y/z` that will not have a glob, right?

mu1f407:
Yes, exactly. It keeps backward compatibility when users use only `zabbix_agent_include`. If they want to use file pattern in `zabbix_agent_include`, then they need to set `zabbix_agent_include_dir` to valid directory path accordingly.

Another possible option was to use some pattern detection and apply `basename` filter and use only `zabbix_agent_include` variable as before, but it would be too much hidden logic in implementation.

leroy0211:
Just from the top of my head. 

Wouldn't it been better to have a `zabbix_agent_include` and `zabbix_agent_include_pattern` variable instead? 

This way you won't have to configure both variables, which are nearly the same, if you want to have a file pattern (which is now the default when you install the agent on debian using apt).

You can create the directory from `zabbix_agent_include`, and use the combination `{{ zabbix_agent_include }}/{{ zabbix_agent_include_pattern }}` in the template configuration. 

When kept empty it results in `/etc/zabbix/zabbix_agentd.d/`
When you use a file pattern like `*.conf` it results in `/etc/zabbix/zabbix_agentd.d/*.conf` and does not create the `*.conf` directory. 

Correct me if I'm missing something here. I just thought that would be a bit simpler.

leroy0211:
> Another possible option was to use some pattern detection and apply `dirname` filter and use only `zabbix_agent_include` variable as before, but it would be too much hidden logic in implementation.

Or throw an error when a glob character is used in the `zabbix_agent_include` variable. Because the glob character can in theory also be `*/*.conf`. So dirname filter would not work.

mu1f407:
Very nice idea @leroy0211! I like it. Much simpler than my original approach. I've reworked it as you proposed. Please take a look and let me know. Thank you. :slightly_smiling_face:

D3DeFi (approved):
Even better!

PR #773, opened 2022-08-12
6.2 update

pyrodie18:
##### SUMMARY
Updated roles to support Zabbix 6.2 referenece #761 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Various Roles

##### ADDITIONAL INFORMATION
N/A

PR #775, opened 2022-08-15
fix sanity tests

D3DeFi:
##### SUMMARY
Fixes problems encountered in #773 

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/plugins/httpapi/jsonrpc.py
+++ b/plugins/httpapi/jsonrpc.py
@@ -46,14 +46,14 @@ class HttpApi(HttpApiBase):
 
     def login(self, username, password):
         payload = self.payload_builder("user.login", username=username, password=password)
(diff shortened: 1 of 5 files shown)


BGmot (approved):
Looks ok, thanks. I am surprised though we suddenly need to change code that was not touched and previously sanity checks passed.

D3DeFi:
> Looks ok, thanks. I am surprised though we suddenly need to change code that was not touched and previously sanity checks passed.

Basically this: https://github.com/ansible-collections/community.zabbix/pull/773#discussion_r945660945

tl;dr, we test against ansible develop so things like this are expected sometimes

mu1f407 (approved):
Nice, thank you! I was confused that the sanity tests are suddenly failing too.

Issue #777, opened 2022-08-16
Update Testing Matrix


pyrodie18:
##### SUMMARY
Currently, we only test against the latest version of Zabbix as well as the latest operating systems but report that we support 4 versions of Zabbix and a few older operating systems.  Should update the testing matrix so that it tests all supported versions and operating systems.

##### ISSUE TYPE
- Feature Idea

pyrodie18:
My suspicion is that either Github gives free minutes to large open-source projects, or Ansible (RedHat) is paying for the runners.  Anyone that we could ask?

The other option is that we build a separate trigger event that would do a complete regression test of everything that would be triggered as part of a release event (not sure what the trigger would be exactly but I'm sure we could figure something out).

D3DeFi:
@Andersson007 would you be able to shine some light into this matter? :)

pyrodie18:
> @Andersson007 would you be able to shine some light into this matter? :)
> 
> Mostly talking about roles here. We do test plugins against zabbix 3.0, 4.0, 5.0, 6.0, 6.2 with stable-2.10-13 and develop
> 
> side note: should add stable-2.14 probably, right?

ya.  And remove 3.0.  Did we decide if we're going to keep or ditch 4.0 when we role versions?

nitzmahone:
Yes, Ansible's paying for the VMs in a gigantic Azure VMSS that serves both ansible-core and the community collections projects, among other things.

webknjaz:
> Yes, Ansible (well, Red Hat) is paying for the VMs in a gigantic Azure VMSS that serves both ansible-core and the community collections projects, among other things.

This project doesn't seem to make use of the Ansible Core CI, though. Evidently, it uses the free minutes that GH has for public projects: https://github.com/ansible-collections/community.zabbix/actions/runs/2867704689/usage.

P.S. By the way, @pyrodie18 if you're looking for areas of improvements, try integrating https://github.com/marketplace/actions/ansible-test — it helps reduce some boilerplate the collection maintainers end up supporting and is already being offered by the collection template project by default: https://github.com/ansible-collections/collection_template/blob/main/.github/workflows/ansible-test.yml.

D3DeFi:
Thanks for clarifying @nitzmahone @webknjaz ! Yes, I don't think that we use Azure in any way, we also cannot utilize ansible-test container as we do need to test same codebase against multiple versions of zabbix-server.

[...]

Issue #768, opened 2022-07-31
Simply Variable Definition


pyrodie18:
##### SUMMARY
The server and web roles make use of both `zabbix_server_database` and `zabbix_server_database_long`.  While the two variables may hold different values (pgsql vs postgresql), the two values will always be tightly related.  Suggest that we simply the configuration process and add logic at the top of the playbook that will automatically set `zabbix_server_database_long` based on the value of `zabbix_server_database`

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
- Server Role
- Web Role

##### ADDITIONAL INFORMATION
N/A

Issue #766, opened 2022-07-30
Remove Support for Unsupported Versions


pyrodie18:
##### SUMMARY
Remove support for Zabbix 4.X

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
Various roles and modules

##### ADDITIONAL INFORMATION
In our README we outline that we support the two most current LTS releases which are currently 6.0 and 5.0.  We still include support for 4.X installs.  I suggest that we go ahead and deprecate that functionality from the collection as part of our next major release.

BGmot:
I am not sure. Zabbix 4.0 still has limited support until October 31, 2023 https://www.zabbix.com/life_cycle_and_release_policy
Why remove what's working? Or you mean to put some 'deprecation notice'?

pyrodie18:
> I am not sure. Zabbix 4.0 still has limited support until October 31, 2023 https://www.zabbix.com/life_cycle_and_release_policy Why remove what's working? Or you mean to put some 'deprecation notice'?

Because that's what we said we would do in https://github.com/ansible-collections/community.zabbix#supported-zabbix-versions.  It would allow us to clean up the code base a lot and simply the roles.  As is, we still have support in there for <= 3.2, and we're not currently testing the installs for any of them.

D3DeFi:
I think when we prepared our release document we stated that major releases would warrant for forking _main_ branch into _stable-1.x_ branch. 

I am all up for removing 4.x functionality at this point. Users can lock community.zabbix at 1.y.z version and we release 2.0.0 without zabbix 3.x and 4.x.

If anything critical is found, we can always backport the fix and git tag 1.y.z + 1 version.

Issue #764, opened 2022-07-29
FR: Add install Support for Debian 11


pyrodie18:
##### SUMMARY
Zabbix 5.0, 6.0 and 6.2 has support for Debian 11. Need to update roles to allow for installation

##### ISSUE TYPE
- Feature Idea

PR #795, opened 2022-09-03
Proxy update

pyrodie18:
##### SUMMARY
Initial changes to proxy role for 2.0.0

##### COMPONENT NAME
Proxy Role

##### ADDITIONAL INFORMATION
Major Changes include:

- Updating Molecule matrix to test all OS, with all versions, and all databases (#777 )
- Updated documentation
- Removed support for anything < 5.0 (#766)
- Cleaned up handling of variables between OSs
- Added Ubuntu 22.04 (#763)
- Added Support for Debian 11 (#764)
- Removed support for EOL operating systems
- Cleaned up some variables (#768 )


D3DeFi (approved):
Looks good

changelogs/fragments/proxy.yml
```diff
@@ -0,0 +1,18 @@
+breaking_changes:
```
D3DeFi, line 1: Same as for zabbix_javagateway changelog fragments, information that is not relevant to the end user is not supposed to be here. Everything else needs to be prepended with the name of this role otherwise CHANGELOG.md will hardly make sense

D3DeFi (changes requested):

changelogs/fragments/proxy.yml
pyrodie18: Updates Made

D3DeFi (approved):

Issue #763, opened 2022-07-29
FR: Add install Support for Ubuntu 22.02


pyrodie18:
##### SUMMARY
Zabbix 5.0, 6.0 and 6.2 has support for Ubuntu 22.02. Need to update roles to allow for installation

##### ISSUE TYPE
- Feature Idea

evilhamsterman:
Just a minor quibble, there is no Ubuntu 22.02. The latest release is 22.04 https://wiki.ubuntu.com/Releases

pyrodie18:
Ya that was a typo

PR #796, opened 2022-09-03
Java Gateway Update

pyrodie18:
##### SUMMARY
Initial changes to java gateway role for 2.0.0

##### COMPONENT NAME
Java Gateway Role

##### ADDITIONAL INFORMATION
Major Changes include:

- Updating Molecule matrix to test all OS, with all versions, and all databases (#777 )
- Updated documentation
- Removed support for anything < 5.0 (#766)
- Cleaned up handling of variables between OSs
- Added Ubuntu 22.04 (#763)
- Added Support for Debian 11 (#764)
- Removed support for EOL operating systems
- Cleaned up some variables (#768 )

### DIFF
--- a/changelogs/fragments/javagateway.yml
+++ b/changelogs/fragments/javagateway.yml
@@ -1,7 +1,0 @@
-minor_changes:
-  - github workflow - Updated workflow files to checked against all supported combinations
-  - documentation - Updated role documentation to reflect changes
-breaking_changes:
-  - Zabbix < 5.0 - Removed support for all versions older than 5.0
-  - EOL Operating Systems - Removed support for EOL operating systems
-  - zabbix_server - Replaced zabbix_version variable with zabbix_javagateway_version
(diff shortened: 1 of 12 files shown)


changelogs/fragments/javagateway.yml
```diff
@@ -0,0 +1,7 @@
+minor_changes:
```
D3DeFi, line 1: both of this sentences does not bring any value to users of this collection and I would recommend to remove them entirely

changelogs/fragments/javagateway.yml
```diff
@@ -0,0 +1,7 @@
...
+  - github workflow - Updated workflow files to checked against all supported combinations
+  - documentation - Updated role documentation to reflect changes
+breaking_changes:
```
D3DeFi, line 4: all changes here need to be prepended with role name otherwise resulting changelog will be a mess

D3DeFi (changes requested):
See my comments related to changelogs, otherwise this is a good patch

changelogs/fragments/javagateway.yml
pyrodie18: Updates Made

D3DeFi (approved):

PR #800, opened 2022-09-12
zabbix_template: Add support for template tags

mu1f407:
##### SUMMARY
Adds support for template tags into zabbix_template module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_template

### DIFF
--- a/plugins/modules/zabbix_template.py
+++ b/plugins/modules/zabbix_template.py
@@ -821,6 +821,11 @@ def main():
                 for key in macroitem:
                     macroitem[key] = str(macroitem[key])
 
+        if template_tags is not None:
+            for tagitem in template_tags:
+                for key in tagitem:
+                    tagitem[key] = str(tagitem[key])
+
         if not template_ids:
             # Assume new templates are being added when no ID's were found
             if template_content is not None:
--- a/plugins/modules/zabbix_template.py
+++ b/plugins/modules/zabbix_template.py
@@ -462,13 +462,13 @@ class Template(ZabbixBase):
             if template_macros != existing_macros:
                 changed = True
 
-        if 'tags' not in existing_template['zabbix_export']['templates'][0]:
-            existing_template['zabbix_export']['templates'][0]['tags'] = []
-
-        if template_tags is not None:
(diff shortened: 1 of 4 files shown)


plugins/modules/zabbix_template.py
```diff
@@ -429,9 +462,17 @@ def check_template_changed(self, template_ids, template_groups, link_templates,
             if template_macros != existing_macros:
                 changed = True
 
+        if LooseVersion(self._zbx_api_version) >= LooseVersion('4.2'):
+            if 'tags' not in existing_template['zabbix_export']['templates'][0]:
+                existing_template['zabbix_export']['templates'][0]['tags'] = []
+            if template_tags is not None:
+                existing_tags = existing_template['zabbix_export']['templates'][0]['tags']
+                if template_tags != existing_tags:
```
D3DeFi, line 470: I wonder if we should not rather utilize helper function for comparing list of dictionaries from module_utils/

D3DeFi (approved):
Very nnice addition, thank you

D3DeFi:
Failing checks may be because you edited template1_50_lower.json, which is also used for zabbix 4.0 I think

mu1f407:
> Failing checks may be because you edited template1_50_lower.json, which is also used for zabbix 4.0 I think

Previously it failed with another issue, but later it failed exactly on this. I've reverted this change and use different approach in tests.

plugins/modules/zabbix_template.py
mu1f407: I agree, will try it later. Initially I've used the same handling as for macros.

mu1f407: Well, I tried to use helpers, but the code was less readable IMO and the result was the same. If its working fine using just simple compare, than I would leave it as it. I think helpers can be useful for some more complex cases.

D3DeFi: ok, in that case I have no further objections :)

D3DeFi (approved):
LGTM

PR #820, opened 2022-10-11
First attempt at User Role plugin

mrvanes:
##### SUMMARY
Adds User Role plugin

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
zabbix_user_role

##### ADDITIONAL INFORMATION
First attempt at creating user_role plugin, based on the logic of user_info.
I added my own is_part_of configuration comparison method (and helpers) so that idempotency is true, even if Zabbix returns more rules than what was configured in the role, but what was configured is at least the same as what was returned.

### DIFF
--- a/plugins/modules/zabbix_user_role.py
+++ b/plugins/modules/zabbix_user_role.py
@@ -12,7 +12,7 @@ DOCUMENTATION = r'''
 module: zabbix_user_role
 short_description: Adds or removes zabbix roles
 author:
-    - Martin van Es
+    - Martin van Es (@mrvanes)
 description:
     - This module adds or removes zabbix roles
 requirements:
@@ -145,7 +145,7 @@ def main():
 
     argument_spec = zabbix_utils.zabbix_common_argument_spec()
     argument_spec.update(dict(
-        state=dict(type='str', required=False, default='present'),
+        state=dict(type='str', required=False, default='present', choices=['present', 'absent']),
         name=dict(type='str', required=True),
         type=dict(type='str', required=False, choices=["User", "Admin", "Super Admin"], default='User'),
         rules=dict(type='dict', required=False, default={}),
--- a/plugins/modules/zabbix_user_role.py
+++ b/plugins/modules/zabbix_user_role.py
@@ -153,8 +153,9 @@ def main():
     state = module.params['state']
     name = module.params['name']
     type = zabbix_utils.helper_to_numeric_value(
-        ['user', 'admin', 'auper admin'], module.params['type'].lower()
+        ['', 'user', 'admin', 'auper admin'], module.params['type'].lower()
     )
+    # raise Exception(f"type: {type}")
     rules = module.params['rules']
 
     user_role = UserRole(module)
--- a/plugins/modules/zabbix_user_role.py
+++ b/plugins/modules/zabbix_user_role.py
@@ -155,7 +155,6 @@ def main():
     type = zabbix_utils.helper_to_numeric_value(
         ['', 'user', 'admin', 'auper admin'], module.params['type'].lower()
     )
-    # raise Exception(f"type: {type}")
     rules = module.params['rules']
 
     user_role = UserRole(module)
--- a/plugins/modules/zabbix_user_role.py
+++ b/plugins/modules/zabbix_user_role.py
@@ -4,12 +4,11 @@
 # Copyright: (c) 2022, mrvanes
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
+
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-
 DOCUMENTATION = r'''
----
 module: zabbix_user_role
 short_description: Adds or removes zabbix roles
 author:
@@ -18,18 +17,34 @@ description:
     - This module adds or removes zabbix roles
 requirements:
     - "python >= 2.6"
-    - "zabbix-api >= 0.5.4"
 options:
     state:
+        description:
+            - State of the user_role.
+            - On C(present), it will create if user_role does not exist or update the user_role if the associated data is different.
+            - On C(absent) will remove a user_role if it exists.
+        default: 'present'
+        choices: ['present', 'absent']
+        type: str
+        required: false
     name:
-        description: Name of the role to be processed
+        description:
+            - Name of the role to be processed
+        type: str
+        required: true
     type:
         description:
-            - User (default)
-            - Admin
-            - Super admin
+            - User type.
+        choices: ["User", "Admin", "Super Admin"]
+        default: "User"
+        type: str
+        required: false
     rules:
-        - description: Rules set as defined in https://www.zabbix.com/documentation/current/en/manual/api/reference/role/object#role-rules
+        description:
+            - Rules set as defined in https://www.zabbix.com/documentation/current/en/manual/api/reference/role/object#role-rules
+        default: {}
+        type: dict
+        required: false
 extends_documentation_fragment:
 - community.zabbix.zabbix
 '''
@@ -132,7 +147,7 @@ def main():
     argument_spec.update(dict(
         state=dict(type='str', required=False, default='present'),
         name=dict(type='str', required=True),
-        type=dict(type='str', required=False, default='User'),
+        type=dict(type='str', required=False, choices=["User", "Admin", "Super Admin"], default='User'),
         rules=dict(type='dict', required=False, default={}),
     ))
 
--- a/plugins/modules/zabbix_user_role.py
+++ b/plugins/modules/zabbix_user_role.py
@@ -1,0 +1,186 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2022, mrvanes
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: zabbix_role
+https://www.zabbix.com/documentation/current/en/manual/api/reference/role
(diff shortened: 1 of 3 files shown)


BGmot:
Please add tests for this module to https://github.com/ansible-collections/community.zabbix/tree/main/tests/integration/targets.

plugins/modules/zabbix_user_role.py
```diff
@@ -0,0 +1,183 @@
...
+    server_url: http://localhost/zabbix
+    login_user: username
+    login_password: password
+    state: exact
+    name: Operators
+      The name of the role
+    type: 1
+      https://www.zabbix.com/documentation/current/en/manual/api/reference/role/object#role
+    rules:
+      https://www.zabbix.com/documentation/current/en/manual/api/reference/role/object#role-rules
```
D3DeFi, line 31: This needs to be valid YAML, it will not parse and pipelines will fail. Links are redundant here, you may add them under `description:` key, which can be either list of strings or just a string

D3DeFi: also we do not allow referencing Zabbix API that much and try to keep documentation clear of the links that require users to understand and parse API documentation. So this needs to be rewritten here.

For `type: 1` it is preferred to specify string notation and let module handle the translation. In this case there would be `choices ["Admin", "User", "etc..."]` and the module would translate them like this: https://github.com/ansible-collections/community.zabbix/blob/10a4040a8a88d50068d3ecc8475f877db8603864/plugins/modules/zabbix_autoregister.py#L143 (the helper function returns and integer based on the position in the list)

same applies for whole rules section. Here is more complicated example of the module:
https://github.com/ansible-collections/community.zabbix/blob/main/plugins/modules/zabbix_mediatype.py

see how it handles `suboptions` in the DOCUMENTATION block and in the main() function, when defining `argument_spec`

plugins/modules/zabbix_user_role.py
```diff
@@ -0,0 +1,183 @@
...
+                roleid = role['roleid']
+                if state == 'absent':
+                    result = user_role._zapi.role.delete([f"{roleid}"])
+                    changed = True
+                    msg = "Role deleted"
+                else:
+                    if not user_role.is_part_of(rules, role['rules']):
+                        result = user_role._zapi.role.update({"roleid": roleid, "rules": rules})
+                        changed = True
+                        msg = "Role updated"
```
D3DeFi, line 167: I think this section and whole functions behind that can be replaced with https://github.com/ansible-collections/community.zabbix/blob/10a4040a8a88d50068d3ecc8475f877db8603864/plugins/module_utils/helpers.py#L113 and the function that follows it.

plugins/modules/zabbix_user_role.py
```diff
@@ -0,0 +1,183 @@
...
+        type=dict(type='int', required=False, default=1),
+        rules=dict(type='dict', required=True),
+    ))
+
+    # the AnsibleModule object
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True
+    )
+
```
D3DeFi, line 145: Something like this https://github.com/ansible-collections/community.zabbix/blob/10a4040a8a88d50068d3ecc8475f877db8603864/plugins/modules/zabbix_mediatype.py#L718-L722 is missing here

D3DeFi (commented):
Module looks good. As @BGmot mentioned, this needs tests before it can be accepted. 

You can try those test as described here:
https://github.com/ansible-collections/community.zabbix/blob/10a4040a8a88d50068d3ecc8475f877db8603864/CONTRIBUTING.md#testing-zabbix-modules

tl;dr `ansible-test integration -v test_zabbix_role`

plugins/modules/zabbix_user_role.py
mrvanes: As far as the role is concerned, I'm not interested in the difference between conf and Zabbix, but whether what I configured is exactly present like that in Zabbix, the other keys are don't care.

D3DeFi (approved):
I would go ahead and merge if tests pass, then improve further. (I suspect some documentation tweaking may be necessary before tests stop complaining)

@BGmot any further tips to give here or are we ok to merge?

BGmot:
@mrvanes would you please fix all the errors you see in tests?

BGmot:
Hi @mrvanes,
please fix all the test failures you see here https://github.com/ansible-collections/community.zabbix/actions/runs/3248034324
we can't merge if any test fails. If you need any help please ask.

plugins/modules/zabbix_user_role.py
```diff
@@ -0,0 +1,191 @@
...
+    - "zabbix-api >= 0.5.4"
+options:
+    state:
+    name:
+        description: Name of the role to be processed
+    type:
+        description:
+            - User (default)
+            - Admin
+            - Super admin
```
D3DeFi, line 30: Something like this is required:
```suggestion
    type:
        description:
            - User type.
        choices: ["User", "Admin", "Super Admin"]
        default: "User"
        type: str
        required: true
```

This also needs to be reflected in the `argument_spec.update()` call near the bottom of the file.

plugins/modules/zabbix_user_role.py
```diff
@@ -0,0 +1,191 @@
...
+    rules:
+        - description: Rules set as defined in https://www.zabbix.com/documentation/current/en/manual/api/reference/role/object#role-rules
```
D3DeFi, line 32: I would suggest this to be covered fully in the module as seen in the [zabbix_user](https://github.com/ansible-collections/community.zabbix/blob/eecef88328d7b80eae1973a2dfcf232624506bd9/plugins/modules/zabbix_user.py#L111-L174) module for the example.

At minimum, it needs:
```
type: dict
required: true/false
default: {}
```

I am not sure however if this will satisfy sanity tests (enforced by the upstream - Ansible) and `suboptions` block may be required.

D3DeFi (changes requested):
As mentioned, last failing tests seems to be regarding the documentation. You can proof it by running the following command on your PC:
```
ansible-doc -t module community.zabbix.zabbix_role
```

D3DeFi:
Thank you for your dedication and patience to contribute this module and bearing with us during the review process. As you probably already understand, we need to satisfy couple of standards presented by the Ansible in order to be included in the releases. That is the reason why we are going through the multiple iterations of reviews at this point.

Unfortunately this is further slowed down by the CI waiting for me to click on `Approve and Run`, which is the default for 1st time contributors. I would be much happier if it would be enough to just click it once and not for every commit that contributor makes 😅 

Please do not hesitate to ask any questions. I personally may not be so immediate with my responses, but will try to do my best to help when needed. We also have Gitter, which you've already used or feel free to contact me directly at email I've published under my profile here on GH if anything arises.

D3DeFi (approved):
Nicely done!

plugins/modules/zabbix_user_role.py
```diff
@@ -0,0 +1,206 @@
...
+        return result
+
+
+def main():
+    msg = "No action"
+    changed = False
+
+    argument_spec = zabbix_utils.zabbix_common_argument_spec()
+    argument_spec.update(dict(
+        state=dict(type='str', required=False, default='present'),
```
D3DeFi, line 148:
```suggestion
        state=dict(type='str', required=False, default='present', choices=['present', 'absent']),
```

plugins/modules/zabbix_user_role.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+module: zabbix_user_role
+short_description: Adds or removes zabbix roles
+author:
+    - Martin van Es
```
D3DeFi, line 15:
```suggestion
    - Martin van Es (@mrvanes)
```

D3DeFi (commented):
small details

BGmot:
Please address my remarks about removing `zabbix_url`, `login_user`, `login_password` from EXAMPLE section of documentation and tests.

tests/integration/targets/test_zabbix_user_role/tasks/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    - assert:
+        that:
+          - create_zabbix_user_role_result.changed is sameas true
+
+    # Check user role idempotency
+    - name: test - Update a Zabbix role with same rules
+      zabbix_user_role:
+        server_url: "{{ zabbix_api_server_url }}"
+        login_user: "{{ zabbix_api_login_user }}"
+        login_password: "{{ zabbix_api_login_pass }}"
```
D3DeFi, line 28:
```suggestion
```

tests/integration/targets/test_zabbix_user_role/tasks/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    - assert:
+        that:
+          - create_zabbix_user_role_result.changed is sameas false
+
+    # Check user role change
+    - name: test - Update a Zabbix role with new rules
+      zabbix_user_role:
+        server_url: "{{ zabbix_api_server_url }}"
+        login_user: "{{ zabbix_api_login_user }}"
+        login_password: "{{ zabbix_api_login_pass }}"
```
D3DeFi, line 50:
```suggestion
```

tests/integration/targets/test_zabbix_user_role/tasks/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    - assert:
+        that:
+          - create_zabbix_user_role_result.changed is sameas true
+
+    # Check user role remove
+    - name: test - Remote Zabbix role
+      zabbix_user_role:
+        server_url: "{{ zabbix_api_server_url }}"
+        login_user: "{{ zabbix_api_login_user }}"
+        login_password: "{{ zabbix_api_login_pass }}"
```
D3DeFi, line 72:
```suggestion
```

D3DeFi (commented):
since this was asked, lets be consistent for all test cases

D3DeFi:
:tada: thank you for your work! :rocket:

Issue #814, opened 2022-10-01
Migrate roles to use httpapi instead of zabbix-api


D3DeFi:
##### SUMMARY
As per #806, httpapi can now be utilized to connect to the Zabbix server. This issue requests that all roles will be migrated to it and drop zabbix-api logic

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
roles/*

BGmot:
Resolved by #822

PR #822, opened 2022-10-14
Use httpapi in roles

BGmot:
##### SUMMARY
Addresses #814 
I only changed zabbix-agent role and would love somebody to review/try it in real life as I currently work with modules not with roles. If everything is ok then I'll change zabbix-proxy role as well and then we'll merge.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix-agent role
zabbix-proxy role


D3DeFi (commented):
Patch in itself looks good to me, but it seems that it will break backwards compatibility for everyone that relies on zabbix-api right know.

I would go with something like this probably in the roles:
```
    login_user: "{{ zabbix_api_login_user | default(omit) }}"
    login_password: "{{ zabbix_api_login_pass | default(omit) }}"
```

Unfortunately, I also don't have an immediate way to test this as I am also relying primarily on modules in my use cases.

BGmot:
> Patch in itself looks good to me, but it seems that it will break backwards compatibility for everyone that relies on zabbix-api right know.
> 
> I would go with something like this probably in the roles:
> 
> ```
>     login_user: "{{ zabbix_api_login_user | default(omit) }}"
>     login_password: "{{ zabbix_api_login_pass | default(omit) }}"
> ```
> 
> Unfortunately, I also don't have an immediate way to test this as I am also relying primarily on modules in my use cases.

Honestly I see a mess with login/password variables in the role (too many of them assigning to each other without clear understanding which one is actually used) and in 2.0 we need to clean it up. My purpose was to save backward compatibility. Currently these variables need to be provided for the role to work correctly with API: `zabbix_api_server_url`, `zabbix_api_http_user` and `zabbix_api_http_password`. My PR re-uses these variables but in a bit different way by making them work with httpapi.
Proposed by you `login_user` and `login_password` are just old "pre-httpapi" parameters for all our modules, we are not using them any longer so I am not sure I understand the reasoning behind introducing these vars (even with omit) as it does not seem to bring any backward compatibility...

BGmot:
In Molecule tests we don't test API so I ran api.yml task manually providing minimum needed variables in main branch (using zabbix-api) and in my bgmot_roles_httpapi branch without changing any variables, it worked after my last commit https://github.com/ansible-collections/community.zabbix/pull/822/commits/382d4b6def197f25b198b6615de0e04bf7bae7b5

D3DeFi:
> > Patch in itself looks good to me, but it seems that it will break backwards compatibility for everyone that relies on zabbix-api right know.
> > I would go with something like this probably in the roles:
> > ```
> >     login_user: "{{ zabbix_api_login_user | default(omit) }}"
> >     login_password: "{{ zabbix_api_login_pass | default(omit) }}"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Unfortunately, I also don't have an immediate way to test this as I am also relying primarily on modules in my use cases.
> 
> Honestly I see a mess with login/password variables in the role (too many of them assigning to each other without clear understanding which one is actually used) and in 2.0 we need to clean it up. My purpose was to save backward compatibility. Currently these variables need to be provided for the role to work correctly with API: `zabbix_api_server_url`, `zabbix_api_http_user` and `zabbix_api_http_password`. My PR re-uses these variables but in a bit different way by making them work with httpapi. Proposed by you `login_user` and `login_password` are just old "pre-httpapi" parameters for all our modules, we are not using them any longer so I am not sure I understand the reasoning behind introducing these vars (even with omit) as it does not seem to bring any backward compatibility...

I am confused with the parameters as well. Because right now `login_{user,password}` is actually used for authenticating against Zabbix API at `zabbix_api_server_url`. As opposed to the `zabbix_api_http_{user,password}`, which are I think only used for HTTP basic auth that you can introduce before Zabbix via NginX/etc.. 

This is valid for modules, I would expect the same convention from the roles, but I am not sure at all. I would guess that this, for example,  should stop working with zabbix-api at all if merged:
```
- name: "API | Create host groups"
  community.zabbix.zabbix_group:
    host_group: "{{ zabbix_host_groups }}"
    state: "{{ zabbix_agent_hostgroups_state }}"
    validate_certs: "{{ zabbix_api_validate_certs|default(omit) }}"
    timeout: "{{ zabbix_api_timeout }}"
  when:
    - zabbix_api_create_hostgroup | bool
  register: zabbix_api_hostgroup_created
  until: zabbix_api_hostgroup_created is succeeded
  delegate_to: "{{ zabbix_api_server_host }}"
  become: false
  tags:
    - api
```

I understand push to deprecate zabbix-api, but if that is the case, then this patch cannot go into the main branch and needs to go to the 2.0.0-dev for now as I expect that we need to at least publish 1.9.0 with httpapi first and couple of other changes.

D3DeFi:
> In Molecule tests we don't test API so I ran api.yml task manually providing minimum needed variables in main branch (using zabbix-api) and in my bgmot_roles_httpapi branch without changing any variables, it worked after my last commit [382d4b6](https://github.com/ansible-collections/community.zabbix/commit/382d4b6def197f25b198b6615de0e04bf7bae7b5)

ofc this changes things that I wrote in the above comment

BGmot:
We don't use zabbix-api in this role anymore (I removed installing it).
Here https://github.com/ansible-collections/community.zabbix/pull/822/files#diff-9b394081adf0fae0d7f1e79bcb4db372adf082ed27f441419b881713b16175afR120 (starting line 117 in roles/zabbix_agent/tasks/main.yml) and here https://github.com/ansible-collections/community.zabbix/pull/822/files#diff-d722eb1d2f161071c11093c3875382316f4cc4b6bf5d37be35bbb3b5ac2e283aR82 (starting line 82 in roles/zabbix_agent/defaults/main.yml) I define variables to use httpapi. That's the whole point of this PR -) *our* roles use modules with *httpapi*. But if some user developed roles want to use zabbix-api (or just don't change from what they are) it is still possible for modules to use zabbix-api.

BGmot:
> So, just to confirm I understand it correctly:
>1) If new user downloads our collection and starts using this role, provided settings will default to using httpapi
Yes.

>2) Same for the existing user, except that zabbix-api will already be installed on their controller, but no longer used
Yes.

>3) If a user wants to (for some reason) continue using zabbix-api with 1.9.x release, can they overwrite ansible_network_os or something else to be served via zabbix-api?
Yes. 

>Since ansible.netcommon is now required dependency, we risk breaking functionality for the user that upgrades. I would at least print message in 1.9 that with 2.0, ansible.netcommon is required.
Do you mean in the role? Would you please advise how I can do it? Just a debug: module?

D3DeFi:
Thank you @BGmot for bearing with me! With your last comment I have no further objections. For me this is LGTM and we can merge.

is tag (WIP) still necessary? Does @mu1f407 and @pyrodie18 want to add something?

BGmot:
> Thank you @BGmot for bearing with me! With your last comment I have no further objections. For me this is LGTM and we can merge.
> 
> is tag (WIP) still necessary? Does @mu1f407 and @pyrodie18 want to add something?

No problem at all, please keep asking as many questions as you want.  It's not ready for a merge as I need to convert one more role - zabbix_proxy - just wanted to gather feedback on zabbix_agent first.

pyrodie18:
I'm good.  Life is slightly crazy right now so haven't had a chance to look at it.  Hoping it will slow down a little in the next few days so I can finish up what I've been doing on 2.0

D3DeFi (approved):
Ok, I am also good with this. I would suggest we move with 1.9 release in the near future

PR #842, opened 2022-11-16
Move fping locations to Distro vars

### LINKED ISSUE #812: Incorrect fping location on Debian derivatives
##### SUMMARY
The default value for fping is `/usr/sbin/fping` however for a while now Debian and derivatives have not placed an fping binary or symlink there. On Debian systems it should default to `/usr/bin/fping` and `/usr/bin/fping6`

In addition it appears there is a `zabbix_server_fpinglocation` variable available that could be overridden, but it is not documented.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix_server role

##### ANSIBLE VERSION
NA

##### CONFIGURATION
NA

Zabbix 6.2.3, Role version 1.8.0, Ubuntu 22.04

##### STEPS TO REPRODUCE
Deploy Zabbix server with the role but do not change the `zabbix_server_fpinglocation`. 

##### EXPECTED RESULTS
The fping location is set appropriately

##### ACTUAL RESULTS
It is not set appropriately leading to the error on Zabbix 

> At least one of '/usr/sbin/fping', '/usr/sbin/fping' must exist. Both are missing in the system.

evilhamsterman:
Since different distros put fping in different locations the variables should be specific to the distro. This fixes #812

### DIFF
--- a/changelogs/fragments/812-fping-binary-location.yml
+++ b/changelogs/fragments/812-fping-binary-location.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zabbix_server - move location of the fping(6) variables to distribution specific files (https://github.com/ansible-collections/community.zabbix/issues/812)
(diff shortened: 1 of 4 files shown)


pyrodie18 (changes requested):
Looks good.  Please also include [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments) (see examples in changelogs/fragments directory). After that, I'm happy with it. Thanks. 🙂

evilhamsterman:
That look good @pyrodie18?

D3DeFi (approved):
looks good to me :)

Issue #526, opened 2021-11-14
tls_accept and tls_conect needs to be the opposite 


maxxer:
... at least for the backend api!

##### SUMMARY
When configuring encrypted communication between proxy and server, there's a mismatch of tls accept/connect params between the config file and the definition created on the Zabbix backend.

In my test I encrypt communications between proxy and server but I allow unencrypted ones between agents and proxy. This is my config:
```
TLSConnect=psk
TLSAccept=unencrypted
```

![immagine](https://user-images.githubusercontent.com/240201/141673324-44412c1d-828a-425a-9d47-3ca88ca33834.png)

But when I configure a host with:

```
zabbix_proxy_name: Proxy test2021
zabbix_proxy_hostname: Proxy test2021
zabbix_proxy_tlspsk_secret: bf59807d14db9d6b8d6e44e9f199
zabbix_proxy_tlspskidentity: 1b3e346a11
```
this config produces the following in the backend:

![immagine](https://user-images.githubusercontent.com/240201/141673539-c457b995-d49c-47cb-b224-4a4e532f8c53.png)

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
- zabbix_agent
- zabbix_proxy

##### ANSIBLE VERSION

```
ansible [core 2.11.6]
  config file = /Users/maxxer/work/yetopen-ansible/ansible.cfg
  configured module search path = ['/Users/maxxer/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/4.8.0/libexec/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/maxxer/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.10.0 (default, Oct 13 2021, 06:44:31) [Clang 12.0.0 (clang-1200.0.32.29)]
  jinja version = 3.0.2
  libyaml = True
```

##### CONFIGURATION

```
DEFAULT_HOST_LIST(/Users/maxxer/work/yetopen-ansible/ansible.cfg) = ['/Users/maxxer/work/yetopen-ansible/hosts.yaml']
DEFAULT_MANAGED_STR(/Users/maxxer/work/yetopen-ansible/ansible.cfg) = This file is managed by YetOpen with Ansible.
Any manual change will be overwritten.%n
template: {file}
date: %Y-%m-%d %H:%M:%S
user: {uid}
host: {host}
[...]

PR #847, opened 2022-11-25
Swap tls connect/accept parameters when creating proxy

maxxer:
.. at least on the backend ansible-collections#526

##### SUMMARY

When configuring encrypted communication between proxy and server, there's a mismatch of tls accept/connect params between the config file and the definition created on the Zabbix backend.

In my test I encrypt communications between proxy and server but I allow unencrypted ones between agents and proxy. This is my config:

```
TLSConnect=psk
TLSAccept=unencrypted

```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

* zabbix_proxy

##### ADDITIONAL INFORMATION

more details on #526

### DIFF
--- a/changelogs/fragments/526-swap-tls-accept-connect-in-proxy.yaml
+++ b/changelogs/fragments/526-swap-tls-accept-connect-in-proxy.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zabbix_proxy - swap configuration of tls_accept and tls_connect
+  - zabbix_proxy - correctly provision tls_accept and tls_connect on Zabbix backend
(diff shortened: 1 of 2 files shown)


changelogs/fragments/526-swap-tls-accept-connect-in-proxy.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zabbix_proxy - swap configuration of tls_accept and tls_connect
```
BGmot, line 2:
```suggestion
  - zabbix_proxy - correctly provision tls_accept and tls_connect on Zabbix backend
```
Since this will be visible to end user I think we need to re-word it a bit.

BGmot (approved):

BGmot:
Thanks @maxxer !

PR #852, opened 2022-12-05
Update Debian.yml

derekpurdy:
Replacing shell command with ansible apt module  to perform clean and update as it was throwing error...

"Unsupported parameters for (ansible.legacy.command) module: warn. Supported parameters include: executable, strip_empty_ends, chdir, stdin, _uses_shell, creates, _raw_params, stdin_add_newline, argv, removes."

##### SUMMARY

Changes from ansible shell to ansible apt module, I was experiencing an issue as shown below. I believe this should be a proper fix for the error message.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_server/tasks/Debian

##### ADDITIONAL INFORMATION

```paste below

TASK [community.zabbix.zabbix_server : apt-get clean] ****************************************************************************************************************************************************************************************
fatal: [XXXXXXXX]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (ansible.legacy.command) module: warn. Supported parameters include: executable, removes, _raw_params, creates, strip_empty_ends, chdir, stdin_add_newline, stdin, argv, _uses_shell."}


```

### DIFF
--- a/roles/zabbix_server/tasks/Debian.yml
+++ b/roles/zabbix_server/tasks/Debian.yml
@@ -27,7 +27,6 @@
   when:
     - ansible_machine == "aarch64"
 
-
 - name: "Debian | Set some facts"
   set_fact:
     datafiles_path: /usr/share/zabbix-server-{{ zabbix_server_database }}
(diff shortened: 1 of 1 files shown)


mu1f407:
Hi @derekpurdy, thank you, nice find. The tests fails because the `clean` parameter of `ansible.builtin.apt` module was added in ansible-core 2.13 and we currently use 2.11 for molecule tests. I guess you use ansible-core 2.14 as the `warn` parameters was removed in this version (see ansible/ansible#77411). As we must currently support ansible-core versions older than 2.13, I think removing:
```
args:
    warn: false
```
alltogether should be enough to make it work on newer ansible-core version without breaking compatibility with olders. From the linked PR I see that from 2.11, the `warn` was `False` by default so it should be find to just remove it.

Could you please test it on your system with just `warn` removed? Thank you.

derekpurdy:
@mu1f407 confirmed that removal of args and warn removes the error.

mu1f407:
@derekpurdy It would be great if you could do it. Thank you very much.

Issue #855, opened 2022-12-09
Allow API call to Zabbix in webserver root directory (httpapi), currently double slash is added


Vanav:
Currently default is API call to Zabbix installed in path `http://example.com/zabbix`. If Zabbix is installed in root directory `https://zabbix.example.com`, then the best I can do is to specify `ansible_zabbix_url_path: ''`, but it will make call to `https://zabbix.example.com//api_jsonrpc.php`, notice the double slash. Please, make it possible to make API call to Zabbix in root directory without double slash (`https://zabbix.example.com/api_jsonrpc.php`).

Maybe default `/zabbix` is not the best also? It can be anything, depending on installation config, it is not defined it Zabbix packages or documentation. I'd prefer empty path as default.

Source: https://github.com/ansible-collections/community.zabbix/blob/85b298715a5ab0fb5ec6a6927c0e43df791084e0/plugins/httpapi/zabbix.py#L88

BGmot:
I'll fix double slash.
I made /zabbix default because when you install Zabbix from packages then it is installed in this path - look at 2f clause here https://www.zabbix.com/download?zabbix=6.2&os_distribution=ubuntu&os_version=22.04&components=server_frontend_agent&db=mysql&ws=apache
I can make it empty (like for Zabbix docker containers), let's vote?

Vanav:
I see. I've checked packages:

zabbix-nginx-conf - installed to root:
```
        root    /usr/share/zabbix;
        location / {
                try_files       $uri $uri/ =404;
        }
        location /assets {
                access_log      off;
                expires         10d;
        }
```
zabbix-apache-conf - installed to path:
```
# Define /zabbix alias, this is the default
<IfModule mod_alias.c>
    Alias /zabbix /usr/share/zabbix
</IfModule>
```

BGmot:
It's fixed in this commit https://github.com/ansible-collections/community.zabbix/commit/b3dedd6beaaf6d26a66ab474479348fa371a83ec but it is going to be a PR for to fix another much bigger problem, waiting on testing by others. If you want you can try that.

johansenha:
We are using zabbix as well configured on webroot and this seems to work using community version 1.9.1. 

https://github.com/ansible-collections/community.zabbix/blob/main/plugins/httpapi/zabbix.py#L102

I guess Issue can be closed

PR #861, opened 2022-12-13
Fix httpapi basic auth

### LINKED ISSUE #848: Collection 1.9.0 breaks HTTP Basic auth
##### SUMMARY

After upgrading the collection to 1.9.0, I can no longer run API transactions with Zabbix after it moved to httpapi.

In my current environment the setup is as follows:

- The HTTP server is joined to an IdM domain
- Basic Authentication is enabled (and required)
- Zabbix itself is not configured to read accounts from LDAP (as most users come from a trusted AD)

This worked fine with the previous implementation (where HTTP and API credentials could be supplied separately), but now it seems to be broken as the credentials are not passed properly.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

zabbix_agent role
##### ANSIBLE VERSION

```paste below
ansible [core 2.12.2] on RHEL8 (can't copy/paste from that system to here)
```

Target OS: RHEL8
Zabbix: 6.0 LTS

BGmot:
##### SUMMARY
Fixes #848, #855

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_agent and zabbix_proxy roles
All modules

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,12 @@ Or you include collection name `community.zabbix` in the playbook's `collections
         ...
 ```
 
+If Basic Authentication is required to access Zabbix server add following variables:
+```
+zabbix_api_http_user: "user"
+zabbix_api_http_password: "password"
+```
+
 ## Supported Zabbix versions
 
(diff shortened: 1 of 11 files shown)


D3DeFi (approved):
Nice patch!

evgeni:
@D3DeFi me? I have no powers in this repo.

PR #871, opened 2023-01-06
Allow using other hash methods for postgresql user

l00d3r:
##### SUMMARY
Added option for zabbix-proxy role to not hash the password md5 so we could use postgresql default of scram-sha-256.

Added a variable and edited the task to only apply when the default is changed 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
zabbix-proxy postgresql

##### ADDITIONAL INFORMATION
Added a new default as to not break existing installs

```paste below

```

### DIFF
--- a/changelogs/fragments/871.yml
+++ b/changelogs/fragments/871.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - zabbix_proxy_dbpassword - when zabbix_proxy_dbpassword_hash_method is set to anything other than md5 then do not hash the password with md5 so you could use postgresql scram-sha-256 hashing method
+  - zabbix_proxy_dbpassword_hash_method - add variable to control whether you want postgresql user password to be hashed with md5 or want to use db default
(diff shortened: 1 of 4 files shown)


D3DeFi (approved):
Looks good to me :) Thank you for the contribution!

l00d3r:
I’d be happy to but can you point me to how I can do that please? Like what are my next steps?

l00d3r:
I added a changelog fragment, I hope it is according to the standards

changelogs/fragments/871.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - zabbix_proxy_dbpassword - when zabbix_proxy_dbpassword_hash_method is set to anything other than md5 then do not hash the password with md5 so you could use postgresql scram-sha-256 hashing method
+  - zabbix_proxy_dbpassword_hash_method - add variable to control whether you want postgresql user password to be hashed with md5 or want to use db default
```
BGmot, line 3:
```suggestion
  - zabbix_proxy role - Add variable zabbix_proxy_dbpassword_hash_method to control whether you want postgresql user password to be hashed with md5 or want to use db default. When zabbix_proxy_dbpassword_hash_method is set to anything other than md5 then do not hash the password with md5 so you could use postgresql scram-sha-256 hashing method. 
```
I'd suggest following re-wording.

l00d3r:
I commited the suggested change

Thanks for helping me!

BGmot:
Thanks for your patience @l00d3r! I'll merge when all the tests pass.

BGmot (approved):

BGmot:
Thanks for the PR! And sorry for delay in merging.

PR #907, opened 2023-02-22
zabbix_user: several mailto: for email media type

### LINKED ISSUE #903: zabbix_user: add support for sendto to be a list (not only a string)
Hello!

##### SUMMARY

module zabbix_user treats medias' `sendto:` to be a string, whereas the zabbix has allowed this field to be a list since zabbix 4.0

sources:
* https://github.com/ansible-collections/community.zabbix/blob/a47fe01ab18954add074c8025f3b8f82d5b07ad0/plugins/modules/zabbix_user.py#L647
* https://www.zabbix.com/documentation/current/en/manual/api/reference/user/object
* https://www.zabbix.com/documentation/4.0/en/manual/api/reference/user/object
* https://www.zabbix.com/documentation/3.4/en/manual/api/reference/user/object

if a module users calls the module with a list, it leads to a zabbix server error due to the module's serialization to JSON

##### ISSUE TYPE
- Feature Idea
- Bugfix (well it's technically not a bug as no one ever claimed this was supported in the first place by the module, anyhow unsuspecting users might feel so)

##### COMPONENT NAME

zabbix_user

##### ADDITIONAL INFORMATION

```yaml
- name: create an user with a single media of type email which has multiple email addresses
  community.zabbix.zabbix_user:
    username: example
    usrgrps:
      - Guests
    passwd: password
    user_medias:
      - sendto:
        - my.first.email@example.com
        - my.second.email@example.com
```

BGmot:
##### SUMMARY
Fixes #903 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_user module


D3DeFi (approved):
Looks good! Not sure why third-party-check fails, but we will address that if the problem persists in the future

shk3bq4d:
Hello,

Thanks for the fix, but it's not 100% correct in its object's type verification.

in here:
https://github.com/ansible-collections/community.zabbix/blob/5ce4d227a49074e8af22b142c9f839ea365b49e3/plugins/modules/zabbix_user.py#L700

your intention is to verify that the `media type`'s type is email

what you are actually doing is checking that the `media type`'s name is email.

the zabbix wording is confusing, indeed, but again from https://www.zabbix.com/documentation/current/en/manual/api/reference/user/object

> _If type of [Media type](https://www.zabbix.com/documentation/current/en/manual/api/reference/mediatype/object#mediatype) is e-mail, values are represented as array. For other types of [Media types](https://www.zabbix.com/documentation/current/en/manual/api/reference/mediatype/object#mediatype), value is represented as a string._


I'm not sure how to improve your code since you probably would have to fetch all medias to check their actual type instead of relying on their name, which seems a bit heavy.

I caught the problem immediately because I have renamed the shipped by default media type `Email` to `Email HTML` and created a second one name `Email plain-text`.

Removing your type verification works, but it's probably not satisfactory either.

BGmot:
Can you please move this comment to the Issue and re-open it?

PR #920, opened 2023-03-15
update example

amg-web:
change examples to be less confusing with httpapi

##### SUMMARY

set_fact makes variables available until play end and makes following tasks failed due to change ansible_connection method

setting variables in task vars section change it for the current task only

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/plugins/modules/zabbix_host.py
+++ b/plugins/modules/zabbix_host.py
@@ -354,16 +354,6 @@ extends_documentation_fragment:
 '''
 
 EXAMPLES = r'''
-# Set following variables for Zabbix Server host in play or inventory
-- name: Set connection specific variables
-  set_fact:
-    ansible_network_os: community.zabbix.zabbix
-    ansible_connection: httpapi
-    ansible_httpapi_port: 80
-    ansible_httpapi_use_ssl: false
-    ansible_httpapi_validate_certs: false
-    ansible_zabbix_url_path: 'zabbixeu'  # If Zabbix WebUI runs on non-default (zabbix) path ,e.g. http://<FQDN>/zabbixeu
-
 # If you want to use Username and Password to be authenticated by Zabbix Server
 - name: Set credentials to access Zabbix Server API
   set_fact:
@@ -376,7 +366,17 @@ EXAMPLES = r'''
   set_fact:
     ansible_zabbix_auth_key: 8ec0d52432c15c91fcafe9888500cf9a607f44091ab554dbee860f6b44fac895
 
-- name: Create a new host or update an existing host's info
+- name: Create a new host or rewrite an existing host's info
+  vars:
+# Set following variables for Zabbix Server host in task
+    ansible_network_os: community.zabbix.zabbix
+    ansible_connection: httpapi
+    ansible_httpapi_port: 443
+    ansible_httpapi_use_ssl: true
+    ansible_httpapi_validate_certs: false
+    ansible_zabbix_url_path: 'zabbixeu'  # If Zabbix WebUI runs on non-default (zabbix) path ,e.g. http://<FQDN>/zabbixeu
+  become: false
+  delegate_to: zabbix-exaple-fqdn.org
   community.zabbix.zabbix_host:
     host_name: ExampleHost
     visible_name: ExampleName
@@ -436,6 +436,7 @@ EXAMPLES = r'''
     tls_psk_identity: test
     tls_connect: 2
     tls_psk: 123456789abcdef123456789abcdef12
+    force: false
 '''
 
 


BGmot:
I don't mind accepting this PR but it's more about how to use code from Ansible side, there is gazillion different ways to provide vars. If we do this for this module then we'll have to update all the modules' documentation too.
Opinions?

pyrodie18:
> I don't mind accepting this PR but it's more about how to use code from Ansible side, there is gazillion different ways to provide vars. If we do this for this module then we'll have to update all the modules' documentation too. Opinions?

I kind of agree.  I personally see three options:

1. Leave it as is in the modules and simply update the documents within the collection
2. Update it in all of the modules
3. Both.

And one other minor point, the word example in the the `delegate_to` is misspelled and

amg-web:
@BGmot 
>  there is gazillion different ways to provide vars. If we do this for this module then we'll have to update all the modules' documentation too. Opinions?

sure. but how many of them are single task level?
as I wrote in issues I was very confused by current docs version.

also about: force which is true by default . you can not simply add template or group with force: true, all host config will be replaced and lost. 

But many thanks for zabbix-api drop for the role. I do not want to build own awx-ee for only 1 extra module

plugins/modules/zabbix_host.py
```diff
@@ -376,7 +366,17 @@
...
+# Set task level following variables for Zabbix Server host in task
+  vars:
+    ansible_network_os: community.zabbix.zabbix
+    ansible_connection: httpapi
+    ansible_httpapi_port: 443
+    ansible_httpapi_use_ssl: true
+    ansible_httpapi_validate_certs: false
+    ansible_zabbix_url_path: 'zabbixeu'  # If Zabbix WebUI runs on non-default (zabbix) path ,e.g. http://<FQDN>/zabbixeu
+  become: false
+  delegate_to: zabbix-example-fqdn.org
```
dj-wasabi, line 379: You can use it differently by specifying a `ansible_host` variable. So `ansible_host: zabbix-example-fqdn.org` in the vars property.

amg-web: @dj-wasabi 
ok. try to do it.
usual inventory
hostname ansible_host=123.123.212.3

so if i use ansible_host for zabbix, it will breake all

dj-wasabi: I have this configured:
```
    ansible_user: "{{ zabbix_server_username }}"
    ansible_host: "{{ zabbix_server_hostname }}"
    ansible_httpapi_pass: "{{ zabbix_server_password }}"
    ansible_httpapi_use_ssl: true
    ansible_httpapi_validate_certs: false
    ansible_httpapi_port: 443
```
And all my Zabbix related tasks will work fine and have access to my Zabbix FQDN.

BGmot: `delegate_to:` and `ansible_host:` achieve essentially the same result (or maybe I don't know Ansible internals well enough).
Let's agree on something guys, we need to release new version with Zabbix 6.4 adoption and I'd love documentation "fixes" to go into new version.
@D3DeFi you did not say anything in this discussion -)

pyrodie18: I would argue that `ansible_host` should only be changed at the play level and not the task.  If I am in the process of adding a new system to the server, then the host I am working on is the client computer, but obviously, we have to modify the server a little bit to make that happen.  When we do that modification we're doing it for the benefit of the client, not the server so the client would be my ansible_host and I would delegate to the server.  Technically, I think either option would work, but I think that leaving the host alone for the duration of the play is the cleaner option.

dj-wasabi: The `ansible_host` is quite important for Ansible (Actually, the most important property) as this tells Ansible on which host the tasks should be executed. In my case I added it as a var to the task, as my playbook is executed on `hosta`. If you set it at host level, then all the tasks are executed on the host that is set in `ansible_host`. So it is very crusial to set the `ansible_host` on the task. And with setting `ansible_host` you can provide the fqdn (especially if FQDN is totally different than the hostname of the instances that Zabbix is running). Ansible runs in my case on ZAB1, ZAB2, but due to Nginx config only available on FQDN.

With the above, the `delegate_to` won't help me, as when a task is delegated the playbook needs 'ssh' access to that host. But, the `delegate_to` is maybe the safest option because of the above message.

My 2 cents. :)

amg-web: @dj-wasabi 
_delegate_to_ does not ssh in this case due to connection method **httpapi**
and task level _ansible_host_ is ok too. but we should not change it on higher level, like play, role or global

dj-wasabi:
I prefer the 2nd, as `ansible_user` is similar like the `ansible_host` variable and should also be configured per task and I don't know if there is an alernative for `ansible_user` like `ansible_host` has with the `delegate_to`.

pyrodie18:
> So what should we put?
> 
> ```
> # If your play is running on hosts other than Zabbix server itself then use delegate_to:
> - name: Create a new host or rewrite an existing host's info
> # Set task level following variables for Zabbix Server host in task
>   vars:
>     ansible_network_os: community.zabbix.zabbix
>     ansible_connection: httpapi
>     ansible_httpapi_port: 443
>     ansible_httpapi_use_ssl: true
>     ansible_httpapi_validate_certs: false
>     ansible_zabbix_url_path: 'zabbixeu'  # If Zabbix WebUI runs on non-default (zabbix) path ,e.g. http://<FQDN>/zabbixeu
>   become: false
>   delegate_to: zabbix-example-fqdn.org
> ```
> 
> Or
> 
> ```
> # If your play is running on hosts other than Zabbix server itself then use delegate_to:
> - name: Create a new host or rewrite an existing host's info
> # Set task level following variables for Zabbix Server host in task
>   vars:
>     ansible_network_os: community.zabbix.zabbix
>     ansible_connection: httpapi
>     ansible_httpapi_port: 443
>     ansible_httpapi_use_ssl: true
>     ansible_httpapi_validate_certs: false
>     ansible_zabbix_url_path: 'zabbixeu'  # If Zabbix WebUI runs on non-default (zabbix) path ,e.g. http://<FQDN>/zabbixeu
>     ansible_host: zabbix-example-fqdn.org
>   become: false 
> ```
> 
> ?

I'll go with number 2 also.

D3DeFi (changes requested):
Oh my bad, I've thought that changes introduced are to the repo's main README.md and not just to one module. Please let's also match it there.

Just a side node. What do we think about removing connection information from every module's examples and instead referencing single page somewhere in this repo that will clearly communicate it for all modules?

amg-web:
> Just a side node. What do we think about removing connection information from every module's examples and instead referencing single page somewhere in this repo that will clearly communicate it for all modules?

it's better to have at least one full example in a module reedme

amg-web:
@D3DeFi

> Oh my bad, I've thought that changes introduced are to the repo's main README.md and not just to one module. Please let's also match it there.

I edited on my pc this time adn could not update this PR. ( may be I missed something)
updated PR https://github.com/ansible-collections/community.zabbix/pull/929

BGmot:
Closing in favour of #929 
Thanks @amg-web !

PR #951, opened 2023-04-11
Use SSL for database connection when zabbix_server_dbtlsconnect is set

ChandlerSwift:
##### SUMMARY
I am setting up Zabbix with a database where SSL is enforced. Zabbix itself works fine with this, thanks to the `zabbix_server_dbtlsconnect` parameter, but a few of the playbook's setup tasks try to connect without TLS, which fails. This adds a parameter to enable TLS if the server would also be set to use it.

This isn't a perfect solution, but the `community.mysql` collection doesn't provide a way to use a TLS connection without setting one of `check_hostname`, `ca_cert`, `client_key`, or `client_hostname`, and we don't (can't, in our case) specify a cert/key. This _will_ potentially fail when the host presents a cert with the wrong hostname (which can happen if `zabbix_server_dbtlsconnect` is set to `required` rather than `verify_ca` or `verify_full`). However, there's not a way to enable TLS without setting one of these options without _also_ changing the MySQL collection.

There's some more background on this issue in the `community.mysql` collection: https://github.com/ansible-collections/community.mysql/issues/90

(I'm not convinced this is the right approach, though it does fix our particular use case! If there's another way to fix this that would be better, I'd appreciate the feedback and can try to update the PR if desired.)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_server role

### DIFF
--- a/roles/zabbix_server/tasks/mysql.yml
+++ b/roles/zabbix_server/tasks/mysql.yml
@@ -90,6 +90,7 @@
     login_password: "{{ zabbix_server_mysql_login_password | default(omit) }}"
     login_port: "{{ zabbix_server_mysql_login_port | default(omit) }}"
     login_unix_socket: "{{ zabbix_server_mysql_login_unix_socket | default(omit) }}"
+    check_hostname: "{{ zabbix_server_dbtlsconnect != '' }}"
   delegate_to: "{{ delegated_dbhost }}"
   register: install_mysql_version
   tags:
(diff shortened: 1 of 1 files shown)


ChandlerSwift:
Hey @pyrodie18, is there anything I can do to help get this moved towards being merged? Thanks!

pyrodie18:
It's not ready to be merged.  It failed the checks and looks like it was based on a pre 2.0 release.  Need to rebase and update your code and make sure that it passes the checks before we can merge it.

BGmot:
@pyrodie18 should we close it without merging?

pyrodie18:
Closed due to lack of corrections. Please feel free to make the required changes and create a new PR.

Issue #950, opened 2023-04-11
Calls to 6.4 server fail due to "user" parameter being deprecated/removed and httpapi variables won't work with set_fact per example


chrisallennc:
##### SUMMARY
Using server_url/login_user/login_password with modules to authenticate to 6.4.1 fails due to "user" parameter on user.login being deprecated/removed and "username" being required.

Furthermore, attempting to set variable "ansible_httpapi_pass" for httpapi with set_fact per examples fails. Since the parameter isn't being passed as "username" and that method is deprecated, we're attempted to switch to use httpapi.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
Noticed with proxy_info but appears to be present in other modules

##### ANSIBLE VERSION
```
ansible [core 2.14.3]
```

##### CONFIGURATION
```
```

##### OS / ENVIRONMENT / Zabbix Version
Target host is RHEL 7, host running ansible is Rocky Linux 8.5

##### STEPS TO REPRODUCE
```yaml
# This errors after upgrading Zabbix server to 6.4 because "user" is no longer a parameter for user.login.
- name: Get proxy
  community.zabbix.zabbix_proxy_info:
    server_url: "https://<server_fqdn>/zabbix/api_jsonrpc.php"
    login_user: zabbixapi
    login_password: "<password>"
    proxy_name: "<proxy_name>"
  register: zabbix_proxy
  delegate_to: localhost
```

```yaml
# This works but we need to be able to make ansible_httpapi_pass set dynamically with set_fact so it's not stored in the play code itself and is sourced from our secret management system.
- name: Get proxy
  community.zabbix.zabbix_proxy_info:
    proxy_name: "<proxy_name>"
  register: zabbix_proxy
  delegate_to: localhost
  vars:
    ansible_network_os: community.zabbix.zabbix
    ansible_connection: httpapi
    ansible_httpapi_port: 443
    ansible_httpapi_use_ssl: true
    ansible_httpapi_validate_certs: true
    ansible_host: <server_fqdn>
    ansible_user: zabbixapi
    ansible_httpapi_pass: "<password>"

# If we use set_fact for "ansible_httpapi_pass" (per the examples) above the task it fails.
- set_fact:
    ansible_httpapi_pass: "<password>"

- name: Get proxy
  community.zabbix.zabbix_proxy_info:
    proxy_name: "<proxy_name>"
  register: zabbix_proxy
  delegate_to: localhost
  vars:
[...]

Issue #946, opened 2023-04-09
Question: How to adapt to httpapi in the task?


duchenpaul:
##### SUMMARY
I would like to connect to my zabbix via httpapi, suggested by the warning

##### ISSUE TYPE
- Question

##### ANSIBLE VERSION

```paste below
ansible [core 2.14.4]
  config file = /app/ansible_code/ansible.cfg
  configured module search path = ['/home/pi/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.11/site-packages/ansible
  ansible collection location = /home/pi/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.11.2 (main, Mar 23 2023, 17:12:29) [GCC 10.2.1 20210110] (/usr/local/bin/python)
  jinja version = 3.1.2
  libyaml = True

```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT / Zabbix Version

Zabbix 6.2.3

##### STEPS TO REPRODUCE
My zabbix server and ansible controller runs on different servers, and this task is running against a zabbix agent host, it is used to register the host to the zabbix service.

My current working task:
```yaml
- name: Register {{ zabbix_host_hostname }} to zabbix server
  delegate_to: 127.0.0.1
  community.zabbix.zabbix_host:
    server_url: "{{ zabbix_server_url }}"
    login_user: "{{ zabbix_username }}"
    login_password: "{{ zabbix_password }}"
    description: "Host server: {{ inventory_hostname }}"
    force: yes
    host_name: "{{ zabbix_host_hostname }}"
    visible_name: "{{ zabbix_host_alias | default(zabbix_host_hostname) }}"
    # inventory_mode: automatic
    interfaces:
      - type: agent
        main: 1
        useip: 1
        ip: "{{ ansible_host }}"
        port: "10050"
    link_templates: "{{ zabbix_link_templates }}"
    host_groups: "{{ host_groups }}"
    timeout: 40
    proxy: "{{ zabbix_proxy | default() }}"
  retries: 10
  delay: 3
```

Tried to convert to httpapi connection:
```yaml
- name: Register {{ zabbix_host_hostname }} to zabbix server
  delegate_to: 127.0.0.1
  community.zabbix.zabbix_host:
    description: "Host server: {{ inventory_hostname }}"
    force: yes
[...]

Issue #894, opened 2023-02-07
Can't change zabbix_url_path


Zuko:
##### SUMMARY
Maybe I'm missed something obvious here, but I can't set `ansible_zabbix_url_path` variable…

My Zabbix Server is running in "root" context `https://<FQDN>/`, but role always try to connect `https://<FQDN>/zabbix`, I tried overwriting `ansible_zabbix_url_path` variable in "multiple" places, but as Ansible docs state that `extra vars (for example, -e "user=my_user")(always win precedence)` "always win" and after adding variable there it still doesn't work - something must be wrong.

Workaround:
Comment [../httpapi/zabbix.py#L117](https://github.com/ansible-collections/community.zabbix/blob/main/plugins/httpapi/zabbix.py#L117) line…

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
Zabbix_Agent Role version 1.9.1 / HTTPAPI Plugin

##### ANSIBLE VERSION
```
ansible [core 2.14.2]
  config file = None
  configured module search path = ['/home/zuko/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/zuko/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/zuko/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/zuko/.local/bin/ansible
  python version = 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### CONFIGURATION
```
CONFIG_FILE() = /home/zuko/repo/ansible/ansible.cfg
DEFAULT_FORKS(/home/zuko/repo/ansible/ansible.cfg) = 50
DEFAULT_HOST_LIST(/home/zuko/repo/ansible/ansible.cfg) = ['/home/zuko/repo/ansible-vm-inventory/hosts']
DEFAULT_VAULT_PASSWORD_FILE(/home/zuko/repo/ansible/ansible.cfg) = <REDACTED>
```

##### OS / ENVIRONMENT / Zabbix Version
Target: Debian 11
Zabbix Server: 6.0.13

##### STEPS TO REPRODUCE
Playbook:
```yaml
---
# code: language=ansible
- name: Install Zabbix Agent
  hosts: "{{ target_host | default([]) }}"

  handlers:
    - name: Import Handlers
      ansible.builtin.import_tasks:
[...]

PR #952, opened 2023-04-11
Correct minimum dependency for ansible.netcommon

flowerysong:
##### SUMMARY

Connection plugin options are broken in ansible.netcommon < 3.1.1, so settings like `ansible_zabbix_url_path` have no effect.

Possibly related issues: ~~#950~~ #946 #894

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

galaxy.yml

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -18,7 +18,7 @@ homepage: https://github.com/ansible-collections/community.zabbix
 issues: https://github.com/ansible-collections/community.zabbix/issues
 dependencies:
   ansible.windows: "*"
-  ansible.netcommon: "*"
+  ansible.netcommon: ">=3.1.1"
   ansible.posix: "*"
   community.general: "*"
   community.mysql: "*"


flowerysong:
`repo-sanity / Sanity (Ⓐdevel+py3.9)` failure is unrelated, `tests/sanity/ignore-2.16.txt` is missing (see #953).

duchenpaul:
Question: Can ansible_zabbix_url_path be something like `https://myzabbixfarm.com/`, so the API request will goes to these?

flowerysong:
> Question: Can ansible_zabbix_url_path be something like `https://myzabbixfarm.com/`, so the API request will go to these?

No, the connection plugin doesn't accept URLs, it accepts several individual settings that are combined to determine the URL.

For https, `ansible_httpapi_use_ssl: true`. `myzabbixfarm.com` should be the `inventory_hostname` or set in `ansible_host`. To indicate that there's no path component, `ansible_zabbix_url_path: ''`.

PR #953, opened 2023-04-11
CI: add ansible-core stable-2.15 to test matrix

flowerysong:
##### SUMMARY

ansible-core devel is 2.16 now, 2.15 is in beta.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

test

##### ADDITIONAL INFORMATION
https://github.com/ansible-collections/news-for-maintainers/issues/41

### DIFF
--- a/.github/workflows/plugins-integration.yml
+++ b/.github/workflows/plugins-integration.yml
@@ -31,6 +31,7 @@ jobs:
           - stable-2.12
           - stable-2.13
           - stable-2.14
+          - stable-2.15
           - devel
         python:
           - 3.9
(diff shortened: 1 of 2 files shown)


D3DeFi:
This is a good catch, but also a bit unfortunate, because @BGmot got rid of ignore-XY.txt files somewhere in #904, so this will have to wait until those changes are merged I think. We don't want to re-introduce this way of ignoring errors in modules.

flowerysong:
It's safe to merge it without waiting for that; ansible-test will also complain when ignores are present but not needed, so that will be a reminder to remove the file.

PR #976, opened 2023-04-30
Bgmot modules old versions 200

BGmot:
##### SUMMARY
In preparation for 2.0 release:
- Removing support of all versions of Zabbix < 6.0 from modules.
- Coding style: all strings in modules should be included in double quotes.


pyrodie18 (approved):
Great work!  Haven't seen anyone else doing reviews or popping up lately so if you want to go ahead and merge it, I'm good.

Issue #990, opened 2023-05-16
Collection 2.0.0 - Zabbix agent installation on windows fails


xdkreij:
https://github.com/ansible-collections/community.zabbix/issues/703
Somewhat similar issue.

##### SUMMARY
Any attempt to install the zabbix agent on Windows 11 (or 10 for that matter) fails. It contains 2 task related issues, both in main.yml of the zabbix_agent role

Issue 1: 'zabbix_valid_agent_versions' does not exists

fine.. so we set one up as a var `zabbix_valid_agent_versions=6.4`

Issue 2: float object has no attribute '10'

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix_agent/tasks/main.yml

##### ANSIBLE VERSION

```bash
ansible [core 2.14.5]
  config file = /home/ansible/projects/zabbix/ansible.cfg
  configured module search path = ['/home/ansible/projects/zabbix/library/modules']
  ansible python module location = /home/ansible/.local/lib/python3.9/site-packages/ansible
  ansible collection location = /home/ansible/projects/zabbix/collections
  executable location = /home/ansible/.local/bin/ansible
  python version = 3.9.13 (main, Nov 16 2022, 15:31:39) [GCC 8.5.0 20210514 (Red Hat 8.5.0-15)] (/usr/bin/python3.9)
  jinja version = 3.1.2
  libyaml = True
```

##### CONFIGURATION

```bash
CACHE_PLUGIN(/home/ansible/projects/zabbix/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/home/ansible/projects/zabbix/ansible.cfg) = cache
CACHE_PLUGIN_TIMEOUT(/home/ansible/projects/zabbix/ansible.cfg) = 300
COLLECTIONS_PATHS(/home/ansible/projects/zabbix/ansible.cfg) = ['/home/ansible/projects/zabbix/collecti>
CONFIG_FILE() = /home/ansible/projects/zabbix/ansible.cfg
DEFAULT_ACTION_PLUGIN_PATH(/home/ansible/projects/zabbix/ansible.cfg) = ['/home/ansible/projects/zabbix>
DEFAULT_CALLBACK_PLUGIN_PATH(/home/ansible/projects/zabbix/ansible.cfg) = ['/home/ansible/projects/zabb>
DEFAULT_FORKS(/home/ansible/projects/zabbix/ansible.cfg) = 5
[...]

PR #995, opened 2023-05-22
Handle Windows Agent Version

pyrodie18:
##### SUMMARY
Provides a fix to #990 by setting a default value specifically for Windows, and then ignoring other sanity checks.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- Agent Role

### DIFF
--- a/roles/zabbix_agent/tasks/main.yml
+++ b/roles/zabbix_agent/tasks/main.yml
@@ -5,13 +5,6 @@
   tags:
     - always
 
-- name: Default to Latest Version for Windows
-  set_fact:
-    zabbix_agent_version: "{{ zabbix_agent_windows_version }}"
-  when:
-    - zabbix_agent_version is not defined or zabbix_agent_version is none
-    - ansible_os_family == "Windows"
-
 - name: Determine Latest Supported Zabbix Version
   set_fact:
     zabbix_agent_version: "{{ zabbix_valid_agent_versions[ansible_distribution_major_version][0] | default(6.4) }}"
@@ -22,8 +15,6 @@
 - name: Set More Variables
   set_fact:
     zabbix_valid_version: "{{ zabbix_agent_version|float in zabbix_valid_agent_versions[ansible_distribution_major_version] }}"
-  when: ansible_os_family != "Windows"
-
   tags:
     - always
 
--- a/roles/zabbix_agent/tasks/main.yml
+++ b/roles/zabbix_agent/tasks/main.yml
@@ -7,7 +7,7 @@
 
 - name: Default to Latest Version for Windows
   set_fact:
-    zabbix_agent_version: 6.4
+    zabbix_agent_version: "{{ zabbix_agent_windows_version }}"
   when:
     - zabbix_agent_version is not defined or zabbix_agent_version is none
     - ansible_os_family == "Windows"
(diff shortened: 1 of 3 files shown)


BGmot (changes requested):
Please add this variable to roles/zabbix_agent/vars/Windows.yml

pyrodie18:
There you go @BGmot

BGmot (commented):
I am sorry, why do we need to pay special attention to WIndows and even introduce new variable `zabbix_agent_windows_version`, why can't we do in vars/Windows.yml exactly the same as in vars/Debian.yml so no need to modify tasks/main.yml at all. What am I missing here?

pyrodie18:
Honest answer is that I have no idea what comes out of `{{ ansible_distribution_major_version }}` for Windows and I don't have a box to test it on.

BGmot:
> Honest answer is that I have no idea what comes out of `{{ ansible_distribution_major_version }}` for Windows and I don't have a box to test it on.

I see. I'll try to test it today.

BGmot:
ansible_distribution_major_version is: 
- '10' for Windows Server 2016
- '10' for Windows Server 2019
- '10' for Windows Server 2022 
I don't have any non-Server versions of Windows available to test -(

pyrodie18:
That'll at least get this started...thanks and updated.

roles/zabbix_agent/tasks/main.yml
```diff
@@ -21,7 +21,9 @@
 - name: Stopping Install of Invalid Version
   fail:
     msg: Zabbix version {{ zabbix_agent_version }} is not supported on {{ ansible_distribution }} {{ ansible_distribution_major_version }}
-  when: not zabbix_valid_version
+  when:
+    - not zabbix_valid_version
+    - ansible_os_family != "Windows"
```
BGmot, line 26: Why do we need this additional condition? this should work for all OS's.

pyrodie18: Because I missed that line lol.  One minute

roles/zabbix_agent/tasks/main.yml
```diff
@@ -21,7 +21,8 @@
 - name: Stopping Install of Invalid Version
   fail:
     msg: Zabbix version {{ zabbix_agent_version }} is not supported on {{ ansible_distribution }} {{ ansible_distribution_major_version }}
-  when: not zabbix_valid_version
+  when:
+    - not zabbix_valid_version
```
BGmot, line 25:
```suggestion
```
I am sorry to be so picky, but I think we should not touch tasks/main.yml.

roles/zabbix_agent/tasks/main.yml
```diff
@@ -21,7 +21,8 @@
 - name: Stopping Install of Invalid Version
   fail:
     msg: Zabbix version {{ zabbix_agent_version }} is not supported on {{ ansible_distribution }} {{ ansible_distribution_major_version }}
-  when: not zabbix_valid_version
+  when:
```
BGmot, line 24:
```suggestion
  when: not zabbix_valid_version
```

BGmot (approved):

roles/zabbix_agent/vars/Windows.yml
```diff
@@ -1,2 +1,8 @@
 ---
 # vars file for zabbix_agent (Windows)
+zabbix_valid_agent_versions:
+  # Debian
```
masa-orca, line 4: Is it debian? Isn't it windows?

PR #996, opened 2023-05-23
Update Windows.yml

denser:
Due windows locales, correct detecting of arch.

In my locale windows arch cant detected in proper way, so we should search just 64 not full string

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
May be should fix not just agent role but others too.


pyrodie18:
> ##### COMPONENT NAME
> May be should fix not just agent role but others too.

The agent role is the only role that supports Windows.  Do me a favor and create a change fragment (https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)

pyrodie18 (approved):

PR #1026, opened 2023-06-30
Exclude ubuntu 18.04 from supported OS list

masa-orca:
##### SUMMARY

- Exclude ubuntu 18.04 from supported OS list.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

- all roles
##### ADDITIONAL INFORMATION

- Ubuntu 18.04 reaches the End of Standard Support in last month. [ubuntu-18-04-eol-for-devices](https://ubuntu.com/blog/ubuntu-18-04-eol-for-devices)

```paste below

```


pyrodie18:
While I don't disagree with this, I think I would also consider this a breaking change which would mean a major release.  @BGmot  agree/disagree?

BGmot:
We just stop testing against not supported OS so no change for users, I'd say it is not a breaking change. @D3DeFi what do you think?

BGmot:
You are right. Then I guess it's for a major release.

pyrodie18:
@masa-orca, we appreciate the contribution but unfortunately given that this is removing functionality, according to the ansible community rules, this would have to be a breaking change that requires a major release.  Given that we just did one of those a couple of months ago, I don't see the need to do one right away just to remove something.  According to the Zabbix roadmap (https://www.zabbix.com/roadmap#v7_0_LTS), they will be releasing 7.0 at the end of the year and I expect we'll likely do a major release then to depreciate support for 6.2 and 6.4.  At that time we'll also remove support for Ubuntu 18.04.

masa-orca:
Hum... I will reopen this pull request at the next major release. Is that okay?

pyrodie18:
Actually, you know what.  What's do this the right way.  I'm going to go ahead and re-open this ticket.  I would suggest that you go through and finish all of the work to actually depreciate Debian 10 though.  There are variables in in the vars files that are specific to Debian 10.  There's probably (although not sure off hand) some tasks too.  Go ahead and clean those up as well.  We'll get the PR in a state that its ready to go, but realize we won't actually pull it in until we get to work on 3.0.

That work for you?  Thanks...Troy

Issue #1051, opened 2023-07-25
Apache vhost bug on "zabix_web" role 


nat-0205:
##### SUMMARY

Hi, I've noticed that when the var "zabbix_web_create_vhost" is set to false, the "Apache | Enable Site (Debian Only)" module is executed in "zabbix_web/tasks/apache.yml", which is a module that should only be executed if the "zabbix_web_create_vhost" variable is set to true, which gives me an error when the role is executed.

I found a solution by modifying part of the module by adding ""and zabbix_web_create_vhost"" to the "when" condition which worked very well.

- name: "Apache | Enable Site (Debian Only)"
  file:
    src: "{{ zabbix_web_vhost_location }}"
    dest: /etc/apache2/sites-enabled/zabbix.conf
    state: link
    owner: "{{ zabbix_web_user }}"
    group: "{{ zabbix_web_group }}"
    mode: 0644
  become: true
  when: ansible_os_family == "Debian" and zabbix_web_create_vhost
  tags:
    - config

##### ISSUE TYPE

- Bug Report
- Bug solution

##### COMPONENT NAME

- ROLE : zabbix_web
- TASK: apache.yml
- MODULE: Apache | Enable Site (Debian Only) 

##### ANSIBLE VERSION

``ansible [core 2.14.1]
  python version = 3.10.6 [GCC 11.3.0] 
  jinja version = 3.1.2
```

##### OS / ENVIRONMENT / Zabbix Version

- OS version : Ubuntu 22.04.2 LTS

pyrodie18:
Great catch @nat-0205 , I'd encourage you to actually create the fix and put a PR in.

PR #1053, opened 2023-07-26
Update apache.yml

nat-0205:
**SUMMARY**

_HINT:_  

- Apache vhost bug on "zabix_web" role #1051 

_CHANGES:_  
 
- Changed "Apache | Enable Site (Debian Only)" module execution conditions by adding 'and zabbix_web_create_vhost'.

**ISSUE TYPE**

- Bugfix Pull Request

**COMPONENT NAME**

- ROLE : zabbix_web
- TASK: apache.yml
- MODULE: Apache | Enable Site (Debian Only)

**ADDITIONAL INFORMATION**

I made this modification because I noticed that when the var "zabbix_web_create_vhost" is set to false, the "Apache | Enable Site (Debian Only)"  module is executed in "zabbix_web/tasks/apache.yml", which is a module that should only be executed if the "zabbix_web_create_vhost" variable is set to true, which gives an error when the role is executed. With this small modification, the module no longer executes when the variable parameter is set to false.

### DIFF
--- a/roles/zabbix_web/tasks/apache.yml
+++ b/roles/zabbix_web/tasks/apache.yml
@@ -82,6 +82,6 @@
     group: "{{ zabbix_web_group }}"
     mode: 0644
   become: true
-  when: ansible_os_family == "Debian"
+  when: ansible_os_family == "Debian" and zabbix_web_create_vhost
   tags:
     - config


nat-0205:
@pyrodie18  I don't understand why the pull request failed, can you help me to understand please ?

pyrodie18:
Has nothing to do with you.  Its the sanity check.  The devel version of ansible requires python 3.10 or higher.  We're currently testing on 3.9.  If you look at my PR I'm updating it.  We'll get you taken care of soon.

PR #1054, opened 2023-07-27
Update python and docker compose for integration tests

pyrodie18:
Ansible Devel requires python 3.10 or higher but when we changed that the current method of installing docker-compose for the integration tests broke.


pyrodie18:
Not sure why Gateway on Rocky 9 is failing but has nothing to do with this.  Will have to look at that later.

PR #1059, opened 2023-08-08
Added additional includes for agent

### LINKED ISSUE #1056: Using multiple Include in zabbix_agent2.conf
Hi,

I'm trying to deploy zabbix-agent2 but I'm struggling using zabbix_agent2_include.
I'm deploying userparameters in zabbix-agent2.d/ and zabbix plugins are located (using ubuntu package) in zabbix-agent2.d/plugins.d/
In zabbix-agent role I can only define one path for the include parameter, if I choose /etc/zabbix/zabbix-agent2.d, my userparamerters files are loaded but the postgresql plugin isn't, so I can't use it.
One way to do that is to use 2 Include directives in zabbix-agent2.conf, for example :
```
Include=/etc/zabbix/zabbix_agent2.d/
Include=/etc/zabbix/zabbix_agent2.d/plugins.d/*.conf 
```

I can't find a way to do it using current role, Am I missing something or should zabbix_agent2_include be a list ?

Thanks !

pyrodie18:
##### SUMMARY
Added ability to include additional includes in the agent role.  Resolves #1056 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Agent

### DIFF
--- a/changelogs/fragments/additional_includes.yml
+++ b/changelogs/fragments/additional_includes.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - zabbix agent - Added capability to add additional configuration includes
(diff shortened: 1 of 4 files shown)


BGmot (approved):
Thanks!

PR #926, opened 2023-03-24
zabbix_agent: Support Raspbian

taskula:
##### SUMMARY

Adds Raspbian support for zabbix_agent. Fixes #57 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zabbix_agent

##### ADDITIONAL INFORMATION

Before:
```
TASK [community.zabbix.zabbix_agent : Debian | Install gpg key] ***************************************************************************************************************************************************
fatal: [pi]: FAILED! => {}

MSG:

The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'bullseye'

The error appears to be in '/home/ansible/.local/lib/python3.8/site-packages/ansible_collections/community/zabbix/roles/zabbix_agent/tasks/Debian.yml': line 26, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: "Debian | Install gpg key"
  ^ here
```

After
```
TASK [community.zabbix.zabbix_agent : Debian | Install gpg key] ***************************************************************************************************************************************************
changed: [pi]

TASK [community.zabbix.zabbix_agent : Debian | Check for zabbix repositories] *************************************************************************************************************************************
ok: [pi]

TASK [community.zabbix.zabbix_agent : Debian | Remove unecessary zabbix repositories] *****************************************************************************************************************************

TASK [community.zabbix.zabbix_agent : Debian | Installing deb-src repository Debian] ******************************************************************************************************************************
skipping: [pi]

TASK [community.zabbix.zabbix_agent : Debian | Installing deb repository Debian] **********************************************************************************************************************************
skipping: [pi]

TASK [community.zabbix.zabbix_agent : Debian | Installing deb-src repository Raspbian] ****************************************************************************************************************************
changed: [pi]

TASK [community.zabbix.zabbix_agent : Debian | Installing deb repository Raspbian] ********************************************************************************************************************************
changed: [pi]

...

TASK [community.zabbix.zabbix_agent : Debian | Installing zabbix-agent] *******************************************************************************************************************************************
changed: [pi]

TASK [community.zabbix.zabbix_agent : Debian | Installing zabbix-{sender,get}] ************************************************************************************************************************************
changed: [pi]
```

PR #1060, opened 2023-08-10
zabbix_agent - Support Raspbian

kivilahtio:
##### SUMMARY

Zabbix has official support for Raspberry.
Correctly selects the proper Raspbian repo.

See http://repo.zabbix.com/zabbix/6.4/ for the available packages, and notice that the only architecure-variant is for ubuntu. Simplify the repository url detection a bit to most easily match all rapository variants of supported OS within the Debian-family.

'ansible_lsb.id' differentiates between debian and raspbian, so use that instead of the 'ansible_distribution'

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zabbix_agent(2)

### DIFF
--- a/changelogs/fragments/926-support-raspbian.yml
+++ b/changelogs/fragments/926-support-raspbian.yml
@@ -1,0 +1,3 @@
+minor_changes:
+ - zabbix_agent - Add support for Raspbian
+
--- a/roles/zabbix_agent/tasks/Debian.yml
+++ b/roles/zabbix_agent/tasks/Debian.yml
@@ -5,12 +5,10 @@
   set_fact:
     zabbix_short_version: "{{ zabbix_agent_version | regex_replace('\\.', '') }}"
     zabbix_agent_apt_repository:
-      - "http://repo.zabbix.com/zabbix/{{ zabbix_agent_version }}/{{ ansible_distribution.lower() }}/"
+      - "http://repo.zabbix.com/zabbix/{{ zabbix_agent_version }}/{{ ansible_lsb.id.lower() }}/"
       - "{{ ansible_distribution_release }}"
       - "main"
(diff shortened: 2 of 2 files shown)


kivilahtio:
This replaces https://github.com/ansible-collections/community.zabbix/pull/926

BGmot:
Please take a look at failed tests, fix them and also provide changelog fragment.

pyrodie18:
Also please update the appropriate documentation and indicate that this is best effort since we don't have a way to test against Rasbian

derekpurdy:
I was actually just about to open a request for this. I am Running Raspberry Pi OS 64 bit and the zabbix agent role by default wants to use the debian branch which doesn't have support debian-arm64. I had customized my role to add support for raspbian builds of zabbix.

Glad to see it getting noticed! Thanks for the support guys.

derekpurdy:
Not sure the process, but wouldn't it be better to build a armhf version of the agent for Debian (vs only for raspbian) and the existing code should for the repo task would work, correct?

Current Raspberry Pi OS

`$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"`

kivilahtio:
> Please take a look at failed tests, fix them and also provide changelog fragment.

Thank you for taking a look at this.
However I have no experience in testing Ansible roles, and I don't have the time to take a look at it at the moment.

pyrodie18:
Ok, we appreciate the update.  We'll close the PR for right now but please feel free to re-open it if/when you get time.

Issue #1048, opened 2023-07-20
`zabbix_api_use_ssl: true` uses port 80 by default.


Bubbgump209:
##### SUMMARY
Registering an agent to the Zabbix server fails.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.zabbix.zabbix_agent

##### ANSIBLE VERSION
```
ansible [core 2.14.3]
  config file = /Users/someuser/.ansible.cfg
  configured module search path = ['/Users/someuser/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.11/site-packages/ansible
  ansible collection location = /Users/someuser/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.11.4 (main, Jun 20 2023, 16:51:49) [Clang 14.0.0 (clang-1400.0.29.202)] (/usr/local/opt/python@3.11/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

##### CONFIGURATION

```
CONFIG_FILE() = /Users/someuser/.ansible.cfg
DEFAULT_HOST_LIST(/Users/someuser/.ansible.cfg) = ['/Users/someuser/ansible/inventory']
DEFAULT_INVENTORY_PLUGIN_PATH(/Users/someuser/.ansible.cfg) = ['/Users/someuser/{{ ANSIBLE_HOME ~ "/plugins/inventory', '/usr/share/ansible/plugins/inventory" }}']
HOST_KEY_CHECKING(/Users/someuser/.ansible.cfg) = False
INVENTORY_ENABLED(/Users/someuser/.ansible.cfg) = ['ini', 'host_list', 'script', 'auto', 'yaml', 'toml']

```

##### OS / ENVIRONMENT / Zabbix Version

Tried  to run against target hosts on Debian 11 and 12 VM (cloud image) and LXC. 

Zabbix server 6.0.17 LTS.  

community.zabbix 2.0.1 and 2.1.0 with both netcommon 4.1.0 and 5.1.2.

##### STEPS TO REPRODUCE

Command
```
ansible-playbook  InstallZabbixLXC.yml --extra-vars "host=192.168.73.112"
```
InstallZabbixLXC.yml 
```
- name: Install the Zabbix agent and register with the Zabbix server
  hosts: "{{ hostvars['localhost']['fqdn'] | default(host) }}"
  tasks:
[...]

PR #1064, opened 2023-08-11
Updated Default API Port

pyrodie18:
##### SUMMARY
Updated default value of `zabbix_api_server_port` based on the value of 'zabbix_api_use_ssl'

Related to #1048 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Proxy and agent roles

### DIFF
--- a/docs/ZABBIX_AGENT_ROLE.md
+++ b/docs/ZABBIX_AGENT_ROLE.md
@@ -246,7 +246,7 @@ Host encryption configuration will be set to match agent configuration.
 
 * `zabbix_api_server_host`: The IP or hostname/FQDN of Zabbix server. Example: zabbix.example.com
 * `zabbix_api_use_ssl`: Is SSL required to connect to the Zabbix API server?  Default: `false`
-* `zabbix_api_server_port`: TCP port to use to connect to Zabbix server. Example: 8080
+* `zabbix_api_server_port`: 80 if `zabbix_api_use_ssl` is `false` and 443 if `true` (Default) TCP port to use to connect to Zabbix server. Example: 8080
 * `zabbix_api_login_user`: Username of user which has API access.
 * `zabbix_api_login_pass`: Password for the user which has API access.
 * `zabbix_api_http_user`: The http user to access zabbix url with Basic Auth (if your Zabbix is behind a proxy with HTTP Basic Auth).
--- a/docs/ZABBIX_PROXY_ROLE.md
+++ b/docs/ZABBIX_PROXY_ROLE.md
@@ -269,7 +269,7 @@ These variables need to be overridden when you want to make use of the Zabbix AP
 
 * `zabbix_api_server_host`: The IP or hostname/FQDN of Zabbix server. Example: zabbix.example.com
 * `zabbix_api_use_ssl`: Is SSL required to connect to the Zabbix API server?  Default: `false`
-* `zabbix_api_server_port`: TCP port to use to connect to Zabbix server. Example: 8080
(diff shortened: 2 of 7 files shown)


docs/ZABBIX_AGENT_ROLE.md
```diff
@@ -246,7 +246,7 @@ Host encryption configuration will be set to match agent configuration.
 
 * `zabbix_api_server_host`: The IP or hostname/FQDN of Zabbix server. Example: zabbix.example.com
 * `zabbix_api_use_ssl`: Is SSL required to connect to the Zabbix API server?  Default: `false`
-* `zabbix_api_server_port`: TCP port to use to connect to Zabbix server. Example: 8080
+* `zabbix_api_server_port`: 80 if `zabbix_api_use_ssl` is `false` and 443 if `true` (Default) TCP port to use to connect to Zabbix server. Example: 8080
```
BGmot, line 249: I think default for `zabbix_api_use_ssl` it `false`

docs/ZABBIX_PROXY_ROLE.md
```diff
@@ -269,7 +269,7 @@ These variables need to be overridden when you want to make use of the Zabbix AP
 
 * `zabbix_api_server_host`: The IP or hostname/FQDN of Zabbix server. Example: zabbix.example.com
 * `zabbix_api_use_ssl`: Is SSL required to connect to the Zabbix API server?  Default: `false`
-* `zabbix_api_server_port`: TCP port to use to connect to Zabbix server. Example: 8080
+* `zabbix_api_server_port`: 80 if `zabbix_api_use_ssl` is `false` and 443 if `true` (Default) TCP port to use to connect to Zabbix server. Example: 8080
```
BGmot, line 272: I think default for `zabbix_api_use_ssl` it `false`

roles/zabbix_agent/defaults/main.yml
```diff
@@ -68,7 +68,7 @@ zabbix_repo_yum:
 
 # Zabbix API stuff
 zabbix_api_server_host: localhost
-zabbix_api_server_port: 80
+# zabbix_api_server_port: 80
```
BGmot, line 71: I like having all the vars in defaults/main.yml, is there a reason you want to remove this line?

roles/zabbix_proxy/defaults/main.yml
```diff
@@ -112,7 +112,7 @@ zabbix_proxy_vmwarefrequency: 60
 # Zabbix API stuff
 zabbix_api_server_host: localhost
 zabbix_api_use_ssl: false
-zabbix_api_server_port: 80
+# zabbix_api_server_port: 80
```
BGmot, line 115: I like having all the vars in defaults/main.yml, is there a reason you want to remove this line?

roles/zabbix_agent/defaults/main.yml
pyrodie18: Because if you look at line 28 in the agent role, if `zabbix_api_server_port` is defined, it will always use that one.  It will only default if its not defined (to allow someone to override the port)

roles/zabbix_proxy/defaults/main.yml
pyrodie18: See comment above

docs/ZABBIX_PROXY_ROLE.md
pyrodie18: So what I was trying to say is that those two actions (combined) is the default behavior, or you can override it to whatever you want.  How would you word it?

PR #1075, opened 2023-08-30
Add module for getting events using host groups

meis4h:
##### SUMMARY

Add the module zabbix_group_events_info for getting problems by specifying a hostgroup instead of a host.
This is useful if not all hosts that need to be checked are available in the inventory for example with a host discovery in Zabbix.

This module is basically just a combined version of the existing zabbix_host_events_info and zabbix_group_info modules.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

zabbix_group_events_info


BGmot (commented):
Please add:
- tests for this new module
- CHANGELOG fragment

BGmot:
Perfect! thank you very much. One more ask though - at the end of your test please delete resources you created - host group, host, template. If we don't do a clean up in the tests then it might affect results of other tests.

meis4h:
Makes sense, I'll add some cleanup tasks for that.
Is the changelog fragment fine like that or do I need to add anything else there?

BGmot:
> Makes sense, I'll add some cleanup tasks for that. Is the changelog fragment fine like that or do I need to add anything else there?

Changelog fragment is ok, thank you!
Once all the test complete I will merge.

PR #1103, opened 2023-10-18
Updated Version Number

pyrodie18:
Changed version number from 2.2.0 to v2.2.0


natemccallum:
I'm getting this error with the "v" in front of 2.2.0
`ERROR! Unexpected Exception, this is probably a bug: Non integer values in LooseVersion ('v2.2.0')`
without the "v" all seems ok.

PR #1124, opened 2023-11-13
Remove EOL Ansible versions from test matrix.

BGmot:
##### SUMMARY
Remove EOL Ansible versions from test matrix.

##### ISSUE TYPE
- Maintenance Pull Request


pyrodie18:
Would argue that this is a breaking change and would need to go into 3.0.0 which is actually where I already made the change (https://github.com/ansible-collections/community.zabbix/commit/29090f2ed49598b3528e3daeb68fd5c3790fb26b)

BGmot:
opps... sorry.

PR #1157, opened 2024-01-05
Add template_yaml parameter to zabbix_template module

masa-orca:
##### SUMMARY

- Adding template_yaml parameter to zabbix_template module.
- Using [configuration.importcompare](https://www.zabbix.com/documentation/current/en/manual/api/reference/configuration/importcompare) to compare  template file(not only yaml but also json and xml) with Zabbix current configurations.
- Removing unused functions.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

- zabbix_template
##### ADDITIONAL INFORMATION
- A type of template_json is not `file` but `json`. Therefore, I cloudn't combine `template_XXX` parameter to a parameter.
- I cloudn't understand that `Filter the template json to contain only the keys we will update` in `filter_template` function is needed or not. Let me know if need.

```paste below

```

### DIFF
--- a/changelogs/fragments/1157-add_template_yaml.yml
+++ b/changelogs/fragments/1157-add_template_yaml.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - zabbix_template - Add template_yaml parameter.
--- a/plugins/modules/zabbix_template.py
+++ b/plugins/modules/zabbix_template.py
@@ -33,17 +33,21 @@ options:
         description:
             - JSON dump of templates to import.
             - Multiple templates can be imported this way.
-            - Mutually exclusive with I(template_name) and I(template_xml).
+            - Mutually exclusive with I(template_name) and I(template_xml) and I(template_yaml).
         required: false
         type: json
     template_xml:
         description:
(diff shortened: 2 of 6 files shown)


flowerysong:
I don't have the time to test it right now, but this shouldn't be necessary.

The point of the `json` argument type is that it ensures that what the module receives is a consistent format regardless of whether it started as a JSON string or a data structure, so instead of complicating the module interface all you should have to do is convert the file from YAML:

```yaml
    - name: Import Zabbix template from YAML file.
      community.zabbix.zabbix_template:
        template_json: "{{ lookup('file', 'template4_60_higher.yaml') | from_yaml }}"
        state: present
      register: import_template_yaml
```

masa-orca:
@flowerysong Thank you fo your response. I understood it is not necessary. However, it is a temporally workaround that you mentioned way, isn't it?

If this pull request is waste, I'll close it.

BGmot (approved):

BGmot (changes requested):
Changelog fragment please?

BGmot (approved):

Issue #1110, opened 2023-10-27
zabbix_host module: updating host inventory data causes IPMI values to be reset


Stivi:
##### SUMMARY
Using zabbix_host module to update a host inventory data produces a bug because IPMI values (ipmi_authtype, ipmi_privilege, ipmi_username, ipmi_password) are set to default.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix_host

##### ANSIBLE VERSION

```paste below
ansible [core 2.13.10]
  config file = /home/ansible/playbooks/ansible.cfg
  configured module search path = ['/home/ansible/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ansible/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/ansible/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/ansible/.local/bin/ansible
  python version = 3.8.10 (default, Jun 22 2022, 20:18:18) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True
```

##### CONFIGURATION

```paste below
DEFAULT_HOST_LIST(/home/ansible/playbooks/ansible.cfg) = ['/home/ansible/playbooks/hosts']
DEFAULT_REMOTE_USER(/home/ansible/playbooks/ansible.cfg) = ansible
HOST_KEY_CHECKING(/home/ansible/playbooks/ansible.cfg) = False
```

##### OS / ENVIRONMENT / Zabbix Version
```
Operating System: Ubuntu 20.04.4 LTS
Zabbix Version: 6.0.8
```

##### STEPS TO REPRODUCE
Run playbook

```yaml
- name: Update an existing host's info
  vars:
    ansible_zabbix_auth_key: 1a921b00a0b482a9e26a45e6fbea7dad
    ansible_network_os: community.zabbix.zabbix
    ansible_connection: httpapi
    ansible_httpapi_port: 443
    ansible_httpapi_use_ssl: true
    ansible_httpapi_validate_certs: true
    ansible_zabbix_url_path: ""
  become: false
  delegate_to: zabbix.hosting.ecnl.ru
  community.zabbix.zabbix_host:
    host_name:  "{{ ansible_facts['nodename'] }}"
    inventory_mode: manual
    inventory_zabbix:
[...]

PR #1162, opened 2024-01-11
Don't reset IPMI setting when update inventory

masa-orca:
##### SUMMARY

- Don't reset IPMI setting when update inventory
- Fixes https://github.com/ansible-collections/community.zabbix/issues/1110

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

zabbix_host module

### DIFF
--- a/changelogs/fragments/1162-do_not_update_ipmi_options.yml
+++ b/changelogs/fragments/1162-do_not_update_ipmi_options.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zabbix_host - Don't reset IPMI setting when update inventory
+  - zabbix_host - Don't reset IPMI setting when update inventory data of a host
--- a/changelogs/fragments/1162-do_not_update_ipmi_options.yml
+++ b/changelogs/fragments/1162-do_not_update_ipmi_options.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zabbix_host - Don't reset IPMI setting when update inventory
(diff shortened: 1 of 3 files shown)


changelogs/fragments/1162-do_not_update_ipmi_options.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zabbix_host - Don't reset IPMI setting when update inventory
```
BGmot, line 2: - zabbix_host - Don't reset IPMI setting when update inventory data of a host

BGmot:
I don't understand why you treat let's say `ipmi_privilege` and `ipmi_username` differently? why some of them checked for being in `zabbix_host_obj` and some not?

BGmot (approved):

masa-orca:
Thank you for reviewing. 

> `ipmi_privilege` and `ipmi_username` differently?

It is no meaning. I add checks about the parameters.

> why some of them checked for being in zabbix_host_obj and some not?

- We need to set the minimum parameters.
- `reseting IPMI setting` was coused by unnecessary parameters.
- In current source code, we cannot update value of string parameters (e.g. `ipmi_username`) with empty string.
  - To evaluate empty string result is false, so, the module skips adding parameter.
```python
test_string = ""
if test_string:
    print('Update a parameter')  # This print function is never called.
```

changelogs/fragments/1162-do_not_update_ipmi_options.yml
masa-orca: Thanks. I have changed to the sentence which you wrote.

BGmot:
Thank you very much @masa-orca -)

PR #1186, opened 2024-02-26
Make Zabbix repository gpg key URL customizable

jtekuelve:
##### SUMMARY

The proposed change introduces a variable in the agent, javagateway, proxy, server, and web role, which replaces the hardcoded string in the "Debian | Download gpg key" task. The default value of the "zabbix_repo_deb_gpg_key_url" variable is the previously hardcoded string: http://repo.zabbix.com/zabbix-official-repo.key

Additionally, a second variable "zabbix_repo_deb_include_deb_src" with default value "true" was introduced to toggle adding deb-src in the "Debian | Installing repository {{ ansible_distribution }}" task. 

Rationale: We are deploying Zabbix via Ansible in an environment without Internet access. We are using our own package mirror, which is configurable via the "zabbix_repo_deb_url". Unfortunately, without the proposed changes, we are unable to change the download location of the gpg key from the official repo to our own mirror. The second change, making deb-src optional is another change that makes the role easier to use with a custom package mirror. In our testing, it turned out that the deb-src repo was not required. Overall, both changes should not change the current functionality given that only hardcoded parts were replaced by variables with default values.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
roles: zabbix_agent, zabbix_javagateway, zabbix_proxy, zabbix_server, and zabbix_web

##### ADDITIONAL INFORMATION
I was unsure if I set the variables at the right location (currently defaults/main.yml) of each role. One could also argue to put them in vars/Debian.yml of each role. In the current repo, the defaults already contained some os dependent vars, for example, the # Yum/APT  Variables section, therefore I added them there. I am always open to moving or renaming them. For now, I stuck with the naming scheme that prefixes the role name, but it could be confusing since, for example, "zabbix_repo_deb_url" does not stick to this naming scheme and is kind of related.

### DIFF
--- a/changelogs/fragments/1186-custom_gpg_key_url.yml
+++ b/changelogs/fragments/1186-custom_gpg_key_url.yml
@@ -1,7 +1,8 @@
 minor_changes:
-  - agent, javagateway, proxy, server, and web role - introduced default variable zabbix_<role>_gpg_key_url with value http://repo.zabbix.com/zabbix-official-repo.key
-  - agent, javagateway, proxy, server, and web role - used zabbix_<role>_gpg_key_url in "Debian | Download gpg key" instead of hardcoded url
+  - agent, javagateway, proxy, server, and web role - introduced default variable zabbix_repo_deb_gpg_key_url with value http://repo.zabbix.com/zabbix-official-repo.key
+  - agent, javagateway, proxy, server, and web role - used zabbix_repo_deb_gpg_key_url in "Debian | Download gpg key" instead of hardcoded url
   - agent, javagateway, proxy, server, and web role - added the http_proxy and https_proxy environment variables to "Debian | Download gpg key" analog to other tasks
-  - agent, javagateway, proxy, server, and web role - introduced default variable zabbix_<role>_include_deb_src with value true
-  - agent, javagateway, proxy, server, and web role - used variable zabbix_<role>_include_deb_src in "Debian | Installing repository" to determine whether deb-src should be added to /etc/apt/sources.list.d/zabbix.sources
+  - agent, javagateway, proxy, server, and web role - introduced default variable zabbix_repo_deb_include_deb_src with value true
+  - agent, javagateway, proxy, server, and web role - used variable zabbix_repo_deb_include_deb_src in "Debian | Installing repository" to determine whether deb-src should be added to /etc/apt/sources.list.d/zabbix.sources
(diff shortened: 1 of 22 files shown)


pyrodie18 (changes requested):
Should have left this earlier but please update the readme for the roles with the new variables and then you're good to go.

jtekuelve:
After updating the readme I noticed that the other deb repository vars all start with zabbix_repo_deb. I adopted the same prefix for my variables.

changelogs/fragments/1186-custom_gpg_key_url.yml
```diff
@@ -0,0 +1,8 @@
+minor_changes:
+  - agent, javagateway, proxy, server, and web role - introduced default variable zabbix_repo_deb_gpg_key_url with value http://repo.zabbix.com/zabbix-official-repo.key
+  - agent, javagateway, proxy, server, and web role - used zabbix_repo_deb_gpg_key_url in "Debian | Download gpg key" instead of hardcoded url
+  - agent, javagateway, proxy, server, and web role - added the http_proxy and https_proxy environment variables to "Debian | Download gpg key" analog to other tasks
+  - agent, javagateway, proxy, server, and web role - introduced default variable zabbix_repo_deb_include_deb_src with value true
+  - agent, javagateway, proxy, server, and web role - used variable zabbix_repo_deb_include_deb_src in "Debian | Installing repository" to determine whether deb-src should be added to /etc/apt/sources.list.d/zabbix.sources
+  - agent, javagateway, proxy, server, and web role - removed superfluous slash in zabbix_gpg_key of the Debian vars and renamed key to zabbix-repo instead of zabbix-official-repo
+  - agent, javagateway, proxy, server, and web role - updated readme with the two new variables "zabbix_repo_deb_gpg_key_url" and "zabbix_repo_deb_include_deb_src"
```
BGmot, line 8:
```suggestion
```
Updating README should not be mentioned in Changelog, it's just natural (the way it should be done).

roles/zabbix_proxy/defaults/main.yml
```diff
@@ -38,6 +38,7 @@ zabbix_proxy_tls_config:
   cert: "certificate"
 zabbix_proxy_version_minor: "*"
 
+
```
BGmot, line 41:
```suggestion
```

BGmot (changes requested):

pyrodie18 (approved):

BGmot:
@pyrodie18 you merged without waiting for test to complete -)))

pyrodie18:
It had completed the first time and there were no code significant changes so ya but sorry

BGmot:
No worries -)

PR #1188, opened 2024-03-01
Fixed IPMI authentication algorithm default setting

balakine:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
roles/zabbix_agent

##### ADDITIONAL INFORMATION
zabbix_agent_ipmi_authtype description and default value are duplicating those of zabbix_agent_ipmi_privilege. This is obviously wrong. See [Host object API documentation](https://www.zabbix.com/documentation/current/en/manual/api/reference/host/object). You have it correct in [plugins/modules/zabbix_proxy_info.py](https://github.com/ansible-collections/community.zabbix/blob/main/plugins/modules/zabbix_proxy_info.py#L99). To reproduce the issue assign a zabbix_agent role to a host, don't set IPMI variables, and observe IPMI field in Zabbix to get set to MD5, not left as default.

It is hypothetically possible that someone might have relied on the wrong default setting and didn't set it explicitly in their inventory, so _technically_ it's a breaking change.

### DIFF
--- a/changelogs/fragments/1188-agent_ipmi_authtype_fix.yml
+++ b/changelogs/fragments/1188-agent_ipmi_authtype_fix.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zabbix_agent - Fixed IPMI authentication algorithm default setting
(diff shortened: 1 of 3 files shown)


BGmot (approved):

BGmot:
Thanks @balakine !
@pyrodie18 please review and merge if no concerns.

PR #1218, opened 2024-05-03
Ci

eb4x:
##### SUMMARY
This attempts to fix some shortcomings in the CI

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
molecule, github-actions

##### ADDITIONAL INFORMATION
It would be good if this could get merged, the change to `molecule/requirements.txt` results in a lot of unnecessary testing of other roles during my refactoring of the code.

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -68,6 +68,14 @@ jobs:
 
       - name: Build the collection
         run: |
+          # Pin versions to speed up CI
+          sed -i 's/ansible\.windows:\s*"\*"/ansible.windows: "2.3.0"/' galaxy.yml
+          sed -i 's/ansible\.netcommon:\s*"\*"/ansible.netcommon: "6.1.1"/' galaxy.yml
+          sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "1.5.4"/' galaxy.yml
+          sed -i 's/community\.general:\s*"\*"/community.general: "8.6.0"/' galaxy.yml
+          sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.9.0"/' galaxy.yml
+          sed -i 's/community\.postgresql:\s*"\*"/community.postgresql: "3.4.0"/' galaxy.yml
+
           collection_file=$( basename $(ansible-galaxy collection build -f | awk -F" " '{print $NF}'))
           echo "COLLECTION_FILE=$collection_file" >> $GITHUB_ENV
 
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -72,7 +72,7 @@ jobs:
           echo "COLLECTION_FILE=$collection_file" >> $GITHUB_ENV
 
       - name: Install the collection
-        run: ansible-galaxy collection install $COLLECTION_FILE
+        run: ansible-galaxy collection install -vvvv $COLLECTION_FILE
 
       - name: Run role tests
         run: >-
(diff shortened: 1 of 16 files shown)


eb4x:
@pyrodie18

Any idea or guesses as to why the `molecule verify` step fails? It spits out `Could not resolve host: zabbix-web-rockylinux9`, but I don't see how it was able to resolve before before the namechange of the containers?

If I `docker exec -it zabbix-web-rockylinux9 /bin/bash` into the container and `ping zabbix-web-rockylinux9`, it resolves. Is the verify step happening on the controller, and do we supply some /etc/hosts overrides any other thing that needs an update?

molecule/requirements.txt
```diff
@@ -3,8 +3,8 @@ ansible==7.0.0
 ansible-compat==3.0.0
 ansible-core==2.14.2
 docker==6.1.3
-molecule==4.0.4
-molecule-docker==2.1.0
+molecule<5
+molecule-docker @ git+https://github.com/ansible-community/molecule-docker@main
```
pyrodie18, line 7: Is there a reason to change the source to straight to github?

.github/workflows/web.yml
```diff
@@ -68,11 +68,19 @@ jobs:
 
       - name: Build the collection
         run: |
+          # Pin versions to speed up CI
+          sed -i 's/ansible\.windows:\s*"\*"/ansible.windows: "2.3.0"/' galaxy.yml
```
pyrodie18, line 72: Why do this?  In my mind we should be testing against what we publish in the galaxy.yml file otherwise we may have an issue of experiencing/not experiencing failures that we should/shouldn't during CI/CD testing.  Not to mention having to change versions in 5 places instead one 1.

pyrodie18 (changes requested):

molecule/requirements.txt
eb4x: Yes, there's a fix that went in after 2.1.0, [cb1c1c31a4a468fb243fc2000656504a3ca00bb3](https://github.com/ansible-community/molecule-docker/commit/cb1c1c31a4a468fb243fc2000656504a3ca00bb3)
that fixes a bug where the molecule.yml would get overwritten during image building. (I introduced the simple image-building to get python3 on centos7. But I think it's a good way for us to always stay current on all distribution images, instead of the geerlingguy images)

.github/workflows/web.yml
eb4x: This part where galaxy looks up every single version, for every single dependency where we use `*` takes multiple minutes, multiply that by the matrix of configurations we're testing in github workflows, you're looking at a couple hours just to resolve these dependencies. (Yes, it's really that stupid, I've provided links to reported issues in ce6b2dae76708a55803081523c87f3080d70f56c)

I think it's better we pin this for CI only, as a user could have installed a different collection that pins postgresql to X.Y.Z and if we're fine with that (given we set it to `*`), galaxy will say requirements are already met and leave it alone anyway, right?

eb4x: Pinning in CI is not ideal, and especially since it's in this many places, but pinning in galaxy.yml would be the worse of two options.

In any case we should probably specify minimum requirements for all the `*` dependencies though.

molecule/requirements.txt
eb4x: Also, the repo is deprecated/archived, so there won't come any further changes to it, but I could pin it to the hash instead?

We should probably go over to molecule 5, but I'd rather get these refactors in before overhauling the new molecule version.

eb4x:
> Any idea or guesses as to why the `molecule verify` step fails? It spits out `Could not resolve host: zabbix-web-rockylinux9`, but I don't see how it was able to resolve before before the namechange of the containers?

Nevermind, I think I see the error of my ways. It's not the full name as used above.

pyrodie18 (approved):

PR #1219, opened 2024-05-07
Update Debian.yml for Ubuntu arm64 compatibility

sulphur:
Use ansible_os_family to determine Ubuntu and Debian flavour.

##### SUMMARY
On ubuntu system (20.04 at least) `ansible_isb.id` value is **ubuntu** not **debian**. Therefore auto-detection for ubuntu `arm64` doesn't work. Considering that **Debian.yml** is included using `{{ansible_os_familiy}}.yml` we can assume that this additional `and` condition is already covered by that.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
Today if installed on Ubuntu arm64 (20.04 for example) installation of zabbix-agent2 fails as Zabbix default repository does not contain arm64 packages anymore.

### DIFF
--- a/roles/zabbix_agent/tasks/Debian.yml
+++ b/roles/zabbix_agent/tasks/Debian.yml
@@ -29,7 +29,7 @@
 
 - name: "Debian | Repo URL"
   ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


pyrodie18 (changes requested):
Please add a change fragment

pyrodie18:
Taken care of in #1225

PR #979, opened 2023-05-03
2.0.0

BGmot:
##### SUMMARY
2.0.0 release

##### ADDITIONAL INFORMATION
To be merged on May 4, 2023

PR #1220, opened 2024-05-07
Include SELinux tasks for zabbix_web and refactor selinux variable names

miikkajo:
##### SUMMARY

Selinux tasks were not included by zabbix_web role,
seems that include had been dropped by accident (?) on pull request #979 (v2.0.0)

Restored `include_tasks: selinux.yml` to roles/zabbix_web/tasks/RedHat.yml

Refactor `zabbix_selinux` variable to `zabbix_web_selinux` as variable names should be prefixed with role name,
also to use similar pattern with `zabbix_proxy_selinux` and `zabbix_server_selinux` variables.

Refactor selinux related variable names in zabbix_web role to correlate with selinux boolean names.

Refactor `zabbix_apache_can_connect_ldap` to `selinux_allow_httpd_can_connect_ldap` 
as it's not apache specific and not correlating with actual selinux boolean name.

Don't set zabbix_can_network boolean in zabbix_web role as it's not relevant (AFAIK).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_web

### DIFF
--- a/changelogs/fragments/1220-refactor_selinux_tasks_and_variable_names_role_zabbix_web.yml
+++ b/changelogs/fragments/1220-refactor_selinux_tasks_and_variable_names_role_zabbix_web.yml
@@ -1,0 +1,4 @@
+- bugfixes: 
+    - zabbix_web role, Add missing selinux.yml tasks.
+- minor_changes:
+    - zabbix_web role, Refactored zabbix_selinux variable names to correlate with selinux boolean names.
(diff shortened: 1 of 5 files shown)


pyrodie18 (changes requested):
Please add a change fragment.

pyrodie18 (approved):

PR #1226, opened 2024-05-13
Refactor zabbix server selinux

eb4x:
##### SUMMARY
Refactor the selinux bits, tasks are either dropped or combined.

##### ISSUE TYPE
- Refactor Pull Request

##### COMPONENT NAME
zabbix-server role

##### ADDITIONAL INFORMATION
There is one potential breaking change, if users are manually setting selinux booleans on their systems, but not in ansible, they could risk the booleans being flipped back to their original `false` state.

### DIFF
--- a/roles/zabbix_server/tasks/RedHat.yml
+++ b/roles/zabbix_server/tasks/RedHat.yml
@@ -36,5 +36,4 @@
 
 - name: "RedHat | Configure SELinux when enabled"
   ansible.builtin.include_tasks: selinux.yml
-  when:
-    - zabbix_server_selinux | bool
+  when: ansible_facts.selinux.status | default('disabled') == 'enabled'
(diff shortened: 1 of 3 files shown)


roles/zabbix_server/tasks/RedHat.yml
pyrodie18, line 1: My one question....does selinux.status shoe enabled if it's in a permissive state?

eb4x:
```
    "ansible_facts.selinux": {
        "config_mode": "enforcing",
        "mode": "permissive",
        "policyvers": 33,
        "status": "enabled",
        "type": "targeted"
    }
```

eb4x: The `config_mode: enforcing` and `mode: permissive` is probably a discrepancy between what the config says it wants, and the fact that i set `enforcing=0` to the linux cmdline at boot.

eb4x: Yup, after edited /etc/selinux/config
```
    "ansible_facts.selinux": {
        "config_mode": "permissive",
        "mode": "permissive",
        "policyvers": 33,
        "status": "enabled",
        "type": "targeted"
    }
```

PR #1229, opened 2024-05-16
Rework/Cleanup Agent Role

### LINKED ISSUE #1004: FR:  Rework tlspk sections of Agent Role
##### SUMMARY
There is significant duplication in code between agent and agent2 tasks in the agent role.  Would recommend that we look at consolidating those tasks and adjust with logic where required.  This will greatly simplify updates in the future and troubleshooting.  

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
agent role

pyrodie18:
##### SUMMARY
This PR cleans up a lot of duplicative code within the Agent role.  It standardizes on the use of the `zabbix_agent` prefix and deprecates the `zabbix_agent2` prefix to make the code more reusable.  Documentation is cleaned up and the config file is moved to the same format as we have done for the other roles.

This PR closes #1004 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Agent


D3DeFi (approved):
To me this looks ok, but I've been soo out of touch with zabbix and this community that I cannot reliably determine what will be the effects on the users. This looks like 3.0.0 release also forces HTTPAPI for everyone?

I must admit that some of my "more up-to-date" repos are still on 2.1.0 :sweat_smile:

pyrodie18:
Understand @D3DeFi.  Pretty sure I didn't break anything, but not 100% sure either since it was a bit of a mess as I was going through everything.  I haven't seen @BGmot much recently either.  Are you guys both tapped out?  Do we need to look at adding someone?

Sopholos:
this PR broke such config and I can't get it work now...

zabbix_agent_hostname: 
zabbix_agent_hostnameitem: system.run["/etc/zabbix/scripts/Get-HostName.ps1"]

zabbix config now contains BOTH values, and I can't imagine how cleanup zabbix_agent_hostname value, it populated from defaults in case of empty, none, null, unset()...

pyrodie18:
> this PR broke such config and I can't get it work now...
> 
> zabbix_agent_hostname: zabbix_agent_hostnameitem: system.run["/etc/zabbix/scripts/Get-HostName.ps1"]
> 
> zabbix config now contains BOTH values, and I can't imagine how cleanup zabbix_agent_hostname value, it populated from defaults in case of empty, none, null, undef()...

@Sopholos, a few things.  First, I would encourage you not to reply to a closed PR.  The only reason I even saw your comment is because I had replied to the thread earlier and was subscribed to it.  Second, if there is a bug, then I'd ask you to please open an issue and include all of the information that we request because right now I'm not following what the problem is.

PR #1230, opened 2024-05-16
Quality of life improvements for zabbix server role

eb4x:
##### SUMMARY
Quality of life improvements for zabbix server role

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_server role

##### ADDITIONAL INFORMATION
I created a Vagrantfile to just test the most basic, common way I can see this role deployed for mysql and pgsql variants, and these are the remaining fixes I needed to deploy the role.

The most common setup being OS with either postgresql or mysql/mariadb installed, and installing zabbix_server to it with the default localhost connection to the database.

### DIFF
--- a/roles/zabbix_server/tasks/initialize-mysql.yml
+++ b/roles/zabbix_server/tasks/initialize-mysql.yml
@@ -27,6 +27,7 @@
 
 - name: "MySQL Database prep"
   when: zabbix_server_database_creation
+  become: "{{ zabbix_server_dbhost_run_install }}"
   delegate_to: "{{ zabbix_server_real_dbhost | default(zabbix_server_dbhost_run_install | ternary(delegated_dbhost, inventory_hostname)) }}"
   vars:
     delegated_dbhost: "{{ (zabbix_server_dbhost == 'localhost') | ternary(inventory_hostname, zabbix_server_dbhost) }}"
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -37,8 +37,6 @@ jobs:
           - v64
           - v62
           - v60
-        include:
-          - interpreter: python3
         exclude:
           - container: debian12
             version: v62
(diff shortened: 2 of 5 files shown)


galaxy.yml
```diff
@@ -24,4 +24,4 @@ dependencies:
   ansible.posix: "*"
   community.general: "*"
   community.mysql: "*"
-  community.postgresql: "*"
+  community.postgresql: ">=2.0.0"
```
pyrodie18, line 27: Is there a reason for this change?  This would be a breaking change.

roles/zabbix_server/tasks/initialize-mysql.yml
```diff
@@ -27,6 +27,7 @@
 
 - name: "MySQL Database prep"
   when: zabbix_server_database_creation
+  become: "{{ zabbix_server_dbhost_run_install }}"
```
pyrodie18, line 30: Why elevate here?  Its not required.

galaxy.yml
eb4x: Not really, no. It's just as far back as I tested the collection. I went to galaxy to find the earliest version, 0.1.0. And that seems to work. Will undo this pinning.

roles/zabbix_server/tasks/initialize-mysql.yml
eb4x: Actually it is if you're supplying `zabbix_server_mysql_login_unix_socket`, which you might do to avoid creating additional credentials for administering your mysql-server. (Similar to how we `become_user: postgres` to avoid creating credentials for postgresql. Atleast I think that's why we do it that way?) This seems to be the most effortless setup.

I have some work in progress testing the roles on vms with vagrant. With these changes, the only tunings i need to install zabbix-server-mysql are [these https://github.com/eb4x/community.zabbix/blob/vagrant/Vagrantfile#L82-L86](https://github.com/eb4x/community.zabbix/blob/vagrant/Vagrantfile#L82-L86). And the mysql installation is simply installing [mariadb-server](https://github.com/eb4x/community.zabbix/blob/vagrant/extensions/vagrant/database-mysql.yml)

I'm curious to know how others get this installed.

pyrodie18 (approved):
Looks good.  Thanks for the add.

PR #1234, opened 2024-05-24
Add Support to Item Modules for setting Master Items by Name

aplathrop:
##### SUMMARY

This change adds support to the zabbix_item and zabbix_itemprototype modules for setting the master item by name rather by ID

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
zabbix_item
zabbix_itemprototype

##### ADDITIONAL INFORMATION
Currently, if you want to create a dependent item using either the zabbix_item and zabbix_itemprototype module you can do so using the master item's ID, but you can't do it by name:

```
community.zabbix.zabbix_item:
    name: ExampleItem
    host_name: ExampleHost
    params:
        type: dependent_item
        key: ExampleKey
        value_type: numeric_unsigned
        master_itemid: 12345
```

This PR adds a new parameter, `master_item` that you can use to set the master item by name rather than by ID:

```
community.zabbix.zabbix_item:
    name: ExampleItem
    host_name: ExampleHost
    params:
        type: dependent_item
        key: ExampleKey
        value_type: numeric_unsigned
        master_item:
          item_name: ExampleMasterItem
          host_name: ExampleHost
```

Note that this PR preserves the previous functionality, if anyone was setting the master item by ID, that will continue to work

### DIFF
--- a/changelogs/fragments/pr_1234.yml
+++ b/changelogs/fragments/pr_1234.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - zabbix_item - add support for setting master items by name
+  - zabbix_itemprototype - add support for setting master items by name
--- a/plugins/modules/zabbix_item.py
+++ b/plugins/modules/zabbix_item.py
@@ -283,6 +283,36 @@ EXAMPLES = r'''
             value: application
     state: present
 
+- name: create a dependent item
+  # set task level variables as we change ansible_connection plugin here
+  vars:
+    ansible_network_os: community.zabbix.zabbix
+    ansible_connection: httpapi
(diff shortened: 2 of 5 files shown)


pyrodie18 (changes requested):
A few things.

- [ ] Please add a change fragment
- [ ] To ensure that we maintain the functionality we already have, please restore the original tests that you modified, and then just add additional tests that fit your use case so we're checking both with and without a master item.
- [ ] Update the argument specs for both modules
- [ ]  Consider adding an example of using this.

Thanks for the PR

pyrodie18:
Fair point.  Thanks for the addition.

PR #1239, opened 2024-05-28
Install GPG key for Red Hat

jonathanspw:
##### SUMMARY
Install Zabbix repo GPG key for rpm distros.

Presently if you try to perform a fresh agent install with `zabbix_repo_yum_gpgcheck` set to `true` it will fail as nothing installed the necessary GPG key.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix-agent

### DIFF
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -21,6 +21,8 @@ selinux_allow_zabbix_run_sudo: false
 zabbix_repo_deb_gpg_key_url: http://repo.zabbix.com/zabbix-official-repo.key
 zabbix_repo_deb_include_deb_src: true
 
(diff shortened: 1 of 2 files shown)


pyrodie18 (changes requested):
Honestly have no idea how someone hasn't brought this up before or noticed it.  Please add a change fragment and update the documentation.  If you would like to add the other roles in here to it would be appreciated.

eb4x:
> Honestly have no idea how someone hasn't brought this up before or noticed it.

That would probably be because the default of  [`zabbix_repo_yum_gpgcheck: 0`](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L31)

I believe there are also different gpg keys signing for el8 and el9, so this is probably not enough.

pyrodie18:
Included in #1272

PR #1240, opened 2024-05-29
zabbix_agent - Use zabbix_agent_distribution_release to backport packages

kivilahtio:
##### SUMMARY
Configuration 'zabbix_agent_distribution_release' is already defined in the role. Actually make it usable to overload the system's ansible_distribution_release. This is useful to for example install Zabbix Agent2 on Raspbian 12 Bookworm using the older packages, as there is not release for the specific target distribution and version.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Zabbix Agent

### DIFF
--- a/roles/zabbix_agent/tasks/Debian.yml
+++ b/roles/zabbix_agent/tasks/Debian.yml
@@ -61,7 +61,7 @@
       Types: deb{{ ' deb-src' if zabbix_repo_deb_include_deb_src }}
       Enabled: yes
       URIs: {{ zabbix_repo_deb_url }}
-      Suites: {{ ansible_distribution_release }}
+      Suites: {{ zabbix_agent_distribution_release }}
       Components: {{ zabbix_repo_deb_component }}
       Architectures: {{ 'amd64' if ansible_machine != 'aarch64' else 'arm64'}}
       Signed-By: {{ zabbix_gpg_key }}
--- a/changelogs/fragments/zabbix_agent_distribution_release.yml
+++ b/changelogs/fragments/zabbix_agent_distribution_release.yml
@@ -1,0 +1,2 @@
+minor_changes:
+ - zabbix_agent_distribution_release - Overload ansible_distribution_release on Debian when choosing packages to install


kivilahtio:
Unrelated to this:
I delivered a PR to address an architecture issue I encountered with the Raspbian support:

https://github.com/ansible-collections/community.zabbix/pull/1060

It is closed, but maybe it can be reopened to continue the discussion as there is an update to the previously fixed issue?

pyrodie18:
I don't see where `zabbix_agent_distribution_release` is defined anywhere.

kivilahtio:
> I don't see where `zabbix_agent_distribution_release` is defined anywhere.

commit f008e178bf0fa31fb74cbcc96054a8d8feee3094
Author: Werner Dijkerman <werner.dijkerman@iwelcome.com>
Date:   Tue Jun 9 19:02:50 2020 +0200

    As it provides error while using a dash (-) in the role name, we will use an underscore (_)

file: roles/zabbix_agent/defaults/main.yml

roles/zabbix_agent/tasks/Debian.yml
```diff
@@ -61,7 +61,7 @@
       Types: deb{{ ' deb-src' if zabbix_repo_deb_include_deb_src }}
       Enabled: yes
       URIs: {{ zabbix_repo_deb_url }}
-      Suites: {{ ansible_distribution_release }}
+      Suites: {{ zabbix_agent_distribution_release }}
```
eb4x, line 64: There's no default provided for this, it's gonna break for everyone that hasn't defined this variable. It was cleaned out by c2a32d5ea130eadbf4c4a0b6160757cd9a7e8ac9

eb4x:
> Unrelated to this: I delivered a PR to address an architecture issue I encountered with the Raspbian support:
> #1060
> It is closed, but maybe it can be reopened to continue the discussion as there is an update to the previously fixed issue?

I think you'll find that #1225 fixes what #1060 wanted to do. But you're probably right that `Suites:` should be based on the lsb.id.

That's not what's going down in this PR though, you'll have to update to reflect the newest changes.

eb4x:
> But you're probably right that `Suites:` should be based on the lsb.id.

Actually, it should probably just say `zabbix` here? Looking at https://repo.zabbix.com/zabbix/6.4/raspbian/dists/bullseye/Release `Suite: zabbix`, or maybe it doesn't really matter?

BGmot:
@kivilahtio are you still wanting to pursue this?

kivilahtio:
> @kivilahtio are you still wanting to pursue this?

Not at them moment.
Hopefully #1225 will fix it

PR #1258, opened 2024-05-31
Add tag support for zabbix_host_events_info module

aplathrop:
##### SUMMARY

Adds support to the zabbix_host_events_info module to  filter events by tag. It also adds the tags returned by last_event

As there were no unit tests for the zabbix_host_events_info module, this also adds unit tests for it, both with and without tags

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- zabbix_host_events_info

##### ADDITIONAL INFORMATION

Being able to filter events by tag is supported by the Zabbix API, but not the current zabbix_host_events_info module, so I've added support for it. An example of this can be seen below:
```
- community.zabbix.zabbix_host_events_info:
    host_identifier: ExampleHost
    host_id_type: hostname
    trigger_severity: warning
    tags:
      - tag: ExampleTag
        value: ExampleValue
        operator: equal
```

I believe it would also be useful to do the inverse, get a list of events for a host and then be able to do something based on the tags the events return:

```
- community.zabbix.zabbix_host_events_info:
    host_identifier: ExampleHost
    host_id_type: hostname
    trigger_severity: warning

- ansible.builtin.debug:
    msg: "{{ zabbix_host_events_tags.triggers_problem[0].last_event.tags }}"
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,15 @@ community.zabbix Release Notes
 
 .. contents:: Topics
 
+v2.5.1
+======
+
+Bugfixes
+-------------
+
+- zabbix_agent - Fix reading existing psk
+- zabbix_agent - Fix role when zabbix_agent_listenip is undefined
+
 v2.5.0
 ======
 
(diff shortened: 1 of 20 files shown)


pyrodie18 (approved):
Nice work!  Thank you for the addition

Issue #1112, opened 2023-11-03
unable to connect to socket error while running task for multiple server


solick:
##### SUMMARY
Running a task against one server is working, running it against a server group cause "unable to connect to socket" error

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.zabbix.zabbix_hostmacro:

##### ANSIBLE VERSION

```
ansible [core 2.15.3]
  config file = /Users/xxx/Development/Ansible/lmc-ansible/ansible.cfg
  configured module search path = ['/Users/xxx/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/8.3.0/libexec/lib/python3.11/site-packages/ansible
  ansible collection location = /Users/xxx/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.11.5 (main, Aug 24 2023, 15:18:16) [Clang 14.0.3 (clang-1403.0.22.14.1)] (/usr/local/Cellar/ansible/8.3.0/libexec/bin/python)
  jinja version = 3.1.2
  libyaml = True

```

##### CONFIGURATION

```
CONFIG_FILE() = /Users/xxx/Development/Ansible/lmc-ansible/ansible.cfg
DEFAULT_HOST_LIST(/Users/xxx/Development/Ansible/lmc-ansible/ansible.cfg) = ['/Users/xxx/Development/Ansible/lmc-ansible/hosts']
DEFAULT_REMOTE_USER(/Users/xxx/Development/Ansible/lmc-ansible/ansible.cfg) = root
DEFAULT_VAULT_PASSWORD_FILE(/Users/xxx/Development/Ansible/lmc-ansible/ansible.cfg) = /Users/xxx/Development/Ansible/lmc-ansible-vault-password
EDITOR(env: EDITOR) = nano

```

##### OS / ENVIRONMENT / Zabbix Version
Zabbix 6.0
MacOS 17.1

##### STEPS TO REPRODUCE

Set macros via zabbix_hostmacro module.
Use the tasks for a single server is working, use the task for a group of servers is not working.

```
---

- name: add maccro trigger_threshold_processes to host
  vars:
    gather_facts: false
    ansible_network_os: community.zabbix.zabbix
    ansible_connection: httpapi
    ansible_httpapi_port: 443
    ansible_httpapi_use_ssl: true  # Set to true for HTTPS
[...]

PR #1259, opened 2024-05-31
Fix problems with Zabbix Agent 2

jeroenvermeulen:
##### SUMMARY
Fix problems with Zabbix Agent 2:
- zabbix_agent2_listenport not defined
- zabbix_agent_tlsaccept not defined
- zabbix_agent_tlsconnect not defined
- API calls were connecting to monitored host instead of zabbix_api_server_host
- timeouts because API was called for many hosts at once

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_agent

##### ADDITIONAL INFORMATION
I am using Zabbix 6.4.15 on a number of Ubuntu 22.04 and 24.04 hosts running Zabbix Agent 2 version 6.4.15.


roles/zabbix_agent/tasks/main.yml
```diff
@@ -118,6 +118,7 @@
 - name: "Run the API calls to Zabbix Server"
   vars:
     gather_facts: false
+    ansible_host: "{{ zabbix_api_server_host }}"
```
pyrodie18, line 121: So I have two points here.

1.  Please reference the conversation in https://github.com/ansible-collections/community.zabbix/pull/920 about using delegate_to vs ansible_host.
2. @D3DeFi  and @BGmot I have a question.  I'll admit I still don't totally understand how the HTTPAPI works.  My question is, why do we have to set vars (i.e. `ansible_httpapi_use_ssl` in ./tasks/main line 118 but we don't have to do the same thing over in the api.yml calls?

pyrodie18 (changes requested):
Please see the comments.  Also please update the documentation for the throttle and add a change fragment.

roles/zabbix_agent/tasks/api.yml
```diff
@@ -1,5 +1,6 @@
 ---
 - name: "API | Create host groups"
+  throttle: 1
```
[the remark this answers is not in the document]
jeroenvermeulen: OK, made it configurabel using `zabbix_api_throttle`

roles/zabbix_agent/defaults/main.yml
```diff
@@ -90,6 +90,9 @@ zabbix_agent_interfaces:
     dns: "{{ ansible_fqdn }}"
     port: "{{ (zabbix_agent2 == True) | ternary(zabbix_agent2_listenport, zabbix_agent_listenport) }}"
 
+zabbix_agent_listenport: 10050
+zabbix_agent2_listenport: 10050
```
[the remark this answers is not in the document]
jeroenvermeulen: Reolved this by removing the `tenary()` construction in ./defaults/main.yml L91
and ./tasks/main.yml line 46.
These `tenary()` constructions don't work when `zabbix_agent2_listenport` is not defined.
We can just use `default()` twice to do multi-stage fallback.

Now it just uses the `zabbix_agent_listenport` defined in ./tasks/main.yml line 46

roles/zabbix_agent/tasks/main.yml
jeroenvermeulen: I have tried several solutions and this was the only thing that fixed the problem of the API calls going to the monitored server instead of `zabbix_api_server_host`

BGmot: @pyrodie18 you need these variables set for any module from this collection (API call). Ansible is flexible enough to set variables at different places (I personally set them up in my inventory for Zabbix server host). Here we set them up at `ansible.builtin.include_tasks` task because they will be used for every task in api.yml (just to avoid duplication).
We don't need this line 121 as every task in api.yml has `delegate_to`
https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/tasks/api.yml#L41
https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/tasks/api.yml#L57

roles/zabbix_agent/tasks/api.yml
```diff
@@ -1,5 +1,6 @@
 ---
 - name: "API | Create host groups"
+  throttle: "{{ zabbix_api_throttle | int }}"
```
BGmot, line 3:
```suggestion
```
We don't need throttle for everyone. I regularly use this role on tens of hosts and never experience any issues except when your Ansible controller is running on MacOS (see https://github.com/ansible-collections/community.zabbix/issues/1112)
If you want you can play with `forks` and `serial` on playbook level.

roles/zabbix_agent/tasks/main.yml
jeroenvermeulen: I see the `delegate_to` but still in my setup it is not working without the `ansible_host` line.

roles/zabbix_agent/tasks/api.yml
```diff
@@ -12,6 +13,7 @@
     - api
 
 - name: "API | Create a new host or update an existing host's info"
+  throttle: "{{ zabbix_api_throttle | int }}"
```
BGmot, line 16:
```suggestion
```
See my comment above.

roles/zabbix_agent/tasks/api.yml
```diff
@@ -44,6 +46,7 @@
     - api
 
 - name: "API | Updating host configuration with macros"
+  throttle: "{{ zabbix_api_throttle | int }}"
```
BGmot, line 49:
```suggestion
```
See my comment above.

roles/zabbix_agent/tasks/main.yml
```diff
@@ -43,7 +43,7 @@
     zabbix_agent_heartbeatfrequency: "{{ zabbix_agent_heartbeatfrequency is defined | ternary(zabbix_agent_heartbeatfrequency, zabbix_agent2_heartbeatfrequency) | default(60) }}"
     zabbix_agent_hostname: "{{ zabbix_agent_hostname is defined | ternary(zabbix_agent_hostname, zabbix_agent2_hostname) | default(inventory_hostname) }}"
     zabbix_agent_include_mode: "{{ zabbix_agent_include_mode is defined | ternary(zabbix_agent_include_mode, zabbix_agent2_include_mode) | default('0750') }}"
-    zabbix_agent_listenport: "{{ zabbix_agent_listenport is defined | ternary(zabbix_agent_listenport, zabbix_agent2_listenport) | default(10050) }}"
+    zabbix_agent_listenport: "{{ zabbix_agent_listenport | default(zabbix_agent2_listenport) | default(10050) }}"
```
BGmot, line 46: Sorry, I don't get this change. Why is it better than `ternary`?

BGmot (changes requested):

roles/zabbix_agent/tasks/main.yml
BGmot: Please troubleshoot your set up. We can help -) But it is definitely working.

roles/zabbix_agent/tasks/main.yml
jeroenvermeulen: Because the tenary does not work when `zabbix_agent2_listenport` is not defined.
I have tested it.

roles/zabbix_agent/tasks/api.yml
jeroenvermeulen: I am indeed often running Ansible on mij Mac.
Thanks for the hint, I will try to find another solution.

roles/zabbix_agent/tasks/main.yml
pyrodie18: Its works fine in all of the lines above it, not sure why it would be different on this one.  Are you sure its not a matter of `zabbix_agent2_listenport` being null instead of not defined?

roles/zabbix_agent/tasks/main.yml
pyrodie18: > @pyrodie18 you need these variables set for any module from this collection (API call). Ansible is flexible enough to set variables at different places (I personally set them up in my inventory for Zabbix server host). Here we set them up at `ansible.builtin.include_tasks` task because they will be used for every task in api.yml (just to avoid duplication). We don't need this line 121 as every task in api.yml has `delegate_to` https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/tasks/api.yml#L41 https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/tasks/api.yml#L57

Oh duh...that's what I get for looking at code at midnight.

roles/zabbix_agent/tasks/main.yml
jeroenvermeulen: OK you are right. We don't need this change because it is already fixed by this change https://github.com/ansible-collections/community.zabbix/pull/1259/files#diff-d722eb1d2f161071c11093c3875382316f4cc4b6bf5d37be35bbb3b5ac2e283aR91

jeroenvermeulen:
I have now added 10 retries to the API calls, to workaround the connection problems when running Ansible on macOS.
This change does not hurt anyone.

roles/zabbix_agent/tasks/main.yml
jeroenvermeulen: @BGmot Even when I hard-code the hostname of my Zabbix server 3x in the `delegate_to` in `zabbix_agent/tasks/api.yml` the HttpApi still connects to the monitored node instead of the Zabbix server. When I use `ansible_host` like in this PR, it works.

BGmot: Ok, let me test more. I'll get back to you on this.

jeroenvermeulen: I noticed when delegating in the `api.xml` the Ansible output contains:
```
fatal: [someserver.mydomain.io -> zabbix.mydomain.io(None)]: ....
```
while another delegation when it works does not contain the `(None)`:
```
ok: [someserver.mydomain.io -> delegated-to.mydomain.io]
```

BGmot: On a template of an issue creation we have environment details to be provided. It's not for fun. It's for us to waste less time to re-produce your problem and try to resolve it. Please provide Python version, Ansible version, and OS you are running Ansible controller on.

jeroenvermeulen: Found https://github.com/ansible/ansible/issues/27351
Indeed it works when I don't set `ansible_ssh_host` for my Zabbix host.
I still think you should consider switching to overriding `ansible_host` instead using `delegate_to` because you want to use the HTTP API to directly connect to the `zabbix_api_server_host` and you don't need any Ansible connection magic to be involved.

```
Python 3.12.3
ansible [core 2.16.7]
  python version = 3.12.3 (main, Apr  9 2024, 08:09:14) [Clang 15.0.0 (clang-1500.3.9.4)] (/opt/homebrew/Cellar/ansible/9.6.0/libexec/bin/python)
  jinja version = 3.1.4
Darwin MyHost.local 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:14:38 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6020 arm64
macOS Sonoma 14.5
```

BGmot: The ansible issue you found claimed to be fixed long time ago.
I don't know I am comfortable with the code we have now and nobody complained about this so far. Looks like the issue is specific to your use case. I am not very comfortable with proposed changed. @pyrodie18, @D3DeFi what do you think?

pyrodie18: > The ansible issue you found claimed to be fixed long time ago. I don't know I am comfortable with the code we have now and nobody complained about this so far. Looks like the issue is specific to your use case. I am not very comfortable with proposed changed. @pyrodie18, @D3DeFi what do you think?

Agree.  We made this change a long time ago and no one else has complained about it.  Given that while I appreciate the effort with the PR, I'm inclined to not pull it in.

jeroenvermeulen: OK I will remove this change. 
I can work around it by **not** setting `ansible_ssh_host` in the host_vars of my Zabbix server.
If you would like to repeat the problem: Just add this setting with the IP address of the Zabbix server.

roles/zabbix_agent/defaults/main.yml
```diff
@@ -88,7 +88,7 @@ zabbix_agent_interfaces:
     useip: "{{ zabbix_useuip }}"
     ip: "{{ zabbix_agent_ip }}"
     dns: "{{ ansible_fqdn }}"
-    port: "{{ (zabbix_agent2 == True) | ternary(zabbix_agent2_listenport, zabbix_agent_listenport) }}"
+    port: "{{ zabbix_agent_listenport }}"
```
jeroenvermeulen, line 91: This change fixes the error:
```
fatal: [host.domain.io -> zabbix.domain.io]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: [{'type': 1, 'main': 1, 'useip': '{{ zabbix_useuip }}', 'ip': '{{ zabbix_agent_ip }}', 'dns': '{{ ansible_fqdn }}', 'port': '{{ (zabbix_agent2 == True) | ternary(zabbix_agent2_listenport, zabbix_agent_listenport) }}'}]: 'zabbix_agent2_listenport' is undefined. 'zabbix_agent2_listenport' is undefined. [{'type': 1, 'main': 1, 'useip': '{{ zabbix_useuip }}', 'ip': '{{ zabbix_agent_ip }}', 'dns': '{{ ansible_fqdn }}', 'port': '{{ (zabbix_agent2 == True) | ternary(zabbix_agent2_listenport, zabbix_agent_listenport) }}'}]: 'zabbix_agent2_listenport' is undefined. 'zabbix_agent2_listenport' is undefined\n\nThe error appears to be in '/path/to/collections/ansible_collections/community/zabbix/roles/zabbix_agent/tasks/api.yml': line 15, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"API | Create a new host or update an existing host's info\"\n  ^ here\n"}
```
During task:
```
TASK [community.zabbix.zabbix_agent : API | Create a new host or update an existing host's info]
```

roles/zabbix_agent/tasks/api.yml
```diff
@@ -27,8 +28,8 @@
     tls_psk_identity: "{{ zabbix_agent_tlspskidentity | default(omit) }}"
     tls_issuer: "{{ zabbix_agent_tlsservercertissuer | default(omit) }}"
     tls_subject: "{{ zabbix_agent_tls_subject | default(omit) }}"
-    tls_accept: "{{ zabbix_agent_tls_config[zabbix_agent_tlsaccept if zabbix_agent_tlsaccept else 'unencrypted'] }}"
-    tls_connect: "{{ zabbix_agent_tls_config[zabbix_agent_tlsconnect if zabbix_agent_tlsconnect else 'unencrypted'] }}"
+    tls_accept: "{{ zabbix_agent_tls_config[zabbix_agent_tlsaccept | default('unencrypted')] }}"
```
jeroenvermeulen, line 31: This change fixes the problem:
```
fatal: [host.domain.io -> zabbix.domain.io]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'zabbix_agent_tlsaccept' is undefined. 'zabbix_agent_tlsaccept' is undefined\n\nThe error appears to be in '/path/to/collections/ansible_collections/community/zabbix/roles/zabbix_agent/tasks/api.yml': line 15, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"API | Create a new host or update an existing host's info\"\n  ^ here\n"}
```
during task
```
TASK [community.zabbix.zabbix_agent : API | Create a new host or update an existing host's info]
```
So the problem is `zabbix_agent_tlsaccept` is undefined and `if zabbix_agent_tlsaccept` is not working.

roles/zabbix_agent/tasks/api.yml
```diff
@@ -27,8 +28,8 @@
...
+    tls_connect: "{{ zabbix_agent_tls_config[zabbix_agent_tlsconnect | default('unencrypted')] }}"
```
jeroenvermeulen, line 32: Same story as the line above.
The problem is `zabbix_agent_tlsconnect` is not defined.

roles/zabbix_agent/tasks/api.yml
```diff
@@ -7,6 +7,7 @@
     - zabbix_api_create_hostgroup | bool
   register: zabbix_api_hostgroup_created
   until: zabbix_api_hostgroup_created is succeeded
+  retries: 10
```
jeroenvermeulen, line 10: I have added 10 retries to the API calls, to workaround the connection problems when running Ansible on macOS.
This change does not hurt anyone. Indeed this is a better solution than a `throttle`.

roles/zabbix_agent/defaults/main.yml
pyrodie18: yep apparently I missed this one when I was updating the role.

pyrodie18 (changes requested):
OK, I'm good with these changes.  If you want to toss in a change fragment we should be good.  Thanks for the contribution.

BGmot:
> Change fragment, is that a line in CHANGELOG.rst?

no, it's a file here https://github.com/ansible-collections/community.zabbix/tree/main/changelogs/fragments

jeroenvermeulen:
Done.

BGmot (approved):

pyrodie18 (approved):

PR #1261, opened 2024-06-01
Remove support for old versions of Python and Ansible

### LINKED ISSUE #1254: Remove Support for Ansible < 2.15
As part of 3.0.0 cleanup, remove support for Ansible < 2.15 which are all EOL.

pyrodie18:
##### SUMMARY
Resolves #1254


.github/workflows/plugins-integration.yml
```diff
@@ -25,14 +25,21 @@ jobs:
...
-          - stable-2.13
-          - stable-2.14
           - stable-2.15
           - stable-2.16
           - devel
         python:
+          - '3.9'
           - '3.10'
+          - '3.11'
+          - '3.12'
```
BGmot, line 35: This version of Python is not used anywhere in roles tests. Why?

pyrodie18: My thought was that the role aren't using raw python, but the modules are.

BGmot: For roles this is the version of Python that is running Ansible code on Ansible controller.
What's your reasoning for adding Python versions 3.9, 3.11 and 3.12 to modules tests (plugins-integration)? why not just replace 3.10 with the latest 3.12? Do we claim support of particular versions of Python anywhere?

pyrodie18: I think we talk about supporting specific versions of Ansible which in turn ties back to several versions of Python.  I'm not dead set on the change.

BGmot (approved):

Issue #1237, opened 2024-05-28
Zabbix agent 2 persistent buffer is broken


badfiles:
The enable parameter is missing in agent template

https://github.com/ansible-collections/community.zabbix/commit/c2a32d5ea130eadbf4c4a0b6160757cd9a7e8ac9#r142464773

badfiles:
now the zabbix_agent_persistentbuffer is missing in the rest of the code

pyrodie18:
@badfiles  take a look at https://github.com/ansible-collections/community.zabbix/pull/1278 and tell me if I actually fixed this finally

badfiles:
yes, now it works

PR #1278, opened 2024-06-08
One more try at fixing zabbix_agent_persistentbuffer

pyrodie18:
##### SUMMARY
Related to #1237

### DIFF
--- a/roles/zabbix_agent/tasks/main.yml
+++ b/roles/zabbix_agent/tasks/main.yml
@@ -46,7 +46,7 @@
     zabbix_agent_listenport: "{{ zabbix_agent_listenport is defined | ternary(zabbix_agent_listenport, zabbix_agent2_listenport) | default(10050) }}"
     zabbix_agent_logfilesize: "{{ zabbix_agent_logfilesize is defined | ternary(zabbix_agent_logfilesize, zabbix_agent2_logfilesize) | default(100) }}"
     zabbix_agent_logtype: "{{ zabbix_agent_logtype is defined | ternary(zabbix_agent_logtype, zabbix_agent2_logtype) | default('file') }}"
-    zabbix_agent_persistentbuffer: "{{ zabbix_agent_persistentbuffer is defined | ternary(zabbix_agent_persistentbuffer, zabbix_agent_persistentbuffer) | default(0) }}"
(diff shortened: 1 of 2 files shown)


roles/zabbix_agent/tasks/main.yml
```diff
@@ -46,7 +46,7 @@
     zabbix_agent_listenport: "{{ zabbix_agent_listenport is defined | ternary(zabbix_agent_listenport, zabbix_agent2_listenport) | default(10050) }}"
     zabbix_agent_logfilesize: "{{ zabbix_agent_logfilesize is defined | ternary(zabbix_agent_logfilesize, zabbix_agent2_logfilesize) | default(100) }}"
     zabbix_agent_logtype: "{{ zabbix_agent_logtype is defined | ternary(zabbix_agent_logtype, zabbix_agent2_logtype) | default('file') }}"
-    zabbix_agent_persistentbuffer: "{{ zabbix_agent_persistentbuffer is defined | ternary(zabbix_agent_persistentbuffer, zabbix_agent_persistentbuffer) | default(0) }}"
+    zabbix_agent_enablepersistentbuffer: "{{ zabbix_agent_enablepersistentbuffer is defined | ternary(zabbix_agent_enablepersistentbuffer, zabbix_agent_enablepersistentbuffer) | default(0) }}"
```
eb4x, line 49: I think the most "ansible-ism" way to do this would be to put

`zabbix_agent_enablepersistentbuffer: false` in `defaults/main.yml` and then do `"{{ zabbix_agent_enablepersistentbuffer | ternary('1', '0') }}`

pyrodie18: Ya I get what you're saying.  Right now the stupid thing keeps failing on github and I can't figure out why because when I run the same exact molecule on my box it runs fine.
't

pyrodie18: OK, ya, I like it, although going make the change across the board for this since there are a number of other variables that are very similar.

eb4x: It seems the zabbix-agent fails to start, could be helpful to see the output. Maybe temporarily put the "start service" in a block, something like

```
- block:
    - service:
        name: zabbix-agent
  rescue:
    - slurp:
        path: /var/log/zabbix/zabbix_agent.log
      register: _res
    - debug:
          var: _res.content | b64decode
    - fail:
```
Just to see what zabbix-agent service is complaining on.

pyrodie18: Thank you.  I was trying to figure out a way to get that.

PR #1279, opened 2024-06-08
Went back to upstream php

pyrodie18:
##### SUMMARY
This change removes the earlier addition of Remi PHP


pyrodie18:
I see that @eb4x  is already making this change so will just drop it and pull his in when its ready

PR #1281, opened 2024-06-08
fix: test matrix for zabbix_proxy

### LINKED ISSUE #1280: Molecule tests for zabbix_proxy only run on rockylinux
##### SUMMARY

#1196 introduced a bug in the testing pipeline which causes the `zabbix_proxy` role to be tested only on rockylinux. 
If you look at the detailed action logs, you'll see that `MY_MOLECULE_CONTAINER`, `MY_MOLECULE_IMAGE`  and `MY_MOLECULE_OS_FAMILY` are always empty.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

Github actions for `zabbix_proxy`

jon4hz:
##### SUMMARY

Closes #1280 

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/.github/workflows/proxy.yml
+++ b/.github/workflows/proxy.yml
@@ -26,13 +26,27 @@ jobs:
       fail-fast: false
       matrix:
         container:
-          - rockylinux9
-          - rockylinux8
-          - ubuntu2404
-          - ubuntu2204
-          - ubuntu2004
-          - debian12
-          - debian11
+          - name: rockylinux9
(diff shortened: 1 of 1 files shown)


pyrodie18:
@jon4hz  not sure why you're saying it only runs on Rocky.  It does Rocky 8 & 9, Debian 10 and 11, and Ubuntu 20, 22, 24

pyrodie18:
OK, I see what you're talking about.  We actually need to make some different changes to put this back inline with how the rest of the testing is being done.

PR #1290, opened 2024-06-12
Fix zabbix_agent role on Windows

emrocha:
##### SUMMARY
Role zabbix_agent is not working on Windows.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Role zabbix_agent

##### ADDITIONAL INFORMATION
- Single quotes to avoid double backslashes.
- Remove default variable zabbix_agent_win_include because zabbix-agent2 may be used.
- Fix agent config template name to agent.conf.j2
- Default values for zabbix_win_package and zabbix_win_download_link
- Default value for zabbix_agent_controlsocket
- zabbix_agent_tlspskidentity_check and zabbix_agent_tlspskcheck must be defined in a separately task. There were two tasks that could define these variables. If the second task is skipped, as it has register keyword, the variable is reset.
- Define a value for _win_include

### DIFF
--- a/roles/zabbix_agent/tasks/Windows.yml
+++ b/roles/zabbix_agent/tasks/Windows.yml
@@ -4,6 +4,7 @@
     zabbix_agent_win_download_link: "{{ zabbix_agent_win_download_link is defined | ternary(zabbix_agent_win_download_link, zabbix_agent2_win_download_link) | default(_win_download_link) }}"
     zabbix_agent_win_logfile: "{{ zabbix_agent_win_logfile is defined | ternary(zabbix_agent_win_logfile, zabbix_agent2_win_logfile) | default(_win_logfile) }}"
     zabbix_agent_win_package: "{{ zabbix_agent_win_package is defined | ternary(zabbix_agent_win_package, zabbix_agent2_win_package) | default(_win_package) }}"
+    zabbix_agent_win_include: "{{ zabbix_agent_win_include is defined | ternary(zabbix_agent_win_include, zabbix_agent2_win_include) | default(_win_include) }}"
 
 - name: "Windows | Set default architecture"
   ansible.builtin.set_fact:
@@ -191,8 +192,8 @@
 
 - name: "Windows | Set installation settings (agent 2)"
   ansible.builtin.set_fact:
-    zabbix_win_package: "{{ zabbix2_win_package }}"
-    zabbix_win_download_link: "{{ zabbix2_win_download_link }}"
+    zabbix_win_package: "{{ zabbix_win_package | default(_win_package) }}"
+    zabbix_win_download_link: "{{ zabbix_win_download_link | default(_win_download_link) }}"
     zabbix_win_exe_path: "{{ zabbix2_win_exe_path }}"
     zabbix_win_config_name: "{{ zabbix2_win_config_name }}"
     zabbix_win_svc_name: "{{ zabbix2_win_svc_name }}"
--- a/roles/zabbix_agent/tasks/Windows_conf.yml
+++ b/roles/zabbix_agent/tasks/Windows_conf.yml
@@ -2,7 +2,14 @@
 - name: "Set Log File Info"
   ansible.builtin.set_fact:
     zabbix_agent_logfile: "{{ zabbix_agent_win_logfile is defined | ternary(zabbix_agent_win_logfile, zabbix_agent2_win_logfile) | default(_win_logfile) }}"
-  when: zabbix_agent_logfile is undefined
+
+- name: "Set Include Path Info"
+  ansible.builtin.set_fact:
+    zabbix_agent_include: "{{ zabbix_agent_win_include is defined | ternary(zabbix_agent_win_include, zabbix_agent2_win_include) | default(_win_include) }}"
+
+- name: "Set Control Socket"
+  ansible.builtin.set_fact:
+    zabbix_agent_controlsocket: "\\\\.\\pipe\\agent.sock"
 
 - name: "Set default ip address for zabbix_agent_ip"
   ansible.builtin.set_fact:
--- a/roles/zabbix_agent/tasks/psk_identity.yml
+++ b/roles/zabbix_agent/tasks/psk_identity.yml
@@ -3,17 +3,21 @@
     - name: AutoPSK | Check for existing TLS PSK identity | Windows
       ansible.windows.win_stat:
         path: "{{ zabbix_agent_tlspskidentity_file }}"
-      register: zabbix_agent_tlspskidentity_check
+      register: zabbix_agent_tlspskidentity_check_windows
       when:
         - ansible_os_family == "Windows"
 
     - name: AutoPSK | Check for existing TLS PSK identity | Linux
       ansible.builtin.stat:
         path: "{{ zabbix_agent_tlspskidentity_file }}"
-      register: zabbix_agent_tlspskidentity_check
+      register: zabbix_agent_tlspskidentity_check_linux
       become: true
       when:
         - ansible_os_family != "Windows"
+
+    - name: Set zabbix_agent_tlspskidentity_check
+      ansible.builtin.set_fact:
+        zabbix_agent_tlspskidentity_check: "{{ zabbix_agent_tlspskidentity_check_windows if ansible_os_family == 'Windows' else zabbix_agent_tlspskidentity_check_linux }}"
   tags:
     - config
 
--- a/roles/zabbix_agent/tasks/psk_secret.yml
+++ b/roles/zabbix_agent/tasks/psk_secret.yml
@@ -12,17 +12,21 @@
     - name: AutoPSK | Check for existing TLS PSK file | Windows
       ansible.windows.win_stat:
         path: "{{ zabbix_agent_tlspskfile }}"
-      register: zabbix_agent_tlspskcheck
+      register: zabbix_agent_tlspskcheck_windows
       when:
         - ansible_os_family == "Windows"
 
     - name: AutoPSK | Check for existing TLS PSK file | Linux
       ansible.builtin.stat:
         path: "{{ zabbix_agent_tlspskfile }}"
-      register: zabbix_agent_tlspskcheck
+      register: zabbix_agent_tlspskcheck_linux
       become: true
       when:
         - ansible_os_family != "Windows"
+
+    - name: Set zabbix_agent_tlspskcheck
+      ansible.builtin.set_fact:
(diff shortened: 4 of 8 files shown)


roles/zabbix_agent/tasks/psk_identity.yml
```diff
@@ -3,17 +3,21 @@
     - name: AutoPSK | Check for existing TLS PSK identity | Windows
       ansible.windows.win_stat:
         path: "{{ zabbix_agent_tlspskidentity_file }}"
-      register: zabbix_agent_tlspskidentity_check
+      register: zabbix_agent_tlspskidentity_check_windows
```
pyrodie18, line 6: I don't see how `zabbix_agent_tlspskidentity_check` could be set twice since the two choices for the `ansible_os_family` variable are mutually exclusive.

roles/zabbix_agent/tasks/psk_secret.yml
```diff
@@ -12,17 +12,21 @@
     - name: AutoPSK | Check for existing TLS PSK file | Windows
       ansible.windows.win_stat:
         path: "{{ zabbix_agent_tlspskfile }}"
-      register: zabbix_agent_tlspskcheck
+      register: zabbix_agent_tlspskcheck_windows
```
pyrodie18, line 15: I don't see how `zabbix_agent_tlspskcheck` could be set twice since the two choices for the `ansible_os_family` variable are mutually exclusive.

roles/zabbix_agent/vars/agent2_vars.yml
```diff
@@ -3,7 +3,8 @@ _logfile: /var/log/zabbix/zabbix_agent2.log
 _include: /etc/zabbix/zabbix_agent2.d
 _tls_subject: "{{ zabbix_agent_tlsservercertsubject | default(omit) }}" # FIXME this is not correct and should be removed with 2.0.0, here only to prevent regression
 _win_package: zabbix_agent2-{{ zabbix_version_long }}-windows-amd64-openssl-static.zip
-_win_download_link: "{{ zabbix_win_download_url }}/{{ zabbix_version_long | regex_search('^\\d+\\.\\d+') }}/{{ zabbix_version_long }}/{{ zabbix2_win_package }}"
-_win_logfile: "{{ zabbix_win_install_dir }}\\zabbix_agent2.log"
+_win_download_link: "{{ zabbix_win_download_url }}/{{ zabbix_version_long | regex_search('^\\d+\\.\\d+') }}/{{ zabbix_version_long }}/{{ zabbix2_win_package | default('_win_package') }}"
```
pyrodie18, line 6: See the latest changes.  Support for `zabbix2_win_package` has been removed with the release of 3.0.0

roles/zabbix_agent/tasks/Windows_conf.yml
```diff
@@ -2,7 +2,14 @@
 - name: "Set Log File Info"
   ansible.builtin.set_fact:
     zabbix_agent_logfile: "{{ zabbix_agent_win_logfile is defined | ternary(zabbix_agent_win_logfile, zabbix_agent2_win_logfile) | default(_win_logfile) }}"
-  when: zabbix_agent_logfile is undefined
+
+- name: "Set Include Path Info"
+  ansible.builtin.set_fact:
+    zabbix_agent_include: "{{ zabbix_agent_win_include is defined | ternary(zabbix_agent_win_include, zabbix_agent2_win_include) | default(_win_include) }}"
```
pyrodie18, line 8: See the latest changes.  Support for `zabbix_agent2_win_include` has been removed with the release of 3.0.0

roles/zabbix_agent/tasks/Windows.yml
```diff
@@ -191,8 +192,8 @@
 
 - name: "Windows | Set installation settings (agent 2)"
   ansible.builtin.set_fact:
-    zabbix_win_package: "{{ zabbix2_win_package }}"
-    zabbix_win_download_link: "{{ zabbix2_win_download_link }}"
+    zabbix_win_package: "{{ zabbix2_win_package | default('_win_package') }}"
```
pyrodie18, line 195: See the latest changes. Support for `zabbix2_win_package` has been removed with the release of 3.0.0

roles/zabbix_agent/tasks/Windows.yml
```diff
@@ -191,8 +192,8 @@
...
+    zabbix_win_download_link: "{{ zabbix2_win_download_link | default('_win_download_link') }}"
```
pyrodie18, line 196: See the latest changes. Support for `zabbix2_win_download_link` has been removed with the release of 3.0.0

pyrodie18 (changes requested):
Thank you for the contribution.  A few comments but looks good.

roles/zabbix_agent/tasks/psk_secret.yml
emrocha: In the original code, there is two tasks that use the "register" keyword for the same variable. This two task are, indeed, mutual exclusive. But Ansible also set the variable if the task is skipped.

> If a task fails or is skipped, Ansible still registers a variable with a failure or skipped status, unless the task is skipped based on tags.

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html

roles/zabbix_agent/vars/agent2_vars.yml
emrocha: What do you suggest as an alternative?

The code below uses the variable `_win_download_link`  when the variables `zabbix_agent_win_download_link`  and `zabbix_agent2_win_download_link` are **not** set.

https://github.com/ansible-collections/community.zabbix/blob/45ed88b2abbaffb8869bcb102ac2ce5ed6694769/roles/zabbix_agent/tasks/Windows.yml#L1-L4

On the other hand, the code below uses the variable `zabbix2_win_package ` that is not set by default.

https://github.com/ansible-collections/community.zabbix/blob/45ed88b2abbaffb8869bcb102ac2ce5ed6694769/roles/zabbix_agent/vars/agent2_vars.yml#L6

Therefore, the script will fail if none of these variables are set when the Role is applied.

roles/zabbix_agent/tasks/Windows_conf.yml
emrocha: In template of zabbix_agent.conf, the parameter `Include=` **will be set** because the variable `zabbix_agent_include` is defined in the main.yml script.

https://github.com/ansible-collections/community.zabbix/blob/45ed88b2abbaffb8869bcb102ac2ce5ed6694769/roles/zabbix_agent/tasks/main.yml#L27-L29

https://github.com/ansible-collections/community.zabbix/blob/45ed88b2abbaffb8869bcb102ac2ce5ed6694769/roles/zabbix_agent/templates/agent.conf.j2#L69

So, for Windows, if the variable `zabbix_agent_include` was not explicited defined, the parameter `Include` will be set to `/etc/zabbix/zabbix_agentd.d` because of the code below.

https://github.com/ansible-collections/community.zabbix/blob/45ed88b2abbaffb8869bcb102ac2ce5ed6694769/roles/zabbix_agent/vars/agent_vars.yml#L3

roles/zabbix_agent/vars/agent2_vars.yml
emrocha: Maybe I misunderstood your comment. Do you just want to unify the parameter for Zabbix_agent and Zabbix_agent2?

roles/zabbix_agent/tasks/Windows_conf.yml
emrocha: Maybe I misunderstood your comment. Do you just want to unify the parameter for Zabbix_agent and Zabbix_agent2?

the-iot:
the agent role is fixed by commit b0f05b4 but the version 3.0.1 is not released.

emrocha:
I was a little busy this week, but I will take a look soo n.

emrocha:
> the agent role is fixed by commit [b0f05b4](https://github.com/ansible-collections/community.zabbix/commit/b0f05b41f189b0bedc1090e6b1632efecc7e4aad) but the version 3.0.1 is not released.
> 
> see [#1301 (comment)](https://github.com/ansible-collections/community.zabbix/pull/1301#event-13250962392)

The code below in zabbix_agent role still include the role zabbix_repo for Windows hosts.

https://github.com/ansible-collections/community.zabbix/blob/28c4921182f45fe0f16796bbb496b1d95016f316/roles/zabbix_agent/tasks/main.yml#L43-L52

In my view, the correct logic in the "when" statement shoud be: 
```yaml
 when:
    - zabbix_manage_repo | default(true)
    - not (zabbix_agent_docker | bool)
    - ansible_os_family != "Windows"
```

In the role zabbix_repo, the code below tries to load the Windows.yml file that do not exists in role zabbix_repo.

https://github.com/ansible-collections/community.zabbix/blob/28c4921182f45fe0f16796bbb496b1d95016f316/roles/zabbix_repo/tasks/main.yml#L7-L8

For this reason, the script still does not work for Windows Hosts.

david-sieg:
Yes mate, you're right. Can u open a PR?

emrocha:
@david-sieg 
I made the correction in the "when" statement for the "Install Repository" task now.
But a keep it in this pull request that include others changes. Is the ok for you?

@pyrodie18

I remove some "zabbix2_*" variables. But not all. The code below is necessary because the script tries to remove old versions and need to test both zabbix agents (1 and 2).

https://github.com/ansible-collections/community.zabbix/blob/28c4921182f45fe0f16796bbb496b1d95016f316/roles/zabbix_agent/tasks/Windows.yml#L28-L35

david-sieg:
That‘s ok for me, thx!

the-iot:
please add merge into main and release 3.0.2 - momentary we cant install zabbix agent on windows with release 3.x

pyrodie18:
OK, this looks good to me.  The only thing still needed is change fragment(s).

emrocha:
> OK, this looks good to me. The only thing still needed is change fragment(s).

I didn't understand. What do you mean by "fragments"?

Do I need to change anything else?

BGmot:
> > OK, this looks good to me. The only thing still needed is change fragment(s).
> 
> I didn't understand. What do you mean by "fragments"?
> 
> Do I need to change anything else?

You need to create a file describing your changes here https://github.com/ansible-collections/community.zabbix/tree/main/changelogs/fragments

Here you can read instructions about how to create changelog fragments https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment

the-iot:
ok...merge is done.

please add tag for the new version 3.0.2 and publish to galaxy

PR #1291, opened 2024-06-13
Clean-up Vars

pyrodie18:
##### SUMMARY
Verified all vars across all roles.  Changed all configuration bool vars from 1/0 to true/false.  Fixed the problem reported in #1237 .


BGmot (approved):
Great job!

Issue #1297, opened 2024-06-18
[Zabbix Agent] TLSPSKFile config is always set breaking configuration


Kajot-dev:
##### SUMMARY

Zabbix agent role has a default value for `zabbix_agent_tlspskfile` set. Then when templating `.conf` file 
https://github.com/ansible-collections/community.zabbix/blob/45ed88b2abbaffb8869bcb102ac2ce5ed6694769/roles/zabbix_agent/templates/agent.conf.j2#L138

There is only check whether this variable is not defined or null. This leads to invalid configuration when using `zabbix_agent_tlsconnect` or `zabbix_agent_tlsaccept` is set to for example `cert`:
```
cannot use encryption configuration: TLSPSKFile configuration parameter set without PSK being used
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

Zabbix agent role

##### ANSIBLE VERSION

```paste below
2.17.0
```

##### OS / ENVIRONMENT / Zabbix Version

Zabbix 7.0

##### STEPS TO REPRODUCE

Try to configure zabbix agent with the role using "cert" mode. Role completes successfully but zabbix-agent2 service fails instantly

##### EXPECTED RESULTS

TLSPSKFile is not set when `zabbix_agent_tlsconnect` and/or `zabbix_agent_tlsconnect` is set to something different than `psk`.

##### ACTUAL RESULTS

```paste below
cannot use encryption configuration: TLSPSKFile configuration parameter set without PSK being used
```

Kajot-dev:
And using newest release `3.0.0`

Kajot-dev:
This is even worse. Situation where `zabbix_agent_tlspskfile` is undefined or `None` is not even possible because of:
https://github.com/ansible-collections/community.zabbix/blob/45ed88b2abbaffb8869bcb102ac2ce5ed6694769/roles/zabbix_agent/tasks/psk_secret.yml#L21
Which will fail

silveraignacio:
+1 to this!

PR #1298, opened 2024-06-19
DRAFT:  Fixed PSK Logic

pyrodie18:
##### SUMMARY
This change adds logic that that the PSK identify and secret files are touched unless encryption is set to PSK and solves #1297.

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/changelogs/fragments/1297.yml
+++ b/changelogs/fragments/1297.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - Agent Role - Fixed logic problem that would break if anything other than PSK was used.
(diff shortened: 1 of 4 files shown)


K3ndu:
"msg": "The conditional check 'zabbix_agent_tlsaccept == 'psk' or zabbix_agent_tlsconnect == 'psk'' failed. The error was: error while evaluating conditional (zabbix_agent_tlsaccept == 'psk' or zabbix_agent_tlsconnect == 'psk'): 'zabbix_agent_tlsaccept' is undefined

Should be (zabbix_agent_tlsaccept is defined and zabbix_agent_tlsaccept == 'psk') or (zabbix_agent_tlsconnect is defined and zabbix_agent_tlsconnect == 'psk')

Kajot-dev:
This solves the problem partially because `zabbix_agent_tlspskfile` has default value and thus config will be always in the configuration (event if it's not supposed to). The same check can be added to configuration template (or default value can be changed).

With the current state of this PR this will be possible to mitigate by explicitly setting `zabbix_agent_tlspskfile` to `None`/`null`

pyrodie18:
OK I think this will fix it.

Issue #1300, opened 2024-06-20
[Zabbix Agent] Windows Agent installation is executed 2 times


david-sieg:
##### SUMMARY
Windows Agent installation is executed 2 times due to include_role task.

[Task](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/tasks/main.yml#L41)

If variable zabbix_manage_repo isn't set to false for Windows systems, task Windows.yml will be executed from:
[Task](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_repo/tasks/main.yml#L7)

Specific task file isn't available in role so Ansible falls back to Windows.yml from role-zabbix-agent.

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

Ansible 2.16

##### OS / ENVIRONMENT / Zabbix Version

Zabbix 6.0.30
Windows Server 2022

##### EXPECTED RESULTS
Running installation with one execution.

##### ACTUAL RESULTS
Agent installation is executed 2 times.

the-iot:
please add merge into main and release 3.0.2 - momentary we cant install zabbix agent on windows with release 3.x

PR #1301, opened 2024-06-20
Fix windows agent installation #1300

david-sieg:
##### SUMMARY

Extend when condition for Docker and Windows.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

[include_role](https://github.com/ansible-collections/community.zabbix/blob/8805f3267854a5cb9f5502d00c32754ac996484a/roles/zabbix_agent/tasks/main.yml#L41)

##### ADDITIONAL INFORMATION
This PR should fix [#1300](https://github.com/ansible-collections/community.zabbix/issues/1300)


pyrodie18 (changes requested):
Thanks for that.  Only thing I say that needs to be changed.  Undo your change to changelog.yml and instead follow the instructions at https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment for putting in a change fragement.

pyrodie18:
@david-sieg  you're failing sanity checks.  I'm guessing you're on a Windows system which by default uses a combined /r/n for new lines instead of the /n by itself that *Nix uses.  This in the change fragment is causing it to fail.  Also the format for the change fragment is `- Agent Role - < My Description>`

pyrodie18 (approved):

PR #1312, opened 2024-06-30
Fixed problem with include file

### LINKED ISSUE #1307: zabbix_server_include should not include `*.conf`
##### SUMMARY
Since https://github.com/ansible-collections/community.zabbix/pull/1291 the variable `zabbix_server_include` is now `/etc/zabbix/zabbix_server.conf.d/*.conf` ([main.yml#L59](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_server/defaults/main.yml#L59)). When the task [Create include dir zabbix-server](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_server/tasks/main.yml#L75) runs, it creates an unwanted sub-folder `*.conf` under `/etc/zabbix/zabbix_server.conf.d/`!

Maybe it should be splitted in two variables?
```yaml
zabbix_server_include_path: '/etc/zabbix/zabbix_server.conf.d'
zabbix_server_include_pattern: '*.conf'
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix_server

pyrodie18:
##### SUMMARY
This fixes a problem introduced in 3.0.0 that causes the include directory to be improperly named.  This resolves #1307 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
Server and agent roles

### DIFF
--- a/changelogs/fragments/1307.yml
+++ b/changelogs/fragments/1307.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zabbix_agent and zabbix_server roles - Fixed problem with include file
--- a/changelogs/fragments/1307.yml
+++ b/changelogs/fragments/1307.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zabbix_agent and zabbix_server roles - Fixed problem with include file
+  - zabbix_agent, zabbix_proxy, and zabbix_server roles - Fixed problem with include file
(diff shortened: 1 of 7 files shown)


changelogs/fragments/1307.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zabbix_agent and zabbix_server roles - Fixed problem with include file
```
BGmot, line 2:
```suggestion
  - zabbix_agent, zabbix_server and zabbix_server roles - Fixed problem with include file
```

PR #1316, opened 2024-07-01
Re-added additional include for agent

Thulium-Drake:
##### SUMMARY

This PR restores #1059 to version 3.0.0 of the collection.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_agent role

##### ADDITIONAL INFORMATION

This was originally added in #1059 but lost when making version 3.0.0 with the revamped config file for the agent.

### DIFF
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -24,6 +24,7 @@ zabbix_agent_forceactivechecksonstart: false
 zabbix_agent_heartbeatfrequency: 60
 zabbix_agent_hostname: "{{ inventory_hostname }}"
 zabbix_agent_include_mode: "0750"
+zabbix_agent_additional_includes: []
 zabbix_agent_listenip: 0.0.0.0
 zabbix_agent_listenport: 10050
 zabbix_agent_logfilesize: 100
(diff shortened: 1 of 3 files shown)


pyrodie18:
We're already including everything in /etc/zabbix/zabbix_agent2.d.  Is there a reason that isn't sufficient?

Thulium-Drake:
No, because, what's also mentioned in #1056 , the includes are not recursive, each directory much be included on it's own.
And the default installation directory for all Zabbix Agent2 Plugins is ```/etc/zabbix/agent2.d/plugins.d```.

pyrodie18:
OK fair.  That being the case, I would do this like how we currently handle `alias` instead of having `include` and `include_additional`.

roles/zabbix_agent/defaults/main.yml
```diff
@@ -24,6 +24,7 @@ zabbix_agent_forceactivechecksonstart: false
 zabbix_agent_heartbeatfrequency: 60
 zabbix_agent_hostname: "{{ inventory_hostname }}"
 zabbix_agent_include_mode: "0750"
+zabbix_agent_additional_includes: []
```
BGmot, line 27:
```suggestion
zabbix_agent_additional_include: []
```
Please decide whether it is `zabbix_agent_additional_includes` or `zabbix_agent_additional_include`. I would go with `zabbix_agent_additional_includes`, i.e. change this variable in other files but up to you.

Thulium-Drake:
@pyrodie18 Thanks for the feedback! I'll rework it based on the ```alias``` logic then. I'll took a look at it and make a new PR :-)

PR #1335, opened 2024-07-09
zabbix_agent- Reworked Include logic similar to the Alias logic

Thulium-Drake:
##### SUMMARY

Reworks the 'Include' logic to work in the same manner as 'Alias'.

Also updates the included paths for Zabbix Agent2 to include the plugins.d directory.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_agent
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/1335-rework-include-logic.yml
+++ b/changelogs/fragments/1335-rework-include-logic.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - zabbix_agent Role - Reworked Include logic based on Alias logic
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -136,7 +136,7 @@ zabbix_agent_docker_ports:
 zabbix_agent_docker_security_opts:
   - apparmor:unconfined
 zabbix_agent_docker_volumes:
-  - /etc/zabbix/zabbix_agentd.d:{{ zabbix_agent_include_dir }}
+  - /etc/zabbix/zabbix_agentd.d:{{ zabbix_agent_docker_include_dir }}
   - /:/hostfs:ro
   - /etc:/hostfs/etc:ro
   - /proc:/hostfs/proc:ro
--- a/roles/zabbix_agent/tasks/Docker.yml
+++ b/roles/zabbix_agent/tasks/Docker.yml
@@ -1,6 +1,7 @@
 ---
 - name: "Create volume mount string"
   ansible.builtin.set_fact:
+    zabbix_agent_docker_include_dir: "{{ (zabbix_agent_include_dir is iterable) | ternary(zabbix_agent_include_dir[0], zabbix_agent_include_dir) }}"
     volume_mount: "{{ zabbix_agent_tlspskfile }}:/var/lib/zabbix/enc/tlspskfile"
     tls_key:
       ZBX_TLSPSKFILE: tlspskfile
--- a/roles/zabbix_agent/tasks/Linux.yml
+++ b/roles/zabbix_agent/tasks/Linux.yml
@@ -147,7 +147,7 @@
     group: zabbix
     mode: "{{ zabbix_agent_include_mode }}"
     state: directory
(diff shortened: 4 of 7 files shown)


pyrodie18:
Please add a change fragment to this.

pyrodie18:
@Thulium-Drake check out the testing failures

Thulium-Drake:
I've moved adding the ```*.conf``` to all paths for includes in the template. I couldn't wrap my head around something that didn't look horribly complex (and prolly also weird).

Also, it doesn't really make sense to use anything else then ```*.conf``` as the suffix for include paths.

BGmot:
Can you please add tests to molecule to cover this "use case"?

BGmot:
Oh actually tests are failing, please take a look.

Thulium-Drake:
Well I missed ensuring permissions for all included dirs on agent2 :-) and by fixing that, I also fixed the tests

pyrodie18:
We still have a problem with the definition of `zabbix_agent_docker_volumes` in the defaults file which is expecting a single item

BGmot:
@Thulium-Drake can you please add molecule tests for this case so the problem is obvious before your fix and it's gone with your code?

Thulium-Drake:
> @Thulium-Drake can you please add molecule tests for this case so the problem is obvious before your fix and it's gone with your code?

I looked at the test that failed, it checks if the ownership of all mentioned ```include_dirs``` is ```root:zabbix```. And because there's 2 scenarios, with a different amount of directories, I can't really see a way to make that 'sensing', as it will check all dirs if it's a list and only one if it's not.


> We still have a problem with the definition of `zabbix_agent_docker_volumes` in the defaults file which is expecting a single item

I've changed the docker_include_dir to a new variable, which will be set in the ```Docker.yml``` task. 

If it's a single dir, include that. If it's a list, mount the first directory on the list, as it doesn't really make sense to scatter all Zabbix config files around and keep them in subdirs in ```/etc/zabbix/zabbix_agent.d```.

Thulium-Drake:
The errors don't seem related, they are all Galaxy errors

Thulium-Drake:
The fact that role breaks agent2's include mechanism, I first made a different solution. Then @pyrodie18 suggested to re-implement that using similar logic as used for adding host ```Alias```-es

Thulium-Drake:
https://github.com/ansible-collections/community.zabbix/pull/1316

BGmot:
> The fact that role breaks agent2's include mechanism, I first made a different solution. Then @pyrodie18 suggested to re-implement that using similar logic as used for adding host `Alias`-es

Ok, can you add to molecule tests that we see it is broken and after your code changes it is fixed?

Thulium-Drake:
Well the problem is, it doesn't result into any errors in the Zabbix agent, but it does not include recursively. So it's not broken because it's spitting errors, it's broken by no longer including configuration files for, among others, ```zabbix-agent2-plugin-*```

I don't know to to make a test for that, because agent2 doesn't care if you not include the ```plugins.d``` folder

BGmot:
So we need this fix to include lets say all the files in
/etc/zabbix/zabbix_agent2.d/plugins.d/
*and*
in /etc/zabbix/zabbix_agent2.d/
?

Thulium-Drake:
Yes, even though they don't mention it explicitly in the docs (https://www.zabbix.com/documentation/current/en/manual/appendix/config/special_notes_include) it will only match on files in the included dirs..

BGmot:
The same story with zabbix-agent (the first version)?

Thulium-Drake:
OOTB, no, because agent (1) does not have plugins, but that depends on the user, if they have their own include setup, it might be something they use. Though I'd say unlikely (but that's from my own experience)

BGmot:
Ok, thanks a lot for contributing and your explanation! I am fine with this PR. @pyrodie18 please approve too as you deal with roles more that I do -)

BGmot (approved):

pyrodie18 (approved):

pyrodie18:
Thanks for the addition

Issue #1355, opened 2024-07-28
Zabbix_Host Module Failed


pyrodie18:
##### SUMMARY
Not sure what happened but at some point in the last two weeks the host module started breaking.  The last plugin-integration test to run (https://github.com/ansible-collections/community.zabbix/actions/runs/9933607178/job/28017631838) ran 100% all the way through (NOTE:  The failure there now is one that I triggered today to see if it still passed like it had previously).  @masa-orca  discovered this failure with #1353.

Failure read-out:

```
2024-07-28T16:06:46.5176821Z TASK [test_zabbix_host : test: change visible_name] ****************************
[...]

PR #1356, opened 2024-07-30
Delete denied parameters from host interfaces

masa-orca:
##### SUMMARY

- Fix zabbix_host module to run on 7.0.1 and higher.
- This commit resolve the #1355 issue.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
In my change, I copied allowed_fields from [this commit](https://git.zabbix.com/projects/ZBX/repos/zabbix/commits/3d5a48d327ec01a4c8e58460873297f9b56f576e#ui/include/classes/api/services/CHostInterface.php) which is Vanav mentioned. Therefore, it will course errorif the allowed_fields chenges.

```paste below

```

### DIFF
--- a/plugins/modules/zabbix_host.py
+++ b/plugins/modules/zabbix_host.py
@@ -944,7 +944,7 @@ class Host(ZabbixBase):
 
 
 # Add all default values to all missing parameters for existing interfaces
-def update_exist_interfaces_with_defaults(exist_interfaces):
+def update_exist_interfaces_with_defaults(exist_interfaces, zabbix_version):
 
     new_exist_interfaces = []
     default_interface = {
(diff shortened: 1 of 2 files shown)


plugins/modules/zabbix_host.py
```diff
@@ -966,9 +966,29 @@ def update_exist_interfaces_with_defaults(exist_interfaces):
...
+        'useip',
+        'main',
+        'details',
+        'interface_ref',
+        'items',
+        'interfaceid'
+    ]
     for interface in exist_interfaces:
+        normalized_interface = interface
+        if LooseVersion("7.0") <= LooseVersion(zabbix_version):
```
pyrodie18, line 984: Shouldn't this be >= ?  6.4 and 6.0 were working fine before.

masa-orca: Is it wrong code? checking fields  will runs for jost 7.0 and higher.

pyrodie18: Right now you're checking for <= 7.0.  That means it won't run in the future for 7.2, but it will run today for 6.4.  6.4 is working fine (don't think it needs the fields, but the API handles them gracefully).  Either it needs to be >= 7.0 or we just sanitize them for all versions (which I don't honestly think is necessarily wrong).

masa-orca: > Right now you're checking for <= 7.0

Does this mean integration test target is 7.0 and lower? 

>  if LooseVersion("7.0") <= LooseVersion(zabbix_version):

If 6.4 and 6.0, zabbix_host module's behavior is not chenged from current behavior.  in addition, this effects on 7.0, 7.2 and higher.


> Either it needs to be >= 7.0 or we just sanitize them for all versions (which I don't honestly think is necessarily wrong).

If you want to check for all version, I will remove version checking condition.

pyrodie18:
I'd remove the check and also add a change fragment.  From there I think we're good.

masa-orca:
Integration tests failed. however it is not depends my change and I create another pull request to fix it.
https://github.com/ansible-collections/community.zabbix/pull/1357

pyrodie18 (approved):

Issue #1365, opened 2024-08-07
Running playbook to install and configure zabbix agent causes system corruption


j007bond007:
##### SUMMARY
Deploying Zabbix agent (1) on Debian 12 machine causes system/user corruption, where the user can no longer login (error displayed on console is "Unable to cd to '/home/josh" and permissions on the home directory seem ok. My wiki software also breaks and displays "Internal Server Error". Odd warning was presented during playbook run:

`10:47:52 AM
TASK [community.zabbix.zabbix_agent : Create include dirs zabbix-agent] ********
10:47:53 AM
[WARNING]: Module invocation had junk after the JSON data: /bin/sh: 1: sleep:
10:47:53 AM
Permission denied
10:47:54 AM
fatal: [home1wikijs]: FAILED! => {"msg": "Failed to connect to the host via ssh: josh@home1wikijs: Permission denied (publickey,password)."}`

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.zabbix.zabbix_agent

##### ANSIBLE VERSION

```paste below
ansible [core 2.17.2]
  config file = /home/josh/Ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/.local/pipx/venvs/ansible/lib/python3.11/site-packages/ansible
  ansible collection location = /home/josh/Ansible/collections:/root/.ansible/collections
  executable location = /root/.local/bin/ansible
  python version = 3.11.2 (main, May  2 2024, 11:59:08) [GCC 12.2.0] (/root/.local/pipx/venvs/ansible/bin/python)
  jinja version = 3.1.4
  libyaml = True

```

##### CONFIGURATION

```paste below
COLLECTIONS_PATHS(/home/josh/Ansible/ansible.cfg) = ['/home/josh/Ansible/collections', '/root/.ansible/collections']
CONFIG_FILE() = /home/josh/Ansible/ansible.cfg
DEFAULT_HOST_LIST(/home/josh/Ansible/ansible.cfg) = ['/home/josh/Ansible/inventory/hosts']
DEFAULT_ROLES_PATH(/home/josh/Ansible/ansible.cfg) = ['/home/josh/Ansible/roles', '/root/.ansible/roles']

```

##### OS / ENVIRONMENT / Zabbix Version

[...]

Issue #1361, opened 2024-08-06
Default value of zabbix_agent_include_dir for zabbix_agent2 breaks userparameter task


BartJM:
##### SUMMARY
The default for `zabbix_agent_include_dir` when using zabbix_agent2 is `['/etc/zabbix/zabbix_agent2.d', '/etc/zabbix/zabbix_agent2.d/plugins.d']`. This is used in
https://github.com/ansible-collections/community.zabbix/blob/2010af5b723dccd39da2b5ed1f35e205543aca96/roles/zabbix_agent/tasks/userparameter.yml#L32

where it causes the `Install user-defined userparameters` task to fail with
```
"Destination directory ['/etc/zabbix/zabbix_agent2.d', '/etc/zabbix/zabbix_agent2.d/plugins.d'] does not exist"
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

zabbix_agent

##### ANSIBLE VERSION

```paste below
ansible [core 2.15.9]
  config file = None
  configured module search path = ['/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  ansible collection location = /runner/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.9.18 (main, Jan 24 2024, 00:00:00) [GCC 11.4.1 20231218 (Red Hat 11.4.1-3)] (/usr/bin/python3)
  jinja version = 3.1.3
  libyaml = True

```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT / Zabbix Version

Target: Ubuntu 22.04

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

PR #1366, opened 2024-08-08
Fix zabbix_agent include_dir creation logic

AlexPykavy:
##### SUMMARY

As a string is iterable of chars, the previous logic changed the permissions for all single-character dirs: '/', 'e', 't', 'c', '/', etc.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- zabbix_agent role

##### ADDITIONAL INFORMATION

- **IMPORTANT**: open terminal session as `root` because the current version of the role resets the `/` ownership for all users except `root` and `zabbix` group
- run the `zabbix_agent` role
- check that the ownership of the `/` is set to `root:zabbix` and the permissions are smth like `drwxr-x---`

### DIFF
--- a/roles/zabbix_agent/tasks/Linux.yml
+++ b/roles/zabbix_agent/tasks/Linux.yml
@@ -128,18 +128,6 @@
   tags:
     - config
 
-- name: "Create include dir zabbix-agent"
-  ansible.builtin.file:
-    path: "{{ zabbix_agent_include_dir }}"
-    owner: root
-    group: zabbix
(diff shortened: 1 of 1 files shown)


AlexPykavy:
Perhaps, it fixes https://github.com/ansible-collections/community.zabbix/issues/1365.

pyrodie18 (approved):
OK I think this makes sense.  Going to go ahead and pull it in but won't cut the release until #1361 gets updated.

PR #1376, opened 2024-08-22
fix: make statusport really optional

darkobas2:
Status port is supposed to be configured when zabbix_agent2_statusport is defined. Since its defined in defaults its always defined. Therefore we remove it from default

roles/zabbix_agent/templates/agent.conf.j2
  - {{ (zabbix_agent_statusport is defined and zabbix_agent_statusport is not none) | ternary('', '# ') }}StatusPort={{ zabbix_agent_statusport | default('') }}

##### COMPONENT NAME
agent2

### DIFF
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -33,7 +33,6 @@ zabbix_agent_persistentbufferperiod: 1h
 zabbix_agent_pluginsocket: /tmp/agent.plugin.sock
 zabbix_agent_plugintimeout: 3
 zabbix_agent_refreshactivechecks: 120
-zabbix_agent_statusport: 9999
 zabbix_agent_timeout: 3
 zabbix_agent_tlspsk_auto: false
 zabbix_agent_tlspskfile: /etc/zabbix/tls_psk_auto.secret


pyrodie18:
I am pretty sure that there are several variables within the configuration files that are optional, but still have a value in defaults/vars for one reason or another.  I hesitate to remove this right now for two reasons.  1.  It being there isn't breaking anything (that I'm tracking).  2.  Removing it may break a someone's existing deployment which would make this a breaking change that we would only do on a major release (don't expect we'll be doing one of those for a while since we just released 3.0.  If this is causing you problems, I would suggest defining in your vars a null value to override it.

Thoughts @BGmot, @masa-orca, @eb4x

darkobas2:
cool, ill go with null for now

BGmot:
> I am pretty sure that there are several variables within the configuration files that are optional, but still have a value in defaults/vars for one reason or another. I hesitate to remove this right now for two reasons. 1. It being there isn't breaking anything (that I'm tracking). 2. Removing it may break a someone's existing deployment which would make this a breaking change that we would only do on a major release (don't expect we'll be doing one of those for a while since we just released 3.0. If this is causing you problems, I would suggest defining in your vars a null value to override it.
> 
> Thoughts @BGmot, @masa-orca, @eb4x

Agree with your statement.

pyrodie18:
OK, I hear what @evilhamsterman is saying in #1384

pyrodie18:
forgot to hit the re-open button

PR #1380, opened 2024-08-26
Fix login and session issues connecting to zabbix

### LINKED ISSUE #1209: Zabbix httpapi does not honor reset_connection
##### SUMMARY
I'm working on Zabbix server bootstrap script, which, among other things, creates new Super admin user and disables default Admin user. Issue, that I'm facing is that zabbix httpapi is unable to drop current connection and reconnect with different credentials.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.zabbix.zabbix httpapi – HttpApi Plugin for Zabbix

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.5]
  config file = None
  configured module search path = ['/home/vaivis/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.11/site-packages/ansible
  ansible collection location = /home/vaivis/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.11.8 (main, Feb 12 2024, 14:50:05) [GCC 13.2.1 20230801] (/usr/bin/python)
  jinja version = 3.1.3
  libyaml = True

```
```paste below
Collection           Version
-------------------- -------
ansible.netcommon    6.1.0  
ansible.posix        1.5.4  
ansible.utils        4.0.0  
ansible.windows      2.3.0  
community.docker     3.4.8  
community.general    5.7.0  
community.mysql      3.9.0  
community.postgresql 3.0.0  
community.zabbix     2.3.1  

```

##### CONFIGURATION

```paste below
CONFIG_FILE() = None
EDITOR(env: EDITOR) = vim

```

* Zabbix server 6.4.13 in Ubuntu Docker container
* Zabbix web nginx 6.4.13 in Ubuntu Docker container
* Play is run just after database initialization

##### STEPS TO REPRODUCE

My approach is to try to connect with new credentials `super` `verysecretpass` and check status of default `Admin` user. If `Admin` is __not disabled__, I expect user `super` to not exist, so connect with default credentials and create user `super`. After that `Admin` can be disabled, so I drop current connection and reconnect as `super` user and disable `Admin`.

```yaml
---
- name: Ensure telemetry server is set up and running
  hosts: all
  tasks:

    - name: Configure zabbix server
      vars:
        ansible_network_os: community.zabbix.zabbix
        ansible_connection: httpapi
        ansible_httpapi_port: 8182
        ansible_httpapi_use_ssl: no
        ansible_zabbix_url_path: ""
        ansible_user: "super"
        ansible_httpapi_pass: "verysecretpass"
      block:
        - name: Ensure default admin is deactivated
          block:
            - name: Check status of default admin
              community.zabbix.zabbix_user_info:
                username: Admin
              register: zbx_admin
              failed_when: zbx_admin.zabbix_user.users_status | int == 0

          rescue:
            - name: Reset connection
              meta: reset_connection

            - name: Configure primary admin user
              vars:
                ansible_user: "Admin"
                ansible_httpapi_pass: "zabbix"
              community.zabbix.zabbix_user:
                username: "super"
                state: present
                name: "New Super"
                surname: "User"
                passwd: "verysecretpass"
                role_name: Super admin role
                usrgrps:
                  - Zabbix administrators
                  - Internal
                autologin: no
                autologout: "0"

            - name: Reset connection
              meta: reset_connection

            - name: Disable default admin user
              community.zabbix.zabbix_user:
                username: "Admin"
                passwd: "zabbix"
                usrgrps:
                  - "Disabled"

```

##### EXPECTED RESULTS

I expect this play to fail on `Ensure default admin is deactivated` and be rescued by subsequent tasks. After this play I expect super users `Admin` (disabled) and `super` (enabled). 
Subsequent runs of this play shold make no changes to user accounts. 

##### ACTUAL RESULTS

On first run play proceeds as expected until task `Disable default admin user`, at which point play fails with following output.
[Play output](https://gist.github.com/aivarsV/eb9d39a73f2c8329cc7976d55d24e77b)

After first run I'm left with two working super admin accounts - `Admin` and `super`.
Second run proceeds till the end and `Admin` get's disabled.

My assumption is that `Reset connection` task has no actual effect on zabbix httpapi module.

CircuitCipher:
* zabbix httpapi plugin - if login failed, it would not be attempted again by later tasks. fix now marks the connection as disconnected on failed login so that later tasks can try
* zabbix_user - if the api user's password was changed, session errors would occur in later tasks. fix will relogin to zabbix after changing password of the api user

## SUMMARY
Fixes #1209 because reset_connection will no longer be necessary in that context

#### Why do we need this ?
- The [zabbix docker images](https://github.com/zabbix/zabbix-docker) do not allow you to change zabbix admin user/pass:  https://github.com/zabbix/zabbix-docker/issues/820
    - This is a security decision and the Zabbix docker maintainer said you can use the API via something like Ansible to do it.

In the case of writing an Ansible role to:

* install Zabbix from these images
*  then update the Zabbix admin password in a idempotent way,

We may need to use a test login call to see if our new password works or if its still the default

- We use `failed_when: false` on the task so our play doesn't fail if we get a failed login
- If the the call fails, we can assume we have not yet setup the admin password and the default zabbix password is still in place

#### Where this fails
Currently the codebase doesn't mark the connection as disconnected when a login fails so subsequent attempts will assume we are logged in and provide a bad auth token

#### The fix
In `community/zabbix/plugins/httpapi/zabbix.py` we update the `login()` method so that when the login is failed, the connection is marked as disconnected.
```python
try:
	# Zabbix <= 6.2
	payload = self.payload_builder("user.login", user=username, password=password)
	code, response = self.send_request(data=payload)
except ConnectionError:
	# Zabbix >= 6.4
	try:
		payload = self.payload_builder("user.login", username=username, password=password)
		code, response = self.send_request(data=payload)
	 except ConnectionError:
		# If this task fails to login, we mark the connection as disconnected so that
		# future tasks can try to reauth
		self.connection._connected = False
		raise
```

#### Why do we have to do this?
It would be nice if we could detect changing the current Zabbix API user's password and re-login the user instead of seeing session terminated errors on subsequent calls

The original intent was to do this all from inside the Zabbix HTTPAPI class but it turns out the update user request is actually two separate requests made by the zabbix_user module:

* Get the user id by searching for the user by the given username using the [user.get method](https://www.zabbix.com/documentation/current/en/manual/api/reference/user/get)
* Call [user.update method](https://www.zabbix.com/documentation/current/en/manual/api/reference/user/update) to change the password by passing in the user id
	* By the time the zabbix httpapi receives the send_request for the user.update:
		* We only have the user id of the request so its hard to determine if the connection user is the user we are changing the password of!
			

#### Zabbix_user module User::update_user() method
- From within the zabbix_user module we still have access to the username of the user we are changing the password for but we don't have access to the Ansible HTTPAPI options such as the ansible_user which is our API user!
	- So we have to notify the Zabbix Ansible HTTPAPI that a user's password has changed.
	- The Ansible HTTPAPI can then determine if we changed our current user's password and re-login with the new credentials
		- We call the `password_changed_for_user()` method of the `ZabbixApiRequest` class
		- ` community/zabbix/plugins/modules/zabbix_user.py`:
		-
		  ```python
		  successfully_changed_password = user_ids and override_passwd
		  if successfully_changed_password:
		      self._zapi.password_changed_for_user(username=username, password=passwd)
		  
		  return user_ids
		  ```
#### Add the password_changed_for_user method to ZabbixApiRequest
* We need to communicate with the connection so it makes sense to encapsulate these calls inside the ZabbixApiRequest class that already has the handle.
* `community/zabbix/plugins/module_utils/api_request.py`:

```python
def password_changed_for_user(self, username: str, password: str):
    return self.connection.password_changed_for_user(username, password)
 ```

- Theres some voodoo that occurs here that involves JSON-RPC calls to get to the actual Zabbix Ansible HTTPAPI because the connection object we have here is just a [dummy from module-utils](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/connection.py)
		
#### Add password_changed_for_user method to Zabbix Ansible/HTTPAPI
- `community/zabbix/plugins/httpapi/zabbix.py`
- This is the method that ends up being called by the module through the ZabbixApiRequest class and the JSON-RPC calls.
- Here we detect if the user whose password was changed is the same user we are using for the Zabbix API requests.
	- If so, we login with the new credentials
	
	  ```python
	  def password_changed_for_user(self, username: str, password: str):
	     '''
	     An event handler for when a zabbix user's password has been changed.
	     We can then check to see if the API user's password has been changed.
	     If so, we re-login so that the next modules are authenticated with the new password.
	  
	     :param str username: Username of the user whose password was changed
	     :param str password: The new password
	     '''
	     changed_current_user_pass = self.connection.get_option('remote_user') == username
	     if changed_current_user_pass:
	        self.login(username, password)
	     return 
	  ```
#### JSON-RPC communication between module and connection
- JSON-RPC requests are handled by the [JsonRpcServer class](https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/jsonrpc.py)
- Modules and their connection usually reside on different machines so their communication goes over RPC
- JSON-RPC request and responses must have the same ID
- By returning nothing from the `password_changed_for_user()`method, the marshaller for the json-rpc requests automatically [adds the header data](https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/jsonrpc.py#L80)  including the correct request id

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
* Zabbix HTTPAPI
* zabbix_user module

##### ADDITIONAL INFORMATION
See the integration test for a playbook on how to reproduce the issue: tests/integration/targets/test_zabbix_user/tasks/change_api_user_pass.yml

### DIFF
--- a/changelogs/fragments/relogin_after_changing-api-users-password.yml
+++ b/changelogs/fragments/relogin_after_changing-api-users-password.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - zabbix httpapi plugin - if login failed, it would not be attempted again by later tasks. fix now marks the connection as disconnected on failed login so that later tasks can try 
+  - zabbix_user - if the api user's password was changed, session errors would occur in later tasks. fix will relogin to zabbix after changing password of the api user
--- a/plugins/httpapi/zabbix.py
+++ b/plugins/httpapi/zabbix.py
@@ -99,8 +99,15 @@ class HttpApi(HttpApiBase):
             code, response = self.send_request(data=payload)
         except ConnectionError:
             # Zabbix >= 6.4
-            payload = self.payload_builder("user.login", username=username, password=password)
-            code, response = self.send_request(data=payload)
+            try:
+                payload = self.payload_builder("user.login", username=username, password=password)
+                code, response = self.send_request(data=payload)
+            except ConnectionError:
+                # If this task fails to login, we mark the connection as disconnected so that
+                # future tasks can try to reauth
+                self.connection._connected = False
+                raise
+
 
         if code == 200 and response != '':
             # Replace auth with real api_key we got from Zabbix after successful login
@@ -194,6 +201,20 @@ class HttpApi(HttpApiBase):
                 return 404, "Object not found"
         except Exception as e:
             raise e
+
+    def password_changed_for_user(self, username: str, password: str):
+        '''
+        A handler for when a Zabbix user's password has been changed.
+        We can then check to see if the API user's password has been changed.
+        If so, we re-login so that the next modules are authenticated with the new password.
+
+        :param str username: Username of the user whose password was changed
+        :param str password: The new password
+        '''
+        changed_current_user_pass = self.connection.get_option('remote_user') == username
+        if changed_current_user_pass:
+            self.login(username, password)
+        return
 
     def _display_request(self, request_method, path):
         self.connection.queue_message(
--- a/plugins/httpapi/zabbix.py
+++ b/plugins/httpapi/zabbix.py
@@ -108,7 +108,6 @@ class HttpApi(HttpApiBase):
                 self.connection._connected = False
                 raise
 
-
         if code == 200 and response != '':
             # Replace auth with real api_key we got from Zabbix after successful login
             self.connection._auth = {'auth': response}
--- a/plugins/module_utils/api_request.py
+++ b/plugins/module_utils/api_request.py
@@ -52,6 +52,19 @@ class ZabbixApiRequest(object):
     def api_version(self):
         return self.connection.api_version()
 
+    def password_changed_for_user(self, username: str, password: str):
+        '''
+        Call to the Zabbix HTTPAPI handler for when a Zabbix user's password has been changed.
+
+        This info is used to determine if the API user's password was changed and re-login
+        if so.
+
+        :param str username: Username of the user whose password was changed
+        :param str password: The new password
+        '''
+        # module_utils.connection.Connection will use JSONRPC to call back to Zabbix HTTPAPI
+        return self.connection.password_changed_for_user(username, password)
+
     @staticmethod
     def payload_builder(method_, params, jsonrpc_version='2.0', reqid=str(uuid4()), **kwargs):
         req = {'jsonrpc': jsonrpc_version, 'method': method_, 'id': reqid}
--- a/plugins/modules/zabbix_user.py
+++ b/plugins/modules/zabbix_user.py
@@ -663,6 +663,10 @@ class User(ZabbixBase):
                     msg="Failed to update user %s: %s" % (username, e)
                 )
 
+        successfully_changed_password = user_ids and override_passwd
+        if successfully_changed_password:
+            self._zapi.password_changed_for_user(username=username, password=passwd)
+
         return user_ids
 
     def delete_user(self, zbx_user, username):
--- a/tests/integration/targets/test_zabbix_user/tasks/change_api_user_pass.yml
+++ b/tests/integration/targets/test_zabbix_user/tasks/change_api_user_pass.yml
@@ -42,10 +42,15 @@
       override_passwd: "true"
       user_medias:
       usrgrps:
-        - Internal
         - Zabbix administrators
     when: "user_info.exception is not defined "
 
+  # Originally I setup an "Always:" block to reset the password back to default but
+  # because the password must be at least 8 char's long and the default pass 'zabbix' is not,
+  # this causes issues!
+  - name: Update the zabbix password var
+    set_fact:
+      zabbix_api_login_pass: "{{ new_pass }}"
 
   # Thanks to the modifications we no longer receive the following error on this task:
   # '{'code': -32602, 'message': 'Invalid params.', 'data': 'Session terminated, re-login, please.'}'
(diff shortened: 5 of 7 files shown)


CircuitCipher:
@pyrodie18  Thanks for running the tests.  I will look into the failures and hopefully have a resolution later today.

CircuitCipher:
So there are a few issues with the build as it stands.  It will take me a few days to address these issues but I have a roadmap. Unfortunately its become quite a rabbit hole of bugs and or constraints.

I noted that tests are failing strangely for zabbix version 6.0.  I will have to look into this further after I resolve the other issues.

# Tests after zabbix_user module changes zabbix password fail
I see the tests are sharing the same zabbix instance and I thought maybe we could try rewriting the test so that its changing the password of another user.  Unfortunately, due to how connections are reused, this does not trigger the required code branch.

My current approach is to use the zabbix_authentication module to change the passwd_check_rules so that we can reset the password back to the default "zabbix"

Unfortunately, there is a bug in the zabbix_authentication module when updating the passwd_check_rules and you cannot set it to an empty list to turn off all checks.
	
As such,  I will be putting together the required tests and fixes to to address the sub-bugs and hopefully get this PR into the green

BGmot:
Sorry for coming so late to this discussion (being busy with real life).
While I appreciate your efforts and so detailed description of the issue I don't like what you are proposing in this PR. Let me take a closer look at https://github.com/ansible-collections/community.zabbix/issues/1209

CircuitCipher:
> Sorry for coming so late to this discussion (being busy with real life). While I appreciate your efforts and so detailed description of the issue I don't like what you are proposing in this PR. Let me take a closer look at #1209

No problem, let me know your preferred direction for solving this issue.

BGmot:
The issue solved by https://github.com/ansible-collections/community.zabbix/pull/1386
Closing this one.

PR #1393, opened 2024-09-14
The api task os role zabbix_agent should be executed with explicit become false

emrocha:
##### SUMMARY

Undo some changes of commit 5480ff750bc9dd9133cce7156e229853923761e9 

The community.zabbix.zabbix_host module with the delegate_to parameter should not be executed with "become true" if the role zabbix_agent is applied with "become true".

It is possible that the user executing the ansible script has sudo permission on the remote computer that the zabbix_agent role will be executed on, but does not have sudo permission on the local computer from which the ansible script is initially executed.

The same applies to the other modules of this task.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
role zabbix_agent

### DIFF
--- a/changelogs/fragments/1393-fix-api-tasks-in-zabbix-agent-role-to-run-without-sudo.yml
+++ b/changelogs/fragments/1393-fix-api-tasks-in-zabbix-agent-role-to-run-without-sudo.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zabbix_agent Role - Set become parameter explicitly to false for API tasks to run without sudo on the local computer.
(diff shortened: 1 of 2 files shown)


pyrodie18:
@BGmot see any issues with this?  I don't think it will break anything.

BGmot (approved):

pyrodie18:
@emrocha add a change fragment please

emrocha:
@pyrodie18
By "add a changelog fragment please", do you mean edit the changelogs/fragments/monitored_by.yml file or create a new file in this directory? Or something else?

pyrodie18:
https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment

PR #1215, opened 2024-04-23
 define postgresql_privs due to priv deprecated in postgresql_user module | Zabbix Proxy

mmayabi:
##### SUMMARY

When installing Zabbix Proxy with PostgreSQL, due to the deprecation of the "priv" parameter in the "community.postgresql.postgresql_user" module, the required permissions for "zabbix_proxy" are not defined. As a result, the schema import to the database is not performed.
For this reason, the "priv" parameter has been removed, and a new task is defined using the "community.postgresql.postgresql_privs" module.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

PostgreSQL | GRANT ALL PRIVILEGES ON SCHEMA public TO zabbix_proxy user

##### ADDITIONAL INFORMATION

Tested on PostgreSQL 15.5 (Delegated)

```paste below

```

PR #1397, opened 2024-09-25
zabbix_proxy database user creation tasks updated to match zabbix_server

karl1k:
##### SUMMARY
zabbix_proxy role install fails on Debian 12 with postgres 15 because of issues with permissions, same issue as in #1215.
I've simply mirrored the tasks from zabbix_server to the zabbix_proxy role as that role works perfectly fine with the same OS and pgsql version.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_proxy, initialize-pgsql.yml

##### ADDITIONAL INFORMATION
Tested on Debian 12 with Postgres 15.8


eb4x:
Thank you for your contribution :+1:

PR #378, opened 2021-05-17
Possibility to disable StatusPort

msl0:
Disable StatusPort if variable zabbix_agent2_statusport not defined #371 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

community.zabbix.zabbix_agent

PR #1400, opened 2024-10-05
fix: make statusport really optional

darkobas2:
Status port is supposed to be configured when zabbix_agent2_statusport is defined. Since its defined in defaults its always defined. Therefore we remove it from default

roles/zabbix_agent/templates/agent.conf.j2
  - {{ (zabbix_agent_statusport is defined and zabbix_agent_statusport is not none) | ternary('', '# ') }}StatusPort={{ zabbix_agent_statusport | default('') }}

- replacement patch for #1376 as it was already deleted

### DIFF
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -33,7 +33,6 @@ zabbix_agent_persistentbufferperiod: 1h
 zabbix_agent_pluginsocket: /tmp/agent.plugin.sock
 zabbix_agent_plugintimeout: 3
 zabbix_agent_refreshactivechecks: 120
-zabbix_agent_statusport: 9999
 zabbix_agent_timeout: 3
 zabbix_agent_tlspsk_auto: false
 zabbix_agent_tlspskfile: /etc/zabbix/tls_psk_auto.secret


pyrodie18:
@darkobas2 thanks for the PR.  As I said in the ticket though, we'll pull this in for 4.0 but I don't expect us to be releasing that any time soon.

Thanks,
Troy

BGmot:
@darkobas2 what's the reason you don't like what's in defaults? What does it break for you?

darkobas2:
If i remember correctly is that i dont use status port at all.. and there fore i had to explicitly disable it to not have it error out.

BGmot:
How did it error out? If you don't us it then it just goes into config with default value.

darkobas2:
For me statusport was enabled but the port was set to an empty string so zabbix didnt start

BGmot:
Please re-produce with minimal playbook and give us your code and Zabbix agent configuration file you got.
I don't understand how it might happen -(

darkobas2:
it seems all is well now... i guess I had added the empty variable at some point (https://github.com/ansible-collections/community.zabbix/pull/378) which I then didnt remove

BGmot:
Ok to close this PR?

darkobas2:
Sure

PR #1404, opened 2024-10-14
Use SSL for database connection when zabbix_server_dbtlsconnect is set

ChandlerSwift:
##### SUMMARY
A reworking of https://github.com/ansible-collections/community.zabbix/pull/951

I am setting up Zabbix with a database where SSL is enforced. Zabbix itself works fine with this, thanks to the `zabbix_server_dbtlsconnect` parameter, but a few of the playbook's setup tasks try to connect without TLS, which fails. This adds a parameter to enable TLS if the server would also be set to use it.

This isn't a perfect solution, but the `community.mysql` collection doesn't provide a way to use a TLS connection without setting one of `check_hostname`, `ca_cert`, `client_key`, or `client_hostname`, and we don't (can't, in our case) specify a cert/key. This _will_ potentially fail when the host presents a cert with the wrong hostname (which can happen if `zabbix_server_dbtlsconnect` is set to `required` rather than `verify_ca` or `verify_full`). However, there's not a way to enable TLS without setting one of these options without _also_ changing the MySQL collection.

There's some more background on this issue in the `community.mysql` collection: https://github.com/ansible-collections/community.mysql/issues/90

(I'm not convinced this is the right approach, though it does fix our particular use case! If there's another way to fix this that would be better, I'd appreciate the feedback and can try to update the PR if desired.)

I've tested this to work in my configuration, and believe the only configuration it should break is cases where the server has a valid TLS certificate but assigned to the wrong hostname, as mentioned above.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_server role

### DIFF
--- a/roles/zabbix_server/tasks/initialize-mysql.yml
+++ b/roles/zabbix_server/tasks/initialize-mysql.yml
@@ -45,6 +45,7 @@
         login_host: "{{ zabbix_server_mysql_login_host | default(omit) }}"
         login_port: "{{ zabbix_server_mysql_login_port | default(omit) }}"
         login_unix_socket: "{{ zabbix_server_mysql_login_unix_socket | default(omit) }}"
+        check_hostname: "{{ zabbix_server_dbtlsconnect is defined and zabbix_server_dbtlsconnect != '' }}"
         name: "{{ zabbix_server_dbname }}"
         encoding: "{{ zabbix_server_dbencoding }}"
         collation: "{{ zabbix_server_dbcollation }}"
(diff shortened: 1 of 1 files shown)


eb4x:
Thanks, I'll have a look at this and set up some test scenarios over the weekend to get a better understanding of the issue. We might want the same type of functionality on the postgres side aswell.

ChandlerSwift:
Hey @eb4x, anything I can do to help this along?

PR #1430, opened 2024-12-07
Multiple agenet listening IPs

### LINKED ISSUE #1429: agent roles failed when multiple addresses set to zabbix_agent_listenip
### Summary

Role fails when zabbix_agent_listenip is set to multiple IPs seperated by comma. According to the Zabbix Documentation it is possible to set multiple IPs. And since the role defaults to 0.0.0.0 when zabbix_agent_listenip is not defined it comes with alot of side effects. So either the zabbix_agent_listenip has to support multiple IPs or the default behavior should be changed when not defined ListenIP is not set in the agent config as that then would result in the expect bahavior during Autoregistration of a host that the Host IP is used and not the ListenIP as Agent IP on the server. 

### Issue Type

Bug Report

### Component Name

Zabbix-agent role

### Ansible Version

```console
$ ansible --version
core 2.15.0
```

### Configuration

```console
# if using a version older than ansible-core 2.12 you should omit the '-t all'
$ ansible-config dump --only-changed -t all
```

### OS / Environment

RHEL8

### Steps to Reproduce

```yaml (paste below)
- set zabbix_agent_listenip to 192.168.1.101,127.0.0.1
```

### Expected Results

The ListenIP value in config is set to the value given. As per Zabbix Documentation the ListenIP value can have multiple IPs seperated by comma. Role shouldn't fail when the variable is set to multiple ListenIPs

### Actual Results

```console
The Role fails and says zabbix_agen_listenip isn't set
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

BGmot:
##### SUMMARY
Fixes #1429 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Zabbix agent role

### DIFF
--- a/roles/zabbix_agent/tasks/Linux.yml
+++ b/roles/zabbix_agent/tasks/Linux.yml
@@ -75,11 +75,13 @@
 
 - name: "Fail invalid specified agent_listenip"
   ansible.builtin.fail:
-    msg: "The agent_listenip does not exist"
+    msg: "{{ item }} The agent_listenip does not exist"
   when:
-    - zabbix_agent_listenip != '0.0.0.0'
-    - zabbix_agent_listenip != '127.0.0.1'
-    - (zabbix_agent_listenip not in ansible_all_ipv4_addresses)
+    - item != '0.0.0.0'
+    - item != '127.0.0.1'
+    - item not in ansible_all_ipv4_addresses
+  with_items:
+    "{{ zabbix_agent_listenip | split(',') }}"
   tags:
     - config
 
--- a/roles/zabbix_agent/tasks/Linux.yml
+++ b/roles/zabbix_agent/tasks/Linux.yml
@@ -75,7 +75,7 @@
 
 - name: "Fail invalid specified agent_listenip"
   ansible.builtin.fail:
-    msg: "{{ item }} The agent_listenip does not exist"
+    msg: "The {{ item }} (agent_listenip) does not exist"
   when:
     - item != '0.0.0.0'
     - item != '127.0.0.1'
--- a/changelogs/fragments/zabbix_agent_listenip.yml
+++ b/changelogs/fragments/zabbix_agent_listenip.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - zabbix_agent role - accept several IPs in `zabbix_agent_listenip` variable.


roles/zabbix_agent/tasks/Linux.yml
```diff
@@ -75,11 +75,13 @@
...
+    msg: "{{ item }} The agent_listenip does not exist"
   when:
-    - zabbix_agent_listenip != '0.0.0.0'
-    - zabbix_agent_listenip != '127.0.0.1'
-    - (zabbix_agent_listenip not in ansible_all_ipv4_addresses)
+    - item != '0.0.0.0'
+    - item != '127.0.0.1'
+    - item not in ansible_all_ipv4_addresses
+  with_items:
+    "{{ zabbix_agent_listenip | split(',') }}"
```
masa-orca, line 84: We need to separate with `;` for HA.

https://www.zabbix.com/documentation/6.0/en/manual/concepts/server/ha

roles/zabbix_agent/tasks/Linux.yml
```diff
@@ -75,11 +75,13 @@
 
 - name: "Fail invalid specified agent_listenip"
   ansible.builtin.fail:
-    msg: "The agent_listenip does not exist"
+    msg: "{{ item }} The agent_listenip does not exist"
```
masa-orca, line 78: Sorry for pointing out the details.
" The {{ item }}(agent_listenip) does not exist" is better to me.

roles/zabbix_agent/tasks/Linux.yml
BGmot: I am sorry, this is listening IP on the agent side? I don't think it can have a semicolon https://www.zabbix.com/documentation/7.0/en/manual/appendix/config/zabbix_agent2#listenip

roles/zabbix_agent/tasks/Linux.yml
BGmot: You are totally right, forgot to update after debugging. Updated.

roles/zabbix_agent/tasks/Linux.yml
masa-orca: Oh, sorry.
I understand it.

masa-orca (approved):

Issue #1340, opened 2024-07-10
Task `Add SEmodule to fix SELinux issue: zabix_proxy_alerter.sock` is not idempotent


andrew-landsverk-win:
##### SUMMARY
Ever since updating to collection 3.0.0 (and also 3.0.3) this task runs and is marked as changed every time we run ansible. It happens with selinux set to permissive or enforcing.

I notice that this task expects a file to be created at `/etc/selinux/targeted/active/modules/400/zabbix_proxy_add/cil` but no such file is created on Rocky Linux 9 at least.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix_proxies role

##### ANSIBLE VERSION

```paste below
ansible [core 2.15.3]
  config file = /home/andrew.landsverk/code/ansible/ansible.cfg
  configured module search path = ['/home/andrew.landsverk/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/andrew.landsverk/.local/lib/python3.9/site-packages/ansible
  ansible collection location = /home/andrew.landsverk/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/andrew.landsverk/.local/bin/ansible
  python version = 3.9.18 (main, May 16 2024, 00:00:00) [GCC 11.4.1 20231218 (Red Hat 11.4.1-3)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### CONFIGURATION

```paste below
ANSIBLE_FORCE_COLOR(/home/andrew.landsverk/code/ansible/ansible.cfg) = True
ANSIBLE_NOCOWS(/home/andrew.landsverk/code/ansible/ansible.cfg) = False
CALLBACKS_ENABLED(/home/andrew.landsverk/code/ansible/ansible.cfg) = ['profile_tasks', 'redhat.satellite.foreman']
CONFIG_FILE() = /home/andrew.landsverk/code/ansible/ansible.cfg
DEFAULT_FORKS(/home/andrew.landsverk/code/ansible/ansible.cfg) = 20
DISPLAY_SKIPPED_HOSTS(/home/andrew.landsverk/code/ansible/ansible.cfg) = False
MAX_FILE_SIZE_FOR_DIFF(/home/andrew.landsverk/code/ansible/ansible.cfg) = 1044480
TASK_TIMEOUT(/home/andrew.landsverk/code/ansible/ansible.cfg) = 900
```

[...]

PR #1433, opened 2024-12-19
Idempotentifies "Add SEmodule to fix SELinux issue: zabbix_alerter.sock" task for Zabbix Server

tux-repos:
##### SUMMARY
The way SELinux was initially applied resulted in a  "CHANGED" every time the task was executed.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_server

##### ADDITIONAL INFORMATION
Originally the SELinux context was applied using a .bsx file, this seems to be some sort of self-extracting shell script.
I have taken the steps that were used in that file and translated them into Ansible tasks, this also makes the play more transparent and also idempotent.

With a little bit of elbow grease you could apply these same steps for #1340 and resolve this issue as it stems from the same cause.


eb4x:
I've been meaning of getting around to that annoying selinux compiled file, cause I'm not that fond of it either. If you've reversed its content and it's represented in the .te file you provided, maybe we should just drop this part completely and install the `zabbix-selinux-policy` package from zabbix themselves?

https://www.zabbix.com/documentation/current/en/manual/installation/install_from_packages/rhel#selinux-configuration

I've just held off on this because it might break the original contributors setup, but the contents of `zabbix-selinux-policy` seems to have most of it covered, provided here for convenience;

```
module zabbix_policy 1.2;
require {
  type zabbix_t;
  type zabbix_port_t;
  type zabbix_var_run_t;
  type postgresql_port_t;
  type httpd_t;
  class tcp_socket name_connect;
  class sock_file { create unlink };
  class unix_stream_socket connectto;
}
#============= zabbix_t ==============
allow zabbix_t self:unix_stream_socket connectto;
allow zabbix_t zabbix_port_t:tcp_socket name_connect;
allow zabbix_t zabbix_var_run_t:sock_file create;
allow zabbix_t zabbix_var_run_t:sock_file unlink;
allow httpd_t zabbix_port_t:tcp_socket name_connect;
#============= httpd_t ==============
allow httpd_t postgresql_port_t:tcp_socket name_connect;
```

Curious to know what @pyrodie18 and @BGmot think?

eb4x:
I had a second look, and extracted the tar.gz archive from the .bsx file, and saw the contents of it. So all that checks out. Then I had a third look, and we're already installing the zabbix-selinux-policy package for zabbix_server.

I think the correct approach is just to remove the .bsx script tasks all together in zabbix_server. Then do the same for zabbix_proxy, and make sure we install the `zabbix-selinux-policy` there aswell.

Where is this `zabbix_proxy_alerter.sock` in #1340 even getting created at this point? From what I can tell, it's just bad configuration where its path is set to `/tmp` instead of `/var/lib/zabbix`, or atleast that's what I've gathered from this other sock related issue; https://www.linuxquestions.org/questions/linux-security-4/selinux-preventing-write-access-on-the-sock_file-4175710406/

tux-repos:
Oh wow, totally skipped over that because I had my attention on the .bsx task. I think you're right, these tasks are superfluous in that case. Much more convenient fix to just snip it out :)

eb4x:
> Oh wow, totally skipped over that because I had my attention on the .bsx task. I think you're right, these tasks are superfluous in that case. Much more convenient fix to just snip it out :)

You wanna take a second stab at this and strip the task out instead?

tux-repos:
Made a new PR with a clean fork; closing this one

PR #1447, opened 2025-01-24
Add the ability to rename objects

aplathrop:
##### SUMMARY
Adds the ability to rename objects in Zabbix

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- zabbix_item
- zabbix_trigger
- zabbix_discoveryrule
- zabbix_itemprototype
- zabbix_triggerprototype

##### ADDITIONAL INFORMATION
Currently the only way to rename resources using the Ansible modules is to delete the resource under the old name and recreate it under the new name, despite that the Zabbix API supports renaming these types of resources.
The UID for these resources from the API's point of view is a number, not the name. As that's not very user friendly, when I wrote the modules, I use the resource name and the host/template name to lookup the ID. Unfortunately, the side effect of this is if you try to rename a resource, the module will try to create a duplicate resource with the new name
I've been using the above work around of deleting and recreating, but that has its own unfortunate side-effect that you lose the historical data
This PR adds a new_name parameter like below:

```
- name: Rename Zabbix item
  community.zabbix.zabbix_item:
    name: agent_ping
    template_name: example_template
    params:
      new_name: new_agent_ping
    state: present
```

In this example, the module lookup the itemid for a Zabbix item with the name `agent_ping` in the template `example_template`. It will then set the name field for the item with the id to `new_agent_ping`


aplathrop:
I agree it kind of leaves a bad taste in my mouth, but it seems like we should be able rename an object without deleting it and recreating it, especially because the Zabbix API supports it. I don't rename things every day, but I probably have renamed things a couple dozen times over the last year
I could have it check if object already exists with the new name and if it does, don't try to find the old name, assume it's already been renamed and set the state to ok rather than changed. Not sure this is a better solution, but would make it idempotent

pyrodie18:
> I agree it kind of leaves a bad taste in my mouth, but it seems like we should be able rename an object without deleting it and recreating it, especially because the Zabbix API supports it. I don't rename things every day, but I probably have renamed things a couple dozen times over the last year I could have it check if object already exists with the new name and if it does, don't try to find the old name, assume it's already been renamed and set the state to ok rather than changed. Not sure this is a better solution, but would make it idempotent

It seems like an edge case to me where I would normally say leave it out but I like your idea.  It's similar to how Ansible handles the cmd module.

aplathrop:
Okay, I've changed it so if the new_name is already used, the module will return OK rather than failing and added unit tests to verify that
It says merging is blocked because the branch contains merge commits. I did have to update the branch with the changes to main, but it was showing that message before I did that

pyrodie18 (approved):

PR #1442, opened 2025-01-13
Zabbix 7.2 support

mu1f407:
##### SUMMARY
This PR adds support for Zabbix 7.2. There are some changes in httpapi plugin which simplifies the code. Fixes modules where tests failed.

Fixes #1431

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
All modules (except inventory plugin)

##### ADDITIONAL INFORMATION
Notable code changes in httpapi plugin (beside of 7.2 support):
- Using `self.connection._auth` for request headers. It makes possible to remove `{'auth': 'fake'}` conditions.
- Check for `api_version()` in the `login` method. It is more clear replacement for try-except block in `login`. We now also need to know Zabbix version because it has to be decided if `auth` should be sent in body or `Authorization: Bearer` should be in the header.
- More detailed logging which includes called JSON-RPC method, if used with -vvvv. Example:
  ```
  <127.0.0.1> Zabbix httpapi request: POST http://127.0.0.1:8080/api_jsonrpc.php (hostgroup.get)
  <127.0.0.1> Zabbix httpapi request: POST http://127.0.0.1:8080/api_jsonrpc.php (host.get)
  <127.0.0.1> Zabbix httpapi request: POST http://127.0.0.1:8080/api_jsonrpc.php (maintenance.get)
  ```

PR #1454, opened 2025-02-11
Add Zabbix 7.2 to version validations

### LINKED ISSUE #1452: Unable to install Zabbix server version 7.2
##### SUMMARY

There is an operatring system version validation which does not allow for installation of zabbix 7.2 on any system.
Zabbix 7.2 version is not included in https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_server/vars/Debian.yml

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

zabbix-server role

Ubuntu 22, Zabbix 7.2

##### STEPS TO REPRODUCE
Try to install zabbix 7.2 with zabbix-server role

##### EXPECTED RESULTS

Zabbix server installs

##### ACTUAL RESULTS

Zabbix server fails at validation with:
```
fatal: [censored]: FAILED! => changed=false 
  assertion: '{{ zabbix_server_version|float in zabbix_valid_server_versions[ ansible_facts[''distribution_major_version''] ] }}'
  evaluated_to: false
  msg: Zabbix version 7.2 is not supported on Ubuntu 22
```

Kajot-dev:
##### SUMMARY
Adds Zabbix 7.2 version to validations. While technically #1442 added Zabbix 7.2 support it did not add it to the valid versions effectively preventing it from deployment anywhere.

Fixes #1452 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
All zabbix roles


pyrodie18:
If you're going to do this, you need to add the version to the testing matrix within molecule also so that it will actually test and we can see if it works.  I haven't tried to install 7.2 with the current automation and it may not work correctly (new versions rarely do) the first time.

Kajot-dev:
Will do

BGmot:
@Kajot-dev please check supported OS versions for different components at https://www.zabbix.com/download?zabbix=7.2&os_distribution=debian&os_version=12&components=java_gateway&db=&ws=
You've added some officially unsupported versions in `var/<OS>.yml files and we do not want to test against them.

pyrodie18:
Replacing this PR with #1470

PR #1484, opened 2025-03-10
Agent refactor (part 1)

eb4x:
##### SUMMARY
Refactoring zabbix-agent, these are attempts to split the refactor into smaller digestible chunks.
These are more fixes and cleanups for what's to come.

##### ISSUE TYPE

- Refactor Pull Request

##### COMPONENT NAME
agent


eb4x:
@BGmot Can I get a second opinion on b998ce0816e83a79b29656f0e78892573a5b162e, I fail to see how that code is not garbage. I think we can keep the assertions (to verify a valid configuration), and drop everything else?

pyrodie18 (approved):

eb4x:
@pyrodie18 You got any opinions on `listeninterface` 7442a37caa37bc9571d9c4fc30219c26fb476096? I''ve got a different commit that just removes that code. I don't think we should be writing multiple tasks for trying to detect something the user can set with one line.

pyrodie18:
I'm perfectly fine with refactoring or even removing the checks.  One thing that tends to happen with the collection over time is people will add commits that make it foolproof to run every single time no matter how edge case it is.  Suppose that is not necessarily a bad thing, but makes the code a not bigger, harder to follow, and harder to maintain.  I've always been of the opinion that this collection (and really all of the open source ones) should be designed to give the ~80% solution and that the people using it should be expected to understand their environment, configure their roles/playbooks properly, and if/when they fail for whatever reason (i.e. using a port that doesn't exist) that they are able to respond appropriately.

eb4x:
Thanks, I think you're right, and I'm gonna use my best judgement here and strip out that code. It's not something the role should be solving.

PR #1489, opened 2025-03-28
Restrict access to the frontend config

baszoetekouw:
##### SUMMARY
Restrict access to the web frontend config file `/etc/zabbix/web/zabbix.conf.php`.
The file contains passwords, so should not be publicly readable. Also fix the owner for the case that the fpm user differs from the www user.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_web

##### ADDITIONAL INFORMATION
Currently, the config file gets these permissions:
```
-rw-r--r--  1 www-data www-data 756 Mar 27 20:38 /etc/zabbix/web/zabbix.conf.php
```
This is undesired, as the file contains the database password.  Also, the owner/group are wrong in the case that the `zabbix_php_fpm_conf_user` differs from `zabbix_web_user`.

This patch restricts the (default) permissions to 0640, changes the owner to `zabbix_web_user` (or `zabbix_php_fpm_conf_user` if it is defined) and the group to `zabbix_web_group` (or `zabbix_php_fpm_conf_group` if it is defined).

### DIFF
--- a/roles/zabbix_web/defaults/main.yml
+++ b/roles/zabbix_web/defaults/main.yml
@@ -4,7 +4,7 @@
 # zabbix_web_version: 6.4
 zabbix_web_package_state: present
 zabbix_web_doubleprecision: false
-zabbix_web_conf_mode: "0644"
+zabbix_web_conf_mode: "0640"
 zabbix_web_connect_ha_backend: false
 zabbix_api_server_url: zabbix.example.com
 zabbix_web_http_server: apache
--- a/roles/zabbix_web/tasks/main.yml
+++ b/roles/zabbix_web/tasks/main.yml
@@ -108,8 +108,8 @@
   ansible.builtin.template:
     src: zabbix.conf.php.j2
     dest: /etc/zabbix/web/zabbix.conf.php
-    owner: "{{ zabbix_web_user }}"
-    group: "{{ zabbix_web_group }}"
+    owner: "{{ zabbix_php_fpm_conf_user if zabbix_php_fpm_conf_user is defined else zabbix_web_user }}"
+    group: "{{ zabbix_php_fpm_conf_group if zabbix_php_fpm_conf_group is defined else zabbix_web_group }}"
     mode: "{{ zabbix_web_conf_mode }}"
   become: true
   tags:


eb4x:
Good catch on the permissions!

This change is so good, I think you should simplify their names and promote those variables to `defaults/main.yml`. It's a very common use-case to have different php users, esp. in shared environments. (or at least it was a common use-case, probably less so with virt and containers these days, but still...)

```yaml
zabbix_web_php_user: "{{ zabbix_web_user }}"
zabbix_web_php_group: "{{ zabbix_web_group }}"
```

And also if you could just add those vars to the `docs/ZABBIX_WEB_ROLE.md` with short explanation for their use case pretty please :smile_cat:

eb4x:
Apparently I missed the part where we actually had a `zabbix_php_fpm_conf_*`, so just promote those to `defaults/main.yml`

BGmot:
Hi @baszoetekouw, are you going to proceed with this?

baszoetekouw:
Yes, it's on my todo list, which, unfortunately is quite long.  But I'll get around to it!

BGmot:
> Apparently I missed the part where we actually had a `zabbix_php_fpm_conf_*`, so just promote those to `defaults/main.yml` (and set the values to the corresponding `zabbix_web_{user,group}`)

I am not sure I follow.
If `zabbix_web_user` is not defined then we set it to `_nginx_user` or `_apache_user` that in turn depend on Linux distro. How do you see it in defaults/main.yml?

eb4x:
In `zabbix_web/templates/php-fpm.conf.j2` we have these two overrides in the form of;
```
user = {{ zabbix_php_fpm_conf_user if zabbix_php_fpm_conf_user is defined else zabbix_web_user }}
group = {{ zabbix_php_fpm_conf_group if zabbix_php_fpm_conf_group is defined else zabbix_web_group }}
```
Which is pretty much the same as setting `zabbix_php_fpm_conf_user: "{{ zabbix_web_user }}"` in `defaults/main.yml`.

I had a look at the docs, and the overrides are actually documented, I just look at `defaults/main.yml` as the place we declare variables for deployers to override. (And them not being there led me to believe they were undocumented.) That's what I meant by "promote them to defaults/main.yml"

BGmot:
> In `zabbix_web/templates/php-fpm.conf.j2` we have these two overrides in the form of;
> 
> ```
> user = {{ zabbix_php_fpm_conf_user if zabbix_php_fpm_conf_user is defined else zabbix_web_user }}
> group = {{ zabbix_php_fpm_conf_group if zabbix_php_fpm_conf_group is defined else zabbix_web_group }}
> ```
> 
> Which is pretty much the same as setting `zabbix_php_fpm_conf_user: "{{ zabbix_web_user }}"` in `defaults/main.yml`.
> 
> I had a look at the docs, and the overrides are actually documented, I just look at `defaults/main.yml` as the place we declare variables for deployers to override. (And them not being there led me to believe they were undocumented.) That's what I meant by "promote them to defaults/main.yml"

@eb4x  like here https://github.com/ansible-collections/community.zabbix/pull/1529 ?

BGmot:
Dropped in favour of https://github.com/ansible-collections/community.zabbix/pull/1529
Thanks @baszoetekouw !

Issue #1416, opened 2024-11-04
Add support for Alpine Linux in Agent Role


EckisWelt:
##### SUMMARY
Please add Alpine Linux to the supported list of operating systems for Agent/Agent2 deployment. This feature would allow seamless integration and automation of Agent/Agent2 installation on machines running Alpine Linux, especially in LXC environments.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
zabbix_agent

##### ADDITIONAL INFORMATION
As a user with a large infrastructure of machines running Alpine Linux within LXC containers, I currently rely on manual installation of Agent2 using the Testing repository. The addition of Alpine Linux to the supported OS list would greatly streamline my workflow, enabling automated and more efficient deployments.

You already have a lot of knowledge with Alpine, as you have ready made deployment scripts for Docker and Docker Compose, this shouldn't take a lot of resources. There are even blogposts from 2021 explaining how to setup agents on Alpine.

eb4x:
Following up on this, how would you like to see zabbix-agent installed on alpine? There doesn't seem to be official builds from Zabbix, so would you just install the package from distro with apk, or run in a container?

EckisWelt:
I would prefer apk. Adding it in docker containers is already supported.

eb4x:
#1490 was just merged, it should allow for basic installation of zabbix-agent(2) on alpine.

PR #1490, opened 2025-03-28
Alpine

eb4x:
##### SUMMARY
Implement #1416 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

roles/zabbix_agent

##### ADDITIONAL INFORMATION
So there's no official repo for alpine, and from my limited understanding no way to install a specific version with apk, so you're pretty much left at the mercy of whatever was packaged at the time of the alpine release. But if you just need the agent and some config, this should get the job done.

I'd like add alpine to testing, but that would require a not insignificant overhaul molecule, where we rely less on geerlingguy images.


BGmot:
I don't see Alpine on the OS list on Zabbix download page so I don't think we need to add molecule test. I am very skeptical in general about this feature request. The requester talks about containers and there are official alpine containers why would we install it using apk?

eb4x:
> The requester talks about containers and there are official alpine containers why would we install it using apk?

Because these are LXC containers, it's basically vm-responsibility-in-a-container, where you're kinda stuck doing package maintenance/upgrades in the container instead of swapping it out with a new container.

This would also cover the case where someone installs alpine directly to hardware (via iso,pxe, whatever)

I'm not to concerned with testing or maintaining support on more then a best-effort basis. But I do like that it remains a possibility.

BGmot:
Ok, sorry I am not that familiar with LXC containers. I don't see a problem with this PR but in regards to extensive testing I would not bother.
Thanks!

BGmot (approved):

PR #1496, opened 2025-04-01
Fix status of zabbix token

manuello1996:
##### SUMMARY

For creation of Zabbix Auth Token via Ansible the option status should create an Enabled or Disabled token. 
Aktuelly the task with Status:true will create a disabled token. 
Accordingly to the documentation the value are inverted. 
https://www.zabbix.com/documentation/7.0/en/manual/api/reference/token/object

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_token.py 

##### ADDITIONAL INFORMATION

```paste below
[...]
community.zabbix.zabbix_token:
    name: Zabbix CI Token
    description: "{{ zabbix_hvrz_api_ansible_user }} auth token"
    username: "{{ zabbix_hvrz_api_ansible_user }}"
    status: true
    generate_token: true
    state: present
```
Will generate a disable token. 
With option  status: false the token will be enabled.

### DIFF
--- a/changelogs/fragments/zabbix_token_module_inverted_value.yaml
+++ b/changelogs/fragments/zabbix_token_module_inverted_value.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - zabbix_token module - Fix status value for zabbix Auth token. 
--- a/plugins/modules/zabbix_token.py
+++ b/plugins/modules/zabbix_token.py
@@ -194,11 +194,11 @@ class Token(ZabbixBase):
 
             if isinstance(status, bool):
                 if status:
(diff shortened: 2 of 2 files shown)


pyrodie18 (changes requested):
Nice catch!  We'll let the tests run through and check (although I don't see why the won't pass).  Only comment is to please create a change fragment (https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments).  I expect this would be a bugfix.

Thanks

manuello1996 (commented):
Looks good to me

pyrodie18 (approved):

PR #1521, opened 2025-05-06
[Suggestion] Change ansible-core version of role tests

masa-orca:
##### SUMMARY

Change ansible-core version of role tests

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

all roles
##### ADDITIONAL INFORMATION

- currently, we are test roles with ansible-core 2.16.
- The version is older, therefore, I'd like to use latest ansible-core if there is no specific reason.
  - ex) ansible-core < 2.17 can run playbook against to RHEL 8 famiy. 

```paste below

```

### DIFF
--- a/.github/workflows/agent.yml
+++ b/.github/workflows/agent.yml
@@ -26,14 +26,16 @@ jobs:
       fail-fast: false
       matrix:
         container:
-          - rockylinux9
-          - rockylinux8
-          - ubuntu2404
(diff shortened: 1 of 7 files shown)


eb4x (approved):
Nice way getting round the issue :+1: 

It was a sad day when I discovered I was gonna be stuck with 2.16 until 2030.

PR #1529, opened 2025-05-14
Fix zabbix.conf.php permissions

BGmot:
##### SUMMARY
Replacement for https://github.com/ansible-collections/community.zabbix/pull/1489

Restrict access to the web frontend config file /etc/zabbix/web/zabbix.conf.php.
The file contains passwords, so should not be publicly readable. Also fix the owner for the case that the fpm user differs from the www user.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_web role


eb4x (approved):
Perfect! :+1:

eb4x:
You could also simplify the check in `zabbix_web/templates/php-fpm.conf.j2` to just use the php value, since it's always provided by defaults/main.yml.

Issue #1531, opened 2025-05-15
Remove Support for Retired Ansible


pyrodie18:
Remove support for Ansible 2.16

pyrodie18:
Keeping while we support RHEL 9

PR #1532, opened 2025-05-15
Remove Support for Ansible 2.16 and 2.17

pyrodie18:
##### SUMMARY
Ref #1531

### DIFF
--- a/.github/workflows/plugins-integration.yml
+++ b/.github/workflows/plugins-integration.yml
@@ -25,8 +25,6 @@ jobs:
           - version: "7.2"
         ansible:
           # https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-changelogs
-          - stable-2.16
-          - stable-2.17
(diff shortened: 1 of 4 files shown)


masa-orca:
In addition, I think we need to delete tests of all roles for rockylinux 8 and opensuseleep15
https://github.com/ansible-collections/community.zabbix/blob/main/.github/workflows/agent.yml#L30 is one of them

masa-orca (changes requested):
Reason: https://github.com/ansible-collections/community.zabbix/pull/1532#issuecomment-2883801339

masa-orca:
> I'm not entirely sure why we're specifying the ansible version to install for the agent tests anyway. 

We are testing with only ansible-core 2,16 against EL8 at all roles.

Rocky is one of the Redhat family. It is obvious that dropping RHEL8 means dropping Redhat family's major version 8, 

You says Rock's life cycle, but it is not ansible support life.

PR #1537, opened 2025-05-15
Sane Selinux defaults

eb4x:
##### SUMMARY
Deal with selinux variables we know something about so this just works out of the box.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
roles for server, proxy and web

##### ADDITIONAL INFORMATION
This alters the defaults, but for good reason, and the people deploying are probably already setting these, so there shouldn't be any surprises.

### DIFF
--- a/changelogs/fragments/1537_selinux.yml
+++ b/changelogs/fragments/1537_selinux.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - roles - sane selinux defaults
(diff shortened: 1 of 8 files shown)


pyrodie18 (commented):
Add a change fragment and then I think you're good.

pyrodie18 (approved):

PR #1572, opened 2025-06-25
Added tasks for correct reinstallation for windows

virys99:
##### SUMMARY

Added tasks for correct reinstallation for windows 
Added task for removing download file after installation

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

install-Windows.yml
##### ADDITIONAL INFORMATION

I apologize that the code is not very neat. But it solves the problem if the old version of the agent is already installed on the system and needs to be reinstalled.

```paste below

```

### DIFF
--- a/roles/zabbix_agent/tasks/install-Windows.yml
+++ b/roles/zabbix_agent/tasks/install-Windows.yml
@@ -1,5 +1,7 @@
 ---
 - name: Get latest zabbix releases
+  vars:
+    ansible_connection: local
   delegate_to: localhost
   uri:
     url: https://services.zabbix.com/updates/v1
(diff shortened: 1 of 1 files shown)


eb4x:
Thanks, it's a bit messy indeed, but I think we can work from this. I'll have to look more into this over the weekend and next week to give a proper review, and test the code.

First glance it looks like this code assumes you'll only ever want to switch to zabbix-agent2 and other way?

roles/zabbix_agent/tasks/install-Windows.yml
```diff
@@ -13,33 +15,140 @@
     _latest_release: "versions[?version=='{{ zabbix_agent_version }}'].latest_release.release | [0]"
   when: zabbix_agent_version_long is undefined
 
-- name: Check for and uninstall old agent
-  when: work_not_in_progress_anymore | default(false)
+- name: Autodetect install zabbix agent version
+  set_fact:
+    zabbix_agent2: false
+  when: ansible_distribution_major_version | int < 10
```
eb4x, line 21: I don't think that reflects which windows version is actually running?

```
    "distribution": "Microsoft Windows Server 2022 Standard Evaluation",
    "distribution_major_version": "10",
    "distribution_version": "10.0.20348.0",
```

Issue #1585, opened 2025-07-11
Add Raspberry Pi OS Support


patrik-plastik:
##### SUMMARY
Add Raspberry Pi OS Support

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
roles/zabbix_repo

##### ADDITIONAL INFORMATION
Currently, I am unable to install zabbix agent on my Raspberry Pi, because it have different architecture
```
fatal: [host]: FAILED! => {
    "msg": "The task includes an option with an undefined variable.. {{ _zabbix_repo_deb_arch_map[ansible_facts['architecture']] }}: 'dict object' has no attribute 'armv7l'

The error appears to be in '/*/.ansible/collections/ansible_collections/community/zabbix/roles/zabbix_repo/tasks/Debian.yml': line 47, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: \"Debian | Installing repository {{ ansible_facts['distribution'] }}\"
  ^ here
We could be wrong, but this one looks like it might be an issue with
missing quotes. Always quote template expression brackets when they
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - \"{{ foo }}\"
"
}
```
I have tried to add the armv7l into the list of all architecture here: ```/roles/zabbix_repo/vars/Debian.yml``` but it ends with error:
```
failed: [host] (item={'name': 'zabbix-sender', 'state': 'present'}) => {
    "ansible_loop_var": "item",
    "attempts": 3,
    "cache_update_time": 1752240687,
    "cache_updated": false,
    "changed": false,
    "invocation": {
        "module_args": {
            "allow_change_held_packages": false,
            "allow_downgrade": false,
            "allow_unauthenticated": false,
            "autoclean": false,
            "autoremove": false,
            "cache_valid_time": 0,
            "clean": false,
            "deb": null,
            "default_release": null,
            "dpkg_options": "force-confdef,force-confold",
            "fail_on_autoremove": false,
            "force": false,
            "force_apt_get": false,
            "install_recommends": false,
            "lock_timeout": 60,
            "name": "zabbix-sender",
            "only_upgrade": false,
            "package": [
                "zabbix-sender"
            ],
            "policy_rc_d": null,
            "purge": false,
            "state": "present",
            "update_cache": true,
            "update_cache_retries": 5,
            "update_cache_retry_max_delay": 12,
            "upgrade": null
        }
    },
[...]

PR #1591, opened 2025-07-17
FEATURE: add support for Raspbian OS in zabbix_repo role

patrik-plastik:
##### SUMMARY
Adding support of Rasberry Pi OS for zabbix_repo role

#1585

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
roles/zabbix_repo

##### ADDITIONAL INFORMATION
I've added support for Raspbian devices and decided to go with different tasks in case of any difference in the OS in (unknown) future.
The 64 bit versions of Raspbian devices works with Debian line perfectly fine, but 32 bits needs different components to be downloaded.


patrik-plastik:
The error wasn't caused by my commit I believe.

This is from log of failed check:
```
2025-07-17T16:00:09.9788748Z TASK [community.zabbix.zabbix_repo : RedHat | Install GPG key] *****************
2025-07-17T16:00:11.0178791Z fatal: [zabbix-proxy-v72-sqlite3-rockylinux9]: FAILED! => {"changed": false, "msg": "failed to fetch key at http://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-08EFA7DD , error was: Connection failure: [Errno 104] Connection reset by peer"}
```
(idk if I am able to re-run specific check but I guess not)

eb4x:
Do you happen to know what this achieves/changes compared to the debian code? I have a feeling this is excessive, and the easy fix is probably 3-4 lines.

I'll spin up a raspi2b under qemu and see if I can figure out what the problem is. The problem is with armhf right?

patrik-plastik:
I thought that it would be better to have it split.
I thought the same and I can "merge" it into one if you would like it this way. I wasn't sure tho.
When I was testing it and playing with it, I realised this change over 4 lines, yet from the original ticket it sounded like it "should" have its own branch if you understand me.

PR #1618, opened 2025-08-21
Add option to turn off installing dependencies AND database client in zabbix-server role

patrik-plastik:
##### SUMMARY
Adding a feature to turn off installing dependencies and database client. This options is focused for when someone use:
```yaml
- ansible.builtin.include_role:
    name: community.zabbix.zabbix_server
    tasks_from: initialize-pgsql.yml
```

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
role/zabbix_server


pyrodie18:
Is there a reason why you can't just exclude the "install" tags tasks?

patrik-plastik:
I believe this gives me more options when calling the tasks from yaml. Yes, I can use them, I don't think that this is as much of a hustle to add this feature.

eb4x:
It's small and at the surface seems reasonable enough, but there are only two items in the dependencies. The "soft" dependency on a database client, and the actual dependency for using community.postgresl (python3-psychopg2) and community.mysql (python3-pymysql) modules.

So while I am inclined to just approve, I'd like to know why you'd want to avoid these. Is it a time-saving issue? (I know the package task can waste a lot of time)

patrik-plastik:
I have my own process to deal with packages on Linux machines and I install the package earlier. I just want to use the zabbix initialisation and timescaleDB. Not packages nor creating databases.
So for me it's a waste of time and I don't want 2 tasks to do the same and if something is wrong, I have to remember that this task in on two places. I hope you understand what I mean.

PR #1635, opened 2025-09-11
Fix zabbix_valuemap module

gjdoornink:
##### SUMMARY

This PR fixes two issues in the existing implementation:
- Add the host_name parameter since the host id is a required parameter when creating value maps.
- Filter out existing value maps that have an uuid value since these value maps are created as result of an import.

This PR fixes issue #1636 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_valuemap

##### ADDITIONAL INFORMATION

In construct_parameters the value mappings "type" property value is converted to a string since the valuemap.get Zabbix api returns the value mappings type property as a string even though the docs state that the value mappings type property has an integer type.


pyrodie18:
> I see that codecov/project is failing but I do not know what that is about.

Don't worry about it.  Looks good, I'll pull it in.  Thanks for the PR

PR #1641, opened 2025-09-22
request trigger dependencies and tags for comparison

vr-jack:
Change comparison fails for dependency and tag changes if we have not selected them.

##### SUMMARY

When setting dependencies or tags, comparison cannot be made to determine change status if trigger.get hasn't requested the data.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_trigger

##### ADDITIONAL INFORMATION

Sorry for the random PR drop, but it was a simple modification. The only other data that MIGHT be of use was functions, as I don't know if it's possible to change the function without the ID in the expression (which does get compared) changing. I have not tested function parameter changes to see if they would get missed. It's only extra data, so it might not be a bad idea to add selectFunctions: extend.

All 3 selects do show up in Zabbix 3.4 api specs. I didn't go further back than that.

```paste below

```


pyrodie18 (changes requested):
Looks good, please add a change fragment and we'll get you pulled in.

vr-jack:
I think I did that correctly. Used existing file as template.

PR #1652, opened 2025-10-23
add variable for releases url

MatthieuLeboeuf:
##### SUMMARY
Add Variable for customizing https://services.zabbix.com/updates/v1
Added 304 status code for nexus proxy "not modified"

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_agent


MatthieuLeboeuf:
Nexus is a proxy that allows caching web pages without complicating their operation.

The only thing to do is provide the original URL, and then I get a new URL that no longer has the rate limit of the original resource.

For example: https://services.zabbix.com/updates/v1

The problem is that the HTTP return code is not the same.

PR #1702, opened 2026-04-05
Add zabbix-web-service role (2nd attempt)

### LINKED ISSUE #1653: Confused about the zabbix-web-service (scheduled reports) implementation status in collection
The scheduled reports functionality: https://www.zabbix.com/documentation/current/en/manual/config/reports

In the server role https://github.com/ansible-collections/community.zabbix/tree/main/roles/zabbix_server , I've found the "zabbix_server_startreportwriters" a "zabbix_server_webserviceurl" options for use.

However, in the web role (frontend) https://github.com/ansible-collections/community.zabbix/tree/main/roles/zabbix_web (or any other role), I'm unable to find anything, that would be installing or using the zabbix-web-service package.
I realize, that the zabbix-web-service also needs the chrome/chromium for it's functionality, nevertheless, I'm confused about the current level of zabbix-web-service implementation in the community.zabbix collection.

Did I perhaps missed something?
Is the implementation perhaps incomplete?
Or because of the chrome/chromium requirement, was it perhaps never intended to continue with it's implementation, left for the end user to handle it?
Or ... ?

dv8472:
##### SUMMARY
The project is missing the Zabbix Web service: https://www.zabbix.com/documentation/current/en/manual/concepts/web_service
Closes https://github.com/ansible-collections/community.zabbix/issues/1653

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix-web-service

##### ADDITIONAL INFORMATION
Previous https://github.com/ansible-collections/community.zabbix/pull/1699 was somehow closed with unknown reason to me.
@eb4x your comments from the https://github.com/ansible-collections/community.zabbix/pull/1699 should be fixed in this commit.


roles/zabbix_web_service/vars/RedHat.yml
```diff
@@ -0,0 +1,23 @@
+---
+__epel_repo:
+  - epel
+
+zabbix_valid_web_service_versions:
+  "10":
+    - 7.4
```
eb4x, line 7: There are EL10 packages for 7.0 aswell, and you can omit references to 7.2, it's been EOL'd by Zabbix

roles/zabbix_web_service/tasks/main.yml
```diff
@@ -0,0 +1,119 @@
...
+    recurse: true
+  notify:
+    - zabbix-web-service restarted
+  changed_when: false
+
+- name: Google-chrome requirement - Create mimeapps.list file
+  become: true
+  ansible.builtin.file:
+    path: "{{ mimeapps_list_path }}/mimeapps.list"
+    state: touch
```
eb4x: `state: touch` is not idempotent. You could do something like
```yaml
- command:
    cmd: "touch {{ mimeapps_list_path }}/mimeapps.list"
    creates: "{{ mimeapps_list_path }}/mimeapps.list"
```
I suppose.

I don't know if chrome would potentially modify this file, otherwise some known configuration would be preferable, like
```yaml
- copy:
    content: |
      our known configuration
      of mimetypes
    dest: "{{ mimeapps_list_path }}/mimeapps.list"
```
But yeah if the content isn't important, a `command:` with `creates:` is fine.

roles/zabbix_web_service/tasks/main.yml
```diff
@@ -0,0 +1,119 @@
...
+    owner: "{{ zabbix_os_user }}"
+    group: "{{ zabbix_os_user }}"
+    mode: '0644'
+  notify:
+    - zabbix-web-service restarted
+  changed_when: false
```
eb4x: Oh, I see what you did here.

I don't think we should hide changes like this, because the first run it should be creating a file, which is an actual change, that's now being supressed.

eb4x (changes requested):
Overall looks good 👍 

Some small nits, just remove 7.2, it's been EOL'd by Zabbix already, so not something we want to introduce. We just haven't gotten around to removing 7.2 from the other roles yet.

And the `touch`, if you can't get the correct owner/permissions on it via `become_user` or some other means, make it two tasks, one to create and one to adjust permissions.

roles/zabbix_web_service/tasks/main.yml
dv8472: got it, fixed now

dv8472:
> Overall looks good 👍
> 
> Some small nits, just remove 7.2, it's been EOL'd by Zabbix already, so not something we want to introduce. We just haven't gotten around to removing 7.2 from the other roles yet.
> 
> And the `touch`, if you can't get the correct owner/permissions on it via `become_user` or some other means, make it two tasks, one to create and one to adjust permissions.

fixes uploaded

eb4x:
Thank you for your contribution :heart:

PR #1714, opened 2026-04-16
For upstream server

eb4x:
##### SUMMARY
Refactoring the version check, and some minor cleanup of unused variables in the zabbix_server role.

##### ISSUE TYPE

- Refactor Pull Request

##### COMPONENT NAME
zabbx_server role

##### ADDITIONAL INFORMATION
I want to change some defaults and rename defaults so there's a bunch of extra tasks in deprecation_checks.yml that attempt to give warnings and not break things for existing installations. And use the corrected defaults for new installs.

### DIFF
--- a/molecule/zabbix_server/tests/test_default.py
+++ b/molecule/zabbix_server/tests/test_default.py
@@ -40,7 +40,7 @@ def test_zabbix_server_dot_conf(host):
 
 
 def test_zabbix_include_dir(host):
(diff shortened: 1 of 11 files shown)


eb4x:
These are good to go, but I'd like a second opinion on the last two commits, the goal is to change defaults and warn users of existing installations about those changes and whether they mitigate them or not. I'm wondering if we should add a `- pause:` in the rescue section just so the warnings stick out more.

(Don't squash-merge this PR)

pyrodie18 (approved):

Issue #1719, opened 2026-04-17
template nginx_vhost.conf.j2 in zabbix_web


ivomitev79-netizen:
##### SUMMARY
The generated zabbix.conf has wrong syntax.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
role zabbix_web

##### ANSIBLE VERSION
Ansible version doesn'e make sense, because of templated file problem

##### CONFIGURATION
Wrong one
```    
    listen {{ zabbix_web_vhost_tls_port }} ssl;
    {{ 'http2 on;' if zabbix_web_ssl_http2 else '' }}
    listen [::]:{{ zabbix_web_vhost_tls_port }} ssl
    {{ 'http2 on;' if zabbix_web_ssl_http2 else '' }};

```
Must be
```    
    listen {{ zabbix_web_vhost_tls_port }} ssl;
    listen [::]:{{ zabbix_web_vhost_tls_port }} ssl;
    {{ 'http2 on;' if zabbix_web_ssl_http2 else '' }}

```

ivomitev79-netizen:
If multiple alias, they are concatenated without space between alias domains
Another wrong line: 
```
server_name {{ zabbix_api_server_url }} {% for alias in zabbix_url_aliases -%}{{ alias -}} {% endfor %};
```
must be
```
server_name {{ zabbix_api_server_url }}{%- for alias in zabbix_url_aliases %} {{ alias }}{%- endfor %};
```

eb4x:
@manuello1996 Can you come with a new PR rectifying the issues caused by #1706

eb4x:
Fixed in upstream with #1721

PR #1721, opened 2026-04-18
nginx vhost template for RHEL10

manuello1996:
##### SUMMARY

As requested in #1719 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_web

##### ADDITIONAL INFORMATION

Given the fact that at the moment the only OS presenting the problem with the actual nginx conf is RHL/EL 10+, i added a conditional check 

```paste below

```

### DIFF
--- a/roles/zabbix_web/defaults/main.yml
+++ b/roles/zabbix_web/defaults/main.yml
@@ -23,6 +23,7 @@ zabbix_web_vhost_tls_port: 443
 zabbix_web_vhost_listen_ip: "*"
 zabbix_web_tls: false
 zabbix_web_tls_http2: false
+zabbix_web_tls_http2: "{{ zabbix_web_ssl_http2 | default(false) }}"
 zabbix_timezone: Europe/Amsterdam
 zabbix_php_fpm_conf_listen: true
 # zabbix_web_tls_crt: /etc/pki/server.crt
--- a/roles/zabbix_web/defaults/main.yml
+++ b/roles/zabbix_web/defaults/main.yml
@@ -22,7 +22,6 @@ zabbix_web_vhost_port: 80
 zabbix_web_vhost_tls_port: 443
 zabbix_web_vhost_listen_ip: "*"
 zabbix_web_tls: false
-zabbix_web_tls_http2: false
 zabbix_web_tls_http2: "{{ zabbix_web_ssl_http2 | default(false) }}"
 zabbix_timezone: Europe/Amsterdam
 zabbix_php_fpm_conf_listen: true
--- a/roles/zabbix_web/defaults/main.yml
+++ b/roles/zabbix_web/defaults/main.yml
@@ -22,7 +22,7 @@ zabbix_web_vhost_port: 80
 zabbix_web_vhost_tls_port: 443
 zabbix_web_vhost_listen_ip: "*"
 zabbix_web_tls: false
-zabbix_web_ssl_http2: false
+zabbix_web_tls_http2: false
 zabbix_timezone: Europe/Amsterdam
 zabbix_php_fpm_conf_listen: true
 # zabbix_web_tls_crt: /etc/pki/server.crt
--- a/roles/zabbix_web/defaults/main.yml
+++ b/roles/zabbix_web/defaults/main.yml
@@ -22,6 +22,7 @@ zabbix_web_vhost_port: 80
 zabbix_web_vhost_tls_port: 443
 zabbix_web_vhost_listen_ip: "*"
 zabbix_web_tls: false
+zabbix_web_ssl_http2: false
 zabbix_timezone: Europe/Amsterdam
 zabbix_php_fpm_conf_listen: true
 # zabbix_web_tls_crt: /etc/pki/server.crt
--- a/roles/zabbix_web/tasks/nginx.yml
+++ b/roles/zabbix_web/tasks/nginx.yml
@@ -47,7 +47,7 @@
         - nginx-light
       when:
         - ansible_facts.packages[item] is defined
-        - (zabbix_web_nginx_installed_version | default('')) == ''
+        - zabbix_web_nginx_installed_version is not defined
 
 - name: "Nginx | Install vhost in conf.d"
   ansible.builtin.template:
--- a/roles/zabbix_web/tasks/nginx.yml
+++ b/roles/zabbix_web/tasks/nginx.yml
@@ -25,54 +25,29 @@
   tags:
     - install
 
-- name: "Nginx | Gather package facts for HTTP/2 syntax detection"
-  ansible.builtin.package_facts:
-    manager: auto
-  become: true
+- name: "Nginx | Determining Nginx version"
   when:
     - zabbix_web_create_vhost
     - zabbix_web_tls
-    - zabbix_web_tls_http2
   tags:
     - config
+  block:
+    - name: "Nginx | Gather package facts"
+      ansible.builtin.package_facts:
+        manager: auto
+      become: true
 
-- name: "Nginx | Default HTTP/2 syntax values"
-  ansible.builtin.set_fact:
-    zabbix_web_http2_directive: ""
-    zabbix_web_http2_listen_option: ""
-  when:
-    - zabbix_web_create_vhost
-    - zabbix_web_tls
-  tags:
-    - config
-
-- name: "Nginx | Detect installed Nginx version"
-  ansible.builtin.set_fact:
-    zabbix_web_nginx_installed_version: "{{ ((ansible_facts.packages[item] | first).version | split(':') | last | split('-') | first) | default('0.0.0', true) }}"
-  loop:
-    - nginx
-    - nginx-core
-    - nginx-full
-    - nginx-light
-  when:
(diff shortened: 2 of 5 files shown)


roles/zabbix_web/templates/nginx_vhost.conf.j2
```diff
@@ -75,14 +75,14 @@ server {
 }
 
 {% if zabbix_web_tls|default(false) %}
+{% set _zabbix_web_use_http2_directive = zabbix_web_ssl_http2 and ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version'] is version('10', '>=') %}
+{% set _zabbix_web_http2_listen_parameter = ' http2' if zabbix_web_ssl_http2 and not _zabbix_web_use_http2_directive else '' %}
```
eb4x: I'm not a fan of this.

The nginx template should not have knowledge of which operating systems it's deployed on. It should only have knowledge on which version of nginx it's deployed to.

So put this logic outside of the template and have the template react to an appropriately named flag.

Also, it would be preferable to query nginx for its version instead of relying on what we currently know about nginx releases on os'es. What happens when 1.30 suddenly becomes available on Suse-15?

roles/zabbix_web/defaults/main.yml
```diff
@@ -22,6 +22,7 @@ zabbix_web_vhost_port: 80
 zabbix_web_vhost_tls_port: 443
 zabbix_web_vhost_listen_ip: "*"
 zabbix_web_tls: false
+zabbix_web_ssl_http2: false
```
eb4x: It's put in the right spot, but since we're already calling it `zabbix_web_tls`, let's not make it more confusing with `ssl` in the mix.

`zabbix_web_ssl_http2 => zabbix_web_tls_http2`

eb4x (changes requested):

roles/zabbix_web/tasks/nginx.yml
```diff
@@ -25,6 +25,55 @@
...
+- name: "Nginx | Gather package facts for HTTP/2 syntax detection"
+  ansible.builtin.package_facts:
+    manager: auto
+  become: true
+  when:
+    - zabbix_web_create_vhost
+    - zabbix_web_tls
+    - zabbix_web_tls_http2
+  tags:
+    - config
```
eb4x: This is actually really clever 😺 !

roles/zabbix_web/templates/nginx_vhost.conf.j2
```diff
@@ -76,13 +76,11 @@ server {
 
 {% if zabbix_web_tls|default(false) %}
 server {
-    listen {{ zabbix_web_vhost_tls_port }} ssl;
-    {{ 'http2 on;' if zabbix_web_ssl_http2 else '' }}
-    listen [::]:{{ zabbix_web_vhost_tls_port }} ssl
-    {{ 'http2 on;' if zabbix_web_ssl_http2 else '' }};
+    listen {{ zabbix_web_vhost_tls_port }} ssl{{ zabbix_web_http2_listen_option | default('') }};
+    listen [::]:{{ zabbix_web_vhost_tls_port }} ssl{{ zabbix_web_http2_listen_option | default('') }};
+    {{ zabbix_web_http2_directive | default('') }}
```
eb4x: you're close, but maybe to complicated? I thinking more something like;
```
{% if _nginx_version is version('1.25.1', '>=') %}
listen ...;
http2 {{ zabbix_web_tls_http2 | ternary('on', 'off') }};
{% else %}
listen ... http2; #FIXME
{% endif %}
```
Making `zabbix_web_tls_http2` adds some complexity since it was always on before, but yeah you fix it we merge it 👍

roles/zabbix_web/tasks/nginx.yml
```diff
@@ -25,6 +25,55 @@
...
+- name: "Nginx | Determine HTTP/2 syntax flags"
+  ansible.builtin.set_fact:
+    zabbix_web_http2_directive: "{{ 'http2 on;' if (zabbix_web_nginx_installed_version | default('0.0.0')) is version('1.25.1', '>=') else '' }}"
+    zabbix_web_http2_listen_option: "{{ ' http2' if (zabbix_web_nginx_installed_version | default('0.0.0')) is version('1.25.1', '<') else '' }}"
+  when:
+    - zabbix_web_create_vhost
+    - zabbix_web_tls
+    - zabbix_web_tls_http2
+  tags:
+    - config
```
eb4x: Instead of this, we'd attach a
```yaml
vars:
  _nginx_version: "{{ zabbix_web_nginx_installed_version }}"
```
to the vhost template task.

roles/zabbix_web/tasks/nginx.yml
```diff
@@ -25,6 +25,55 @@
...
+    - nginx-full
+    - nginx-light
+  when:
+    - zabbix_web_create_vhost
+    - zabbix_web_tls
+    - zabbix_web_tls_http2
+    - ansible_facts.packages[item] is defined
+    - (zabbix_web_nginx_installed_version | default('')) == ''
+  tags:
+    - config
```
eb4x: I'm assuming this works, but it's a bit messy. Too many defaults.

We're looking for a package in the sequence specified in the loop, and it will stop (breaks out of the loop) once it detects a package and sets `zabbix_web_nginx_installed_version` because we've added it to the when condition.

```yaml
- set_fact:
    zabbix_web_nginx_installed_version: "{{ ansible_facts.packages[item] | first).version }}"
  loop:
    - nginx
    - ...
  when:
    - zabbix_web_nginx_installed_version is not defined # funny gotcha, exits the loop when set
```
I wasn't sure why you were doing splits on `:` and `-`, but I see debian reports back
`1.22.1-9+deb12u4`, so I think the cleanest is probably a `regex_replace` on this.

And maybe move all the logic under a `block:` so we can put the common when conditions there, and just the relevant ones on individual tasks.

```yaml
- name: Logic for deremining nginx version
  when:
    - zabbix_web_create_vhost
    - zabbix_web_tls
  block:
    - package_facts:
    - set_fact:
```

eb4x (changes requested):
Good effort, I really liked the use of `package_facts:` 👍

roles/zabbix_web/templates/nginx_vhost.conf.j2
manuello1996: it's a good idea, it make the template more readable

roles/zabbix_web/tasks/nginx.yml
manuello1996: I implemented the suggested change

roles/zabbix_web/tasks/nginx.yml
```diff
@@ -25,13 +25,39 @@
...
+  block:
+    - name: "Nginx | Gather package facts"
+      ansible.builtin.package_facts:
+        manager: auto
+      become: true
+
+    - name: "Nginx | Parse installed package version"
+      ansible.builtin.set_fact:
+        zabbix_web_nginx_installed_version: "{{ (((ansible_facts.packages[item] | first).version | regex_replace('^[0-9]+:', '') | regex_replace('-.*$', '')) | default('0.0.0', true)) }}"
```
eb4x: Oof, sorry. I meant `regex_search`. Just find the x.y.z version string, and extract it, like this;

`"{{ (...).version | regex_search('(\\d+\\.\\d+\\.\\d+)', '\\1') | first }}"`

don't do a `default(...)` here. If none of the packages are present, all the looped items are skipped.

roles/zabbix_web/tasks/nginx.yml
```diff
@@ -25,13 +25,39 @@
...
+      loop:
+        - nginx
+        - nginx-core
+        - nginx-full
+        - nginx-light
+      when:
+        - ansible_facts.packages[item] is defined
```
eb4x: `item in ansible_facts.packages`

roles/zabbix_web/defaults/main.yml
```diff
@@ -22,6 +22,7 @@ zabbix_web_vhost_port: 80
 zabbix_web_vhost_tls_port: 443
 zabbix_web_vhost_listen_ip: "*"
 zabbix_web_tls: false
+zabbix_web_tls_http2: false
```
eb4x: I'm sorry I didn't see this the first time. I thought you had introduced a variable, but zabbix_web_ssl_http2 was just undocumented and hidden in the template. I still like this name better, so just change this to

`zabbix_web_tls_http2: "{{ zabbix_web_ssl_http2 | default(false) }}"`

to avoid breaking any setups using the old. I'll add to a list of deprecation warnings in my fixes for roles/zabbix_web.

eb4x (changes requested):
Looks great, just a few simplifications left 👍 

Sorry about the tls_http2 ssl_http2 mixup.

manuello1996:
> Looks great, just a few simplifications left 👍
> 
> Sorry about the tls_http2 ssl_http2 mixup.

I introduced the change you suggested.

eb4x (approved):
Looks perfect, thanks ❤️

PR #1724, opened 2026-04-20
Streamline Testing Workflows

### LINKED ISSUE #1723: Testing Strategy
@eb4x (any anyone else that wants to chime in) I've been thinking about how we can make our testing more efficient based on the discussion within #1718 .  I see a few ways we can do this.

1.  Concurrency:  Right now, every test trigger runs to completion.  I want to update the concurrency rules so that if a person makes multiple pushes, the latest one will cancel previous ones taking into account the branch (so subsequent pushes to main will kill other pushes to main, but if I have a branch I'm working on and you have another, I wouldn't cancel you.
2. Tiggers:  Right now we run on bull PR and on Push.  In theory we shouldn't push to main until after the PR run has already completed and run which makes the Push trigger completely duplicative.  My suggestion is that we keep the PR triggers (my understanding is that it will re-trigger every time the PR is updated by subsequent pushes that affect that PR), and ditch the push condition entirely.  I would add a new PR condition that triggers only on branches *-prep (branches that we are using to do a numbered release) that does a complete test of everything prior to that being pushed into main (I would say do it on the label event but the label is already created by the time the trigger fires which is to late to catch something.
3. Integration tests:  As the issue points out, integration tests take way more time than everything else combined.  I have a few ideas and curious what your thoughts are.  First, I would argue that if we're making a change to the user module only, only the user test should run.  I found a process where we can selectively target the workflow based on which files were updated that I think is very workable.  This will substantially cut down on the number of tests run.  Changes to certain files (that affect multiple modules) would trigger a complete test but everything else would focus only on the limited changes.  I would argue also that rather than testing all combinations of python/ansible we instead change the strategy to the min and max version of python for each version of ansible.  My assumption would be that if the early and the late versions both work, that there isn't anything in between that would break.  If we wanted to do the complete all combinations tests are part of the updated version release, I would be ok with that as well since it's a relatively rare event.

Thoughts?

pyrodie18:
##### SUMMARY
This PR accomplishes a few things that will improve the overall testing of the project.

1. Adds concurrency checks to all flows so that if a subsequent check is initiated for the same thing, it will cancel any currently running executions so only the latest is running.
2. Implements caching where possible with installing python packages.  Doesn't save a ton of time, but some.
3. Adds targeted checks of the plugin integrations.  This way if only a single (or even multiple) modules or test scenarios change, it will only operate against those vs the entire collection.
4. Builds an overarching final test when a branch with the pattern '*-prep' is created.  This is to support final testing prior to releasing an updated release of the collection.
5. Adds caching to plugin integration tests to greatly reduce the amount of time required to run the ansible setup at the start of each test
6. Changes integration tests to only test against the max and min python versions for each version of ansible.
7. Reduces number of PSK scenarios for proxy so that now it just is really checking the logic against a single sample of each OS and version vs all of the variations.

This closes #1723 

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/.github/workflows/agent.yml
+++ b/.github/workflows/agent.yml
@@ -1,6 +1,7 @@
 ---
 name: "community.zabbix.zabbix_agent"
 on:
+  workflow_call:
   push:
     paths:
       - "roles/zabbix_agent/**"
@@ -15,6 +16,9 @@ on:
       - "molecule/requirements.txt"
       - ".github/workflows/agent.yml"
       - "roles/zabbix_repo/**"
+concurrency:
+  group: agent-${{ github.workflow }}-${{ github.event.pull_request.head.repo.full_name || github.repository }}-${{ github.event.pull_request.head.ref || github.ref }}
+  cancel-in-progress: true
 jobs:
   base:
     runs-on: ubuntu-22.04
--- a/.github/workflows/agent.yml
+++ b/.github/workflows/agent.yml
@@ -20,7 +20,45 @@ concurrency:
   group: agent-${{ github.workflow }}-${{ github.event.pull_request.head.repo.full_name || github.repository }}-${{ github.event.pull_request.head.ref || github.ref }}
   cancel-in-progress: true
 jobs:
+  build-collection:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v6
+        with:
+          python-version: 3.11
+          cache: pip
+          cache-dependency-path: molecule/requirements.txt
+
+      - name: Install latest pip
+        run: python -m pip install --upgrade pip
+
+      - name: Install dependencies
+        run: pip install -r molecule/requirements.txt
+
+      - name: Build the collection
+        run: |
+          # Pin versions to speed up CI
+          sed -i 's/ansible\.windows:\s*"\*"/ansible.windows: "2.8.0"/' galaxy.yml
+          sed -i 's/ansible\.netcommon:\s*"\*"/ansible.netcommon: "7.2.0"/' galaxy.yml
+          sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "2.0.0"/' galaxy.yml
+          sed -i 's/community\.general:\s*"\*"/community.general: "10.5.0"/' galaxy.yml
+          sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.13.0"/' galaxy.yml
+          sed -i 's/community\.postgresql:\s*"\*"/community.postgresql: "3.12.0"/' galaxy.yml
+
+          ansible-galaxy collection build -f
+
+      - name: Upload collection artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: agent-collection
+          path: community-zabbix-*.tar.gz
+
   base:
+    needs: build-collection
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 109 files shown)


.github/workflows/agent.yml
```diff
@@ -58,32 +100,37 @@ jobs:
...
         run: python -m pip install --upgrade pip
 
+      - name: Cache Ansible Galaxy content
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.ansible/collections
+            ~/.ansible/roles
+            ~/.cache/ansible-compat
+          key: ${{ runner.os }}-molecule-galaxy-${{ hashFiles('molecule/requirements.txt', 'molecule/zabbix_agent_tests/common/collections.yml') }}
```
eb4x, line 116: molecule/requirements.txt is pip stuff, and doesn't affect the contents of what we're caching here. Maybe swap it with galaxy.yml, or just remove?

.github/workflows/plugins-integration-full.yml
```diff
@@ -24,45 +31,59 @@ jobs:
...
+          - ansible: stable-2.20
+            python: '3.12'
+          - ansible: stable-2.20
+            python: '3.14'
           - ansible: devel
-            python: '3.10'
+            python: '3.12'
           - ansible: devel
-            python: '3.11'
+            python: '3.14'
```
eb4x, line 60: I think maybe we can do this somewhat simpler/interesting?

I get that ansible-core has a minimum python version requirement. And even though ansible-core-2.16 is only "supported" up to 3.12, I run it on python 3.14.3.

So I'd like to know if it works with the minimum python requirements, and whatever is the latest python. And the bit I care most about is when the python project actually breaks backwards compatibility to code written and working on python 3.10.

.github/workflows/plugins-integration-full.yml
```diff
@@ -72,7 +93,7 @@ jobs:
       # the Zabbix server. To do this we spin up a Docker container using the `matrix`
       # of version and ports specified earlier.
       - name: Zabbix container server provisioning
-        uses: hoverkraft-tech/compose-action@v2.5.0
+        uses: hoverkraft-tech/compose-action@v2
```
eb4x, line 96: downgrade?

.github/workflows/plugins-integration-targeted.yml
```diff
@@ -0,0 +1,179 @@
...
+            echo "has_targets=false" >> "${GITHUB_OUTPUT}"
+            echo "targets=" >> "${GITHUB_OUTPUT}"
+            exit 0
+          fi
+
+          printf -v targets_csv '%s,' "${targets[@]}"
+          targets_csv="${targets_csv%,}"
+
+          echo "has_targets=true" >> "${GITHUB_OUTPUT}"
+          echo "targets=${targets_csv}" >> "${GITHUB_OUTPUT}"
```
eb4x, line 89: Holy s#!t man 😅

eb4x (approved):
I had to read up on `actions/cache` and have ai explain some of the steps. I probably still don't really understand the magic behind it.

Like why don't we cache the collection install and put it behind an `if: ${{ ...cache-hit != 'true }}` type guard? Because 🌈 cache magic 🌈

.github/workflows/agent.yml
pyrodie18: So I'll admit and say I got a little bit lazy and didn't double check CODEX's homework but its a fair catch.  That being said as I'm looking into this deeper, I am rethinking some of the cache strategy so this is likely going to change some more.

.github/workflows/plugins-integration-full.yml
pyrodie18: Is there a more elegant/interesting (I think simpler is the exact opposite) way to do this?  Yes, probably but I think when it comes to just simple readability this is about as simple as it gets.

I understand that you (and likely plenty of other people) use Ansible and many other titles in unsupported configurations that work completely fine but that can't be what we build to.  Ansible has clear left and right limits that we have to support, and for anyone that draws outside of those lines, they are choosing their own adventure and accepting the risks associated with that.  Additionally, trying to figure out at what point something breaks falls outside of what the tests are trying to do.  By your logic success would be one of the versions finally failing but for integration tests by their very definition they must pass before you merge.

.github/workflows/plugins-integration-full.yml
pyrodie18: Actions use the same tagging logic as a lot of docker containers.  In this case 2 != 2.0.0  currently 2 == 2.6.0 (a few weeks ago it was 2.5.0).  Really it means 2.latest

.github/workflows/plugins-integration-targeted.yml
pyrodie18: Again CODEX for this win on this one.

.github/workflows/plugins-integration-full.yml
eb4x: It would be "simpler" in the way you only need to track min version, but yeah I get you.

.github/workflows/javagateway.yml
```diff
@@ -34,44 +65,57 @@ jobs:
           cache-dependency-path: molecule/requirements.txt
 
       - name: Install latest pip
-        run: python -m pip install --upgrade pip
+        run: python -m pip install --upgrade pip yq
```
eb4x, line 68: The `yq` you want to use is [this one](https://github.com/mikefarah/yq), and not installed via pip.

PR #1729, opened 2026-04-23
Fix deprecation warning

### LINKED ISSUE #1728: Deprecated postgresql aliases
##### SUMMARY
Task from community.zabbix.zabbix_proxy : PostgreSQL | Get current database version is using deprecated aliases for community.postgresql collection

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
zabbix_proxy

##### ANSIBLE VERSION

```
ansible [core 2.19.8]
  config file = /Users/username/Documents/repos/ansible/ansible.cfg
  configured module search path = ['/Users/username/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/bin/ansible/lib/python3.12/site-packages/ansible
  ansible collection location = /Users/username/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible/bin/ansible
  python version = 3.12.12 (main, Oct  9 2025, 11:07:00) [Clang 17.0.0 (clang-1700.4.4.1)] (/opt/homebrew/bin/ansible/bin/python3.12)
  jinja version = 3.1.3
  pyyaml version = 6.0.1 (with libyaml v0.2.5)
```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### EXPECTED RESULTS

No deprecation warning

##### ACTUAL RESULTS

```
TASK [community.zabbix.zabbix_proxy : PostgreSQL | Get current database version] ******************************************************************************************************************************************************************************
[WARNING]: Deprecation warnings can be disabled by setting `deprecation_warnings=False` in ansible.cfg.
[DEPRECATION WARNING]: Alias 'port' is deprecated. See the module docs for more information. This feature will be removed from collection 'community.postgresql' version 5.0.0.
[DEPRECATION WARNING]: Alias 'db' is deprecated. See the module docs for more information. This feature will be removed from collection 'community.postgresql' version 5.0.0.
```

l00d3r:
##### SUMMARY

Fixes #1728 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_proxy

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/agent.yml
+++ b/.github/workflows/agent.yml
@@ -77,6 +77,7 @@ jobs:
           sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "2.0.0"/' galaxy.yml
           sed -i 's/community\.general:\s*"\*"/community.general: "10.5.0"/' galaxy.yml
           sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.13.0"/' galaxy.yml
+          sed -i 's/community\.postgresql:\s*"\*"/community.postgresql: "3.5.0"/' galaxy.yml
 
           collection_file=$( basename $(ansible-galaxy collection build -f | awk -F" " '{print $NF}'))
           echo "COLLECTION_FILE=$collection_file" >> $GITHUB_ENV
--- a/.github/workflows/agent.yml
+++ b/.github/workflows/agent.yml
@@ -77,7 +77,7 @@ jobs:
           sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "2.0.0"/' galaxy.yml
           sed -i 's/community\.general:\s*"\*"/community.general: "10.5.0"/' galaxy.yml
           sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.13.0"/' galaxy.yml
-          sed -i 's/community\.postgresql:\s*"\*"/community.postgresql: "3.5.0"/' galaxy.yml
+          sed -i 's/community\.postgresql:\s*"\*"/community.postgresql: "3.12.0"/' galaxy.yml
 
           collection_file=$( basename $(ansible-galaxy collection build -f | awk -F" " '{print $NF}'))
           echo "COLLECTION_FILE=$collection_file" >> $GITHUB_ENV
--- a/.github/workflows/agent.yml
+++ b/.github/workflows/agent.yml
@@ -77,7 +77,6 @@ jobs:
           sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "2.0.0"/' galaxy.yml
           sed -i 's/community\.general:\s*"\*"/community.general: "10.5.0"/' galaxy.yml
           sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.13.0"/' galaxy.yml
-          sed -i 's/community\.postgresql:\s*"\*"/community.postgresql: "3.12.0"/' galaxy.yml
 
           collection_file=$( basename $(ansible-galaxy collection build -f | awk -F" " '{print $NF}'))
           echo "COLLECTION_FILE=$collection_file" >> $GITHUB_ENV
--- a/.github/workflows/javagateway.yml
+++ b/.github/workflows/javagateway.yml
@@ -76,6 +76,7 @@ jobs:
           sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "2.0.0"/' galaxy.yml
           sed -i 's/community\.general:\s*"\*"/community.general: "10.5.0"/' galaxy.yml
           sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.13.0"/' galaxy.yml
+          sed -i 's/community\.postgresql:\s*"\*"/community.postgresql: "3.5.0"/' galaxy.yml
 
           collection_file=$( basename $(ansible-galaxy collection build -f | awk -F" " '{print $NF}'))
           echo "COLLECTION_FILE=$collection_file" >> $GITHUB_ENV
--- a/.github/workflows/javagateway.yml
+++ b/.github/workflows/javagateway.yml
@@ -76,7 +76,6 @@ jobs:
           sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "2.0.0"/' galaxy.yml
           sed -i 's/community\.general:\s*"\*"/community.general: "10.5.0"/' galaxy.yml
           sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.13.0"/' galaxy.yml
-          sed -i 's/community\.postgresql:\s*"\*"/community.postgresql: "3.12.0"/' galaxy.yml
 
           collection_file=$( basename $(ansible-galaxy collection build -f | awk -F" " '{print $NF}'))
           echo "COLLECTION_FILE=$collection_file" >> $GITHUB_ENV
--- a/.github/workflows/javagateway.yml
+++ b/.github/workflows/javagateway.yml
@@ -76,7 +76,7 @@ jobs:
           sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "2.0.0"/' galaxy.yml
           sed -i 's/community\.general:\s*"\*"/community.general: "10.5.0"/' galaxy.yml
           sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.13.0"/' galaxy.yml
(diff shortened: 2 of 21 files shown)


l00d3r:
I don't understand why the tests failed, my commit did not even change the files for mysql

eb4x:
Just speculating here, but you'll probably have to address these two aswell;
https://github.com/ansible-collections/community.zabbix/blob/main/.github/workflows/proxy.yml#L83
https://github.com/ansible-collections/community.zabbix/blob/main/galaxy.yml#L27

And if you're bumping the collection in galaxy.yml you should update the pinning in all the other workflows, and provide a changelog fragment about the new minimum requirement.

pyrodie18:
All of the proxy installs with MySQL started to fail for some reason and I can't figure out why.  It's the import schema step.  When I test it on my local using the same molecule run we do for testing, it runs fine but keeps failing for some reason on here.  Any ideas @eb4x ?

pyrodie18:
Also one other comment, we're gonna have to hold this until 5.0.0 because we need to update the version lock on the mysql collection with this one.

l00d3r:
@eb4x Sorry I don't follow, I updated only roles/zabbix_proxy/tasks/initialize-pgsql.yml and the alias that I changed was already in use in the other block in the same file so the minimum requirement should stay the same for my change

l00d3r:
Found the problem, I wrongly assumed the parameter names were the same for postgresql modules, corrected the issue

l00d3r:
pgsql tests now ran successfully.
@pyrodie18 I don't understand why do you need to update the version lock for mysql here?

eb4x:
> All of the proxy installs with MySQL started to fail for some reason and I can't figure out why. It's the import schema step. When I test it on my local using the same molecule run we do for testing, it runs fine but keeps failing for some reason on here. Any ideas @eb4x ?

No idea yet, just like you I haven't managed to reproduce it locally. My immediate theories that I haven't concluded fully yet;

* Some change introduced to the tables in a recent release of zabbix, but it's odd that it affects all v60, v70, v72 and v74, so probably not that.
* some issue with community.mysql collection, not present in newer version, but i downgraded mine to the version we pin in CI, and still works for me.
* pymysql version issue, but across debian/redhat/suse/ubuntu all at the same time, unlikely.
* some issue in the underlying ubuntu-22.04 workflow runner with python/pymysql there seems most plausible. (would we see the same errors on zabbix_server for mysql right now?)

pyrodie18:
> pgsql tests now ran successfully. @pyrodie18 I don't understand why do you need to update the version lock for mysql here?

If you look at the galaxy.yml under postgresql version is set to `*`.  Obviously that is no longer true.  At some point in time the setting was called 'port' and there was was no setting for 'login_port'.  At some point, they added 'login_port' and deprecated 'port'.  Wherever that occured needs to now be at the very least the min version of the plugin.  To speed up testing, during the testing pipeline we force the version to 3.12.0 but a user as the galaxy file says, they could be using 1.0.0 and still be fine.

That being said, we already have a 5.0 dev branch going that we can just merge this into that when we're ready.

pyrodie18:
> > All of the proxy installs with MySQL started to fail for some reason and I can't figure out why. It's the import schema step. When I test it on my local using the same molecule run we do for testing, it runs fine but keeps failing for some reason on here. Any ideas @eb4x ?
> 
> No idea yet, just like you I haven't managed to reproduce it locally. My immediate theories that I haven't concluded fully yet;
> 
> 1.   Some change introduced to the tables in a recent release of zabbix, but it's odd that it affects all v60, v70, v72 and v74, so probably not that.
> 2.   some issue with community.mysql collection, not present in newer version, but i downgraded mine to the version we pin in CI, and still works for me.
> 3.   pymysql version issue, but across debian/redhat/suse/ubuntu all at the same time, unlikely.
> 4.   some issue in the underlying ubuntu-22.04 workflow runner with python/pymysql there seems most plausible. (would we see the same errors on zabbix_server for mysql right now?)

Yep agree.  I forced an install on the latest minor version and was still good on my local system.  Did the same as you on 2 and same results.  Agree with you on 3.  I just kicked a build of server on rocky10 v74 mysql on github and it failed on the same spot.

pyrodie18:
We were both wrong @eb4x, its actually the mysql container.  They just released 8.0.46.  My local system was using 8.0.45.  When I forced the upgrade, I got it to fail locally.  Looking at it, 8.0 is actually EOL next week.  I am going to check the Zabbix compatibility matrix.  If all current versions support 8.4, I'm gonna role to that and if not then I will force pining to 8.0.45.

pyrodie18:
OK, latest update.  Yes all versions work fine with 8.4 but when I upgrade the version and did a molecule run I immediately failed even trying to create the dbuser.  Not going to spend cycles troubleshooting this right this second.  Will open a separate issue for that.  For right now, going to force 8.0.45 so that CI/CD passes and life can continue.

eb4x:
> OK, latest update. Yes all versions work fine with 8.4 but when I upgrade the version and did a molecule run I immediately failed even trying to create the dbuser. Not going to spend cycles troubleshooting this right this second. Will open a separate issue for that. For right now, going to force 8.0.45 so that CI/CD passes and life can continue.

Good find :+1:  Yeah, just pin it to last reliable release.

pyrodie18 (changes requested):
OK @l00d3r sorry about that.  If I can ask for two things. 

1.  Add a change fragment.  Again I would argue that this is a breaking change (unless you disagree @eb4x )
2.  Assuming that eb4x doesn't disagree, point this over to the 5.0-dev branch.  I've already updated the testing workflow there as well so it should pass without issue.

eb4x:
> OK @l00d3r sorry about that. If I can ask for two things.
> 
> 1. Add a change fragment.  Again I would argue that this is a breaking change (unless you disagree @eb4x )
> 2. Assuming that eb4x doesn't disagree, point this over to the 5.0-dev branch.  I've already updated the testing workflow there as well so it should pass without issue.

I agree, pinning the minimum version, although a welcome change is a breaking change.

pyrodie18:
looks like 3.14 is where they made the change

l00d3r:
I hope I understood correctly, the earliest community.postgresql that runs zabbix_proxy and zabbix_server roles successfully is 3.5.0, should I then update the version in files @eb4x mentioned to that version?

pyrodie18:
Nope you're good.  Tests passed.  I'm at the airport but will pull it in later.  Thanks for the PR.

l00d3r:
Sorry before you answered I already pushed the changelog fragment and set the minimum version to 3.5.0, I figured since I set the minimum supported version to 3.5.0 I could remove the sed usage for it in the workflow.

I also have a question if maybe we could get rid of the entire workflow part that modifies galaxy.yml with sed and set the versions used in workflows to galaxy.yml?

l00d3r:
Sorry, never mind my last question, I wrote the question, posted it and then thought about it, readded the version pin

pyrodie18:
I gotta dig through the PG collection and see where they first added the login_port and login_db aliases to the original options.  That will be the version number.  I don't think its 3.5 though.  Just leave this where its at for right now.

l00d3r:
I already digged thorugh it, login_port and login_db are there from the [beginning](https://github.com/ansible-collections/community.postgresql/blob/0.1.0/plugins/module_utils/postgres.py) however why 3.5.0 is minimum version is because of missing postgresql_user configuration parameter that got added [here](https://github.com/ansible-collections/community.postgresql/pull/722) and is used by this collection [here](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_server/tasks/initialize-pgsql.yml#L80) and [here](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_proxy/tasks/initialize-pgsql.yml#L83).

masa-orca:
I don’t think it makes sense to change the testing version for the MySQL ansible collection to Minimum supported version.
In reality, since we don’t test every version, we don’t know which versions will work with this role.
All we can say is that this role works with the tested version.
If we want to say “Supported versions" clealy, we would need to test every version. However, due to limitations in GitHub’s workflow, that is not possible.

pyrodie18:
OK, I get why we're setting the min version to 3.5.0 but also agree with @masa-orca.  Do me a favor and remove the changes to the workflow file (I'm making some significant changes to it anyway).  Once that happens I'll move it into the 5.0-dev branch and close this out.

l00d3r:
@pyrodie18 I removed the changes but I'm unable to move the change to proxy.yml for some reason

eb4x (commented):
LGTM 👍

PR #1730, opened 2026-04-23
For upstream molecule web

eb4x:
##### SUMMARY
Speed up testing by using official zabbix container images for zabbix-server portion.
Drop use of geerlingguy roles to avoid cludge in regards to missing or delayed support of debian/suse/ubuntu.

##### ISSUE TYPE
- Streamlining Pull Request

##### COMPONENT NAME
molecule/web

### DIFF
--- a/molecule/zabbix_web/create.docker.yml
+++ b/molecule/zabbix_web/create.docker.yml
@@ -1,0 +1,84 @@
+---
+- name: Create
+  hosts: localhost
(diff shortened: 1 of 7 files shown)


eb4x:
Had one more issue with testing suse with nginx, but I believe this is done now.

pyrodie18 (approved):

PR #1731, opened 2026-04-24
Fix MySQL for Molecule

pyrodie18:
##### SUMMARY
The MySQL docker recently rolled to 8.0.46 which seems to have a breaking change in it.  8.0.45 works though so forcing the container to use that version

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Molecule Testing

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -23,25 +23,25 @@ jobs:
       matrix:
         container:
           - image: rockylinux10
-          - image: rockylinux9
-          - image: rockylinux8
-            ansible_core: ansible-core<2.17
-          - image: ubuntu2404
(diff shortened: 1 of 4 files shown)


l00d3r:
I hope someone gets a notification for this but the issue comes from mysql [bug](https://bugs.mysql.com/bug.php?id=120323), affected versions I've tested with this issue are 8.4.9 and 8.0.46

Since it's not mentioned here the error encountered this PR worked around was:

```
ERROR 1118 (42000) at line 1306: Row size too large (> 8126). Changing some columns to TEXT or BLOB may help. In current row format, BLOB prefix of 0 bytes is stored inline.
```

PR #1735, opened 2026-04-25
Mysql 8.4

### LINKED ISSUE #1732: Troubleshoot MySQL 8.4
##### SUMMARY
While troubleshooting #1731 and #1729 discovered that there was a problem with the latest MySQL docker for 8.0.  8.0 goes EOL next week so was going to upgrade to 8.4 but 8.4 immediately failed when trying to create the user.  Need to troubleshoot and check the logic of that and see why it's breaking.

Probably also need to add integration testing with all of the LTS versions of MySQL and PGS.  I don't think we need to test the entire matrix but just a single run across each LTS version to ensure we stay compatible.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
Server and Proxy roles

l00d3r:
##### SUMMARY
Fixes #1732 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
zabbix_proxy
zabbix_web

##### ADDITIONAL INFORMATION
Uses mysql 8.4.8 cause 8.4.9 has a [bug](https://bugs.mysql.com/bug.php?id=120323) that breaks db schema import
Adds command to db container creation cause [mysql_native_password is getting deprecated](https://dev.mysql.com/doc/refman/8.4/en/native-pluggable-authentication.html) and needs to be enabled for mysql 8.4 by using --mysql-native-password=ON

```
The mysql_native_password authentication plugin is deprecated as of MySQL 8.0.34, disabled by default in MySQL 8.4, and removed as of MySQL 9.0.0. 
```

### DIFF
--- a/molecule/default/create.docker.yml
+++ b/molecule/default/create.docker.yml
@@ -30,6 +30,7 @@
         image: "{{ _container[_db_type].image }}"
         state: started
         recreate: true
+        command: "{{ _container[_db_type].command | default(omit) }}"
         networks:
           - name: zabbix
         env: "{{ _container[_db_type].env }}"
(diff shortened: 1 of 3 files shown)


l00d3r:
Closing in favor of #1736

PR #1736, opened 2026-04-26
Add role support for mysql 8.4 and 9.0

### LINKED ISSUE #1732: Troubleshoot MySQL 8.4

l00d3r:
##### SUMMARY

Adds support for mysql 8.4 and 9.0, fixes #1732 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zabbix_server
zabbix_proxy

##### ADDITIONAL INFORMATION

Mysql is deprecating support for mysql_native_password authentication plugin, this plugin is disabled by default in mysql 8.4 and removed from mysql 9.0, this is the default when creating a mysql user using community.mysql.mysql_user, this PR adds two variables for zabbix_server role called `zabbix_server_mysql_auth_plugin`and `zabbix_server_mysql_auth_plugin_salt` and two for zabbix_proxy role called `zabbix_proxy_mysql_auth_plugin` and `zabbix_proxy_mysql_auth_plugin_salt`

Using conditional tests this change makes sure nobodys existing use of this role breaks something for them


l00d3r:
The failed tasks are lack of support for caching_sha2_password on opensuse leap 15, it's EOL is in 4 days, I'm thinking maybe we should deprecate running the checks on leap 15 instead of fixing that issue by installing an extra package on leap 15?

l00d3r:
I actually don't understand where the issue comes from, running molecule test locally does not produce the same error

l00d3r:
Figured out why my local tests were successful, I did not specify image correctly, figured out the correct package name for caching_sha2_password support is libmariadb_plugins and even if we deprecate support for leap 15 it is also needed on  leap 16.

l00d3r:
opensuse tests are now fixed, the last run had some transient error:

```
Starting collection install process
Installing 'community.zabbix:4.2.0' to '/home/runner/.ansible/collections/ansible_collections/community/zabbix'
community.zabbix:4.2.0 was installed successfully
Server https://galaxy.ansible.com/api/ has not signed ansible.posix:2.0.0
Fetching a collection tarball for 'ansible.posix:2.0.0' from Ansible Galaxy
Downloading https://galaxy.ansible.com/api/v3/plugin/ansible/content/published/collections/artifacts/ansible-posix-2.0.0.tar.gz to /home/runner/.ansible/tmp/ansible-local-2047gnisr_ye/tmp5j1sjnrq/ansible-posix-2.0.0-khw68ehz
Error: Process completed with exit code 1.
```

roles/zabbix_proxy/tasks/initialize-mysql.yml
```diff
@@ -61,7 +61,16 @@
...
         name: "{{ zabbix_proxy_dbuser }}"
-        password: "{{ zabbix_proxy_dbpassword }}"
+        password: >-
+          {{ zabbix_proxy_dbpassword
+            if zabbix_proxy_mysql_auth_plugin == 'mysql_native_password'
+            else omit }}
+        plugin_auth_string: >-
+          {{ zabbix_proxy_dbpassword
+            if zabbix_proxy_mysql_auth_plugin in ['caching_sha2_password', 'sha256_password']
+            else omit }}
```
eb4x: Switch to the ternary wherever possible, it just groups the things you evaluate/care about better.

`{{ (zabbix_proxy_mysql_auth_plugin == 'mysql_native_password') | ternary(zabbix_proxy_dbpassword, omit) }}`

roles/zabbix_proxy/vars/Suse.yml
```diff
@@ -7,6 +7,7 @@ _zabbix_proxy_pgsql_dependencies:
 _zabbix_proxy_mysql_dependencies:
   - "{{ zabbix_proxy_install_database_client | ternary('mysql', '') }}"
   - python3-PyMySQL
+  - "{{ 'libmariadb_plugins' if zabbix_proxy_mysql_auth_plugin in ['caching_sha2_password', 'sha256_password'] | default('') }}"
```
eb4x: `(condition) | ternary(...` here aswell.

pyrodie18:
why change the community.mysql min version?

roles/zabbix_proxy/defaults/main.yml
```diff
@@ -29,6 +29,7 @@ zabbix_proxy_dbencoding: utf8
 zabbix_proxy_dbpassword_hash_method: md5
 zabbix_proxy_install_database_client: true
 zabbix_proxy_database_file_dir: "{{ zabbix_proxy_version is version('7.0', '>') | ternary(_zabbix_proxy_database_file_dir_post72, _zabbix_proxy_database_file_dir_pre72 ) }}"
+zabbix_server_mysql_auth_plugin: mysql_native_password
```
eb4x: Defaulting to something that won't work in the future, feels counter-productive.

Is there any way for us to query this information from mysql itself? Determine if we can use the new recommended practice on fresh installs, and if we detect old versions or preexisting users, we could use set_fact to override for legacy, or even upgrade an existing user from native to some sha2 version?

eb4x (changes requested):
This is really outstanding work 👍 

I just wonder regarding defaulting to the old, is there any way for us to query this information from mysql itself? Determine if we can use the new recommended practice on fresh installs, and if we detect old versions or preexisting users, we could use set_fact to override for legacy. (or even upgrade an existing user from native_password to sha256_password version?)

l00d3r:
@pyrodie18 I changed the minimum version of mysql to the minimal version community.zabbix needs if this PR was going to be accepted as community.mysql.mysql_user added the `salt:` parameter in 3.10.0 and I understand this change is good for this project by a comment made by @eb4x [here](https://github.com/ansible-collections/community.zabbix/pull/1729#issuecomment-4316877910)?

I made the change in the workflow files since if we set the minimum then we should also make sure it works with minimum.

roles/zabbix_proxy/vars/Suse.yml
l00d3r: Thank you for the suggestion, implemented in both zabbix_proxy and zabbix_server

roles/zabbix_proxy/tasks/initialize-mysql.yml
l00d3r: Thank you for the suggestion, implemented in both zabbix_proxy and zabbix_server

roles/zabbix_proxy/vars/Suse.yml
```diff
@@ -7,7 +7,7 @@ _zabbix_proxy_pgsql_dependencies:
 _zabbix_proxy_mysql_dependencies:
   - "{{ zabbix_proxy_install_database_client | ternary('mysql', '') }}"
   - python3-PyMySQL
-  - "{{ 'libmariadb_plugins' if zabbix_proxy_mysql_auth_plugin in ['caching_sha2_password', 'sha256_password'] | default('') }}"
+  - "{{ (zabbix_proxy_mysql_auth_plugin in ['caching_sha2_password', 'sha256_password']) | ternary('libmariadb_plugins', omit) }}"
```
eb4x: `ternary('libmariadb_plugins', '')`

'libmariadb_plugins' or empty string. (omit will put a really weird placeholder here)

l00d3r: Sorry, fixed in the latest commit

eb4x:
> @pyrodie18 I changed the minimum version of mysql to the minimal version community.zabbix needs if this PR was going to be accepted as community.mysql.mysql_user added the `salt:` parameter in 3.10.0 and I understand this change is good for this project by a comment made by @eb4x [here](https://github.com/ansible-collections/community.zabbix/pull/1729#issuecomment-4316877910)?
> 
> I made the change in the workflow files since if we set the minimum then we should also make sure it works with minimum.

If we use a feature not covered by earlier versions of a collection we use, we need to pin that requirement in `galaxy.yml`, but I don't want us to pin our CI tests against the minimum version. We'd want to bump the version of that collection in testing to catch and/or discover new features and changes.

Not pinning 3.10.0 is an oversight on my part, and really good you caught it :+1:

roles/zabbix_proxy/defaults/main.yml
l00d3r: Agreed, I can whip up something like that later. 

We can get supported plugins info from INFORMATION_SCHEMA.PLUGINS (checked on mysql  Ver 15.1 Distrib 10.11.10-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper) with this query 

```
SELECT * FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_TYPE='AUTHENTICATION';
```

We can get info about which plugin the user currently uses with this query:

```
SELECT user, plugin FROM mysql.user where user='zabbix_proxy';
``` 

I think we could go either way but opensuse does not even support these plugins by default so I still think we need some toggle that enables/disables the upgrade and/or usage of the sha2 plugins.

l00d3r:
> > @pyrodie18 I changed the minimum version of mysql to the minimal version community.zabbix needs if this PR was going to be accepted as community.mysql.mysql_user added the `salt:` parameter in 3.10.0 and I understand this change is good for this project by a comment made by @eb4x [here](https://github.com/ansible-collections/community.zabbix/pull/1729#issuecomment-4316877910)?
> > I made the change in the workflow files since if we set the minimum then we should also make sure it works with minimum.
> 
> If we use a feature not covered by earlier versions of a collection we use, we need to pin that requirement in `galaxy.yml`, but I don't want us to pin our CI tests against the minimum version. We'd want to bump the version of that collection in testing to catch and/or discover new features and changes.
> 
> Not pinning 3.10.0 is an oversight on my part, and really good you caught it 👍

Just so we're on the same page should I remove the changes done to workflow files and keep the minimum version in galaxy.yml?

Also I wanted to point out the 3.10.0 requirement came from my suggested change, I'm not sure what the earlier requirements were.

roles/zabbix_proxy/defaults/main.yml
l00d3r: I added automatic selection of authentication plugins to zabbix_proxy role, what do you think about doing it like that? 

To be honest to me it seems to introduce unneeded fragility to the role...

> or even upgrade an existing user from native to some sha2 version?

This is already taken care of by mysql_user, it will replace the existing user with the more modern authentication plugin when it is used.

roles/zabbix_proxy/tasks/initialize-mysql.yml
```diff
@@ -40,6 +40,32 @@
...
+            zabbix_proxy_mysql_auth_plugin_priority
+            | select('in', ((zabbix_proxy_active_mysql_authentication_plugins.query_result | flatten) | map(attribute='PLUGIN_NAME')))
+            | list
+            | first
+          }}
+      vars:
+        zabbix_proxy_mysql_auth_plugin_priority:
+          - caching_sha2_password
+          - sha256_password
+          - mysql_native_password
```
eb4x: This is excellent, I'd only suggest putting the query_result in the vars section, and shorten some of the names, and prefix anything "non-user-overridable" with underscore. Sorta like

```yaml
- set_fact:
    zabbix_proxy_mysql_auth_plugin: >-
      {{ _auth_plugin_priority | select('in', _active_auth_plugins) | first }}
  vars:
    _active_auth_plugins: "{{ zabbix_proxy_mysql_active_auth_plugins.query_result | flatten | map(attribute='PLUGIN_NAME') }}"
    _auth_plugin_priority:
      - caching_sha2_password
      - ...
```

We should probably let the user override this with a `when` condition? So the user can specify `zabbix_proxy_mysql_auth_plugin` to something else. Incase the `thundercougar_auth2000_falconbird` comes along. And since we already have the list of active plugins, we could assert it exists in that list?

molecule/zabbix_proxy/molecule.yml
```diff
@@ -62,6 +62,8 @@ provisioner:
         zabbix_proxy_mysql_login_user: root
         zabbix_proxy_mysql_login_password: changeme
         zabbix_proxy_mysql_login_port: 3306
+        zabbix_proxy_mysql_auth_plugin: caching_sha2_password
+        zabbix_proxy_mysql_auth_plugin_salt: 0123456789abcdfeghij
```
eb4x: We can probably count on these being set in the `set_facts` task where we set the auth_plugin if it's not declared. (maybe an additional task for the salt if the first set_facts triggers.)

```yaml
- when: zabbix_proxy_mysql_auth_plugin is not defined
  set_fact: ...
  register: _auth_plugin_configured
- when:
    - _auth_plugin_configured is changed
    - zabbix_proxy_mysql_auth_plugin in ['...sha2...', '...sha256']
  set_fact:
    ..._plugin_salt: "{{ lookup('password', '/tmp/saltfile length=16') }}
```
The `lookup('password' ...)` should preserve idempotency for molecule/github-workflow runs. Maybe the salt could always live in `defaults/main.yml` ?

roles/zabbix_proxy/tasks/initialize-mysql.yml
```diff
@@ -61,7 +87,10 @@
         login_port: "{{ zabbix_proxy_mysql_login_port | default(omit) }}"
         login_unix_socket: "{{ zabbix_proxy_mysql_login_unix_socket | default(omit) }}"
         name: "{{ zabbix_proxy_dbuser }}"
-        password: "{{ zabbix_proxy_dbpassword }}"
+        password: "{{ (zabbix_proxy_mysql_authentication_plugin in ['mysql_native_password']) | ternary(zabbix_proxy_dbpassword, omit) }}"
+        plugin_auth_string: "{{ (zabbix_proxy_mysql_authentication_plugin in ['caching_sha2_password', 'sha256_password']) | ternary(zabbix_proxy_dbpassword, omit) }}"
+        plugin: "{{ zabbix_proxy_mysql_authentication_plugin | default(omit) }}"
+        salt: "{{ zabbix_proxy_mysql_auth_plugin_salt | default(omit) }}"
```
eb4x: 👍

roles/zabbix_proxy/defaults/main.yml
```diff
@@ -29,6 +29,7 @@ zabbix_proxy_dbencoding: utf8
 zabbix_proxy_dbpassword_hash_method: md5
 zabbix_proxy_install_database_client: true
 zabbix_proxy_database_file_dir: "{{ zabbix_proxy_version is version('7.0', '>') | ternary(_zabbix_proxy_database_file_dir_post72, _zabbix_proxy_database_file_dir_pre72 ) }}"
+zabbix_proxy_mysql_auth_plugin: mysql_native_password
```
eb4x: probably commented out, but the salt could go here.

.github/workflows/proxy.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           sed -i 's/ansible\.netcommon:\s*"\*"/ansible.netcommon: "7.2.0"/' galaxy.yml
           sed -i 's/ansible\.posix:\s*"\*"/ansible.posix: "2.0.0"/' galaxy.yml
           sed -i 's/community\.general:\s*"\*"/community.general: "10.5.0"/' galaxy.yml
-          sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.13.0"/' galaxy.yml
+          sed -i 's/community\.mysql:\s*"\*"/community.mysql: "3.10.0"/' galaxy.yml
```
eb4x: Don't downgrade in workflows, but you can pin a minimum in galaxy.yml (as you've done).

eb4x (commented):
Super research 🤩, thanks for this ❤️

I think we want to provide overrides for the users to either future-proof this or shoot themselves in the foot, but default to the latest and securest, if we can.

And it may be more ansible-esq to just put

```yaml
zabbix_proxy_auth_plugin: ""
zabbix_proxy_auth_salt: ""
```

and update the when conditions accordingly, instead of `is defined` as i suggested in some comments.

roles/zabbix_proxy/tasks/initialize-mysql.yml
eb4x: actually, maybe abuse the `vars:` section with the booleans/flags for this task,

```yaml
  plugin: "{{ zabbix_proxy_mysql_auth_plugin }}"
  password: "{{ _legacy_auth | ternary(zabbix_proxy_dbpassword, omit) }}"
  ...
vars:
  _legacy_auth: "{{ zabbix_proxy_mysql_authentication_plugin == 'mysql_native_password' }}"
  _modern_auth: "{{ zabbix_proxy_mysql_authentication_plugin in ['caching_sha2_password', 'sha256_password'] }}"
```

The plugin doesn't need an omit, since we're always gonna set it.

l00d3r: Removed the omit and used vars which, by the way, I think is genius, thanks

roles/zabbix_proxy/tasks/initialize-mysql.yml
l00d3r: Genius use of vars. reworked

molecule/zabbix_proxy/molecule.yml
l00d3r: I think the lookup('password' for salt is perfect and with that we can remove the default altogether

l00d3r: I removed the molecule overrides

roles/zabbix_proxy/defaults/main.yml
l00d3r: I think we should not set a default salt, I like the lookup('password', way much better here

.github/workflows/proxy.yml
l00d3r: Reverted

l00d3r:
I think the variables should have mysql in them as all the other db specific variables seem to have it.

I think we should prompt users to set the salt with a warning if it's empty, do you know of any other way for it instead of bugging with non-idempotency on reboot? 

What does the third party check do and why did it fail?

roles/zabbix_proxy/tasks/initialize-mysql.yml
```diff
@@ -40,6 +40,52 @@
...
   block:
+    - name: "MySQL | Register active MySQL authentication plugins"
+      community.mysql.mysql_query:
+        login_user: "{{ zabbix_proxy_mysql_login_user | default(omit) }}"
+        login_password: "{{ zabbix_proxy_mysql_login_password | default(omit) }}"
+        login_host: "{{ zabbix_proxy_mysql_login_host | default(omit) }}"
+        login_port: "{{ zabbix_proxy_mysql_login_port | default(omit) }}"
+        login_db: "mysql"
+        query: "SELECT PLUGIN_NAME FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_TYPE='AUTHENTICATION' AND PLUGIN_STATUS='ACTIVE';"
+      register: _zabbix_proxy_auth_plugins
```
eb4x, line 51: Perfection, no notes 👍

roles/zabbix_proxy/tasks/initialize-mysql.yml
```diff
@@ -61,10 +107,16 @@
...
+        password: "{{ _zabbix_proxy_legacy_auth | ternary(zabbix_proxy_dbpassword, omit) }}"
+        plugin_auth_string: "{{ _zabbix_proxy_modern_auth | ternary(zabbix_proxy_dbpassword, omit) }}"
+        plugin: "{{ _zabbix_proxy_auth_plugin }}"
+        salt: "{{ _zabbix_proxy_modern_auth | ternary(_zabbix_proxy_auth_salt, omit) }}"
         host: "{{ zabbix_proxy_privileged_host }}"
         priv: "{{ zabbix_proxy_dbname }}.*:ALL"
         state: present
+      vars:
+        _zabbix_proxy_legacy_auth: "{{ _zabbix_proxy_auth_plugin == 'mysql_native_password' }}"
+        _zabbix_proxy_modern_auth: "{{ _zabbix_proxy_auth_plugin in ['caching_sha2_password', 'sha256_password'] }}"
```
eb4x: 👍

roles/zabbix_proxy/tasks/initialize-mysql.yml
```diff
@@ -61,10 +107,16 @@
         login_port: "{{ zabbix_proxy_mysql_login_port | default(omit) }}"
         login_unix_socket: "{{ zabbix_proxy_mysql_login_unix_socket | default(omit) }}"
         name: "{{ zabbix_proxy_dbuser }}"
-        password: "{{ zabbix_proxy_dbpassword }}"
+        password: "{{ _zabbix_proxy_legacy_auth | ternary(zabbix_proxy_dbpassword, omit) }}"
+        plugin_auth_string: "{{ _zabbix_proxy_modern_auth | ternary(zabbix_proxy_dbpassword, omit) }}"
+        plugin: "{{ _zabbix_proxy_auth_plugin }}"
+        salt: "{{ _zabbix_proxy_modern_auth | ternary(_zabbix_proxy_auth_salt, omit) }}"
```
eb4x: Change to
```yaml
plugin: "{{ zabbix_proxy_auth_plugin }}"
salt: "{{ _zabbix_proxy_modern_auth | ternary(zabbix_proxy_auth_salt, omit) }}"
```
The variables with underscores are for us and our logic. the non-underscored are for the user to be able to configure alternate values.

eb4x (changes requested):
This is close 👍

eb4x:
> What does the third party check do and why did it fail?

It's some third-party infrastructure that's having an outage and can't schedule some testing we've defined. It's not a hindrance to this getting merged :+1:

roles/zabbix_proxy/tasks/initialize-mysql.yml
```diff
@@ -40,6 +40,52 @@
...
+        _zabbix_proxy_auth_plugin: >-
+          {{
+              zabbix_proxy_mysql_auth_plugin is defined
+              | ternary(zabbix_proxy_mysql_auth_plugin, _zabbix_proxy_auth_plugin_priority | select('in', _zabbix_proxy_auth_plugins_list) | first)
+          }}
+        _zabbix_proxy_auth_salt: >-
+          {{
+              zabbix_proxy_mysql_auth_salt is defined
+              | ternary(zabbix_proxy_mysql_auth_salt, lookup('ansible.builtin.password', '/tmp/saltfile', length=20))
+          }}
```
eb4x: (Sorry, this was supposed to come with the previous review, not sure what happened. And it's not even pointing at the code I was mentioning)

The assert and set_facts tasks got weird. Like is there a point to assert the length of the salt if it's defined but auth_plugin is set to native?

We can do this simpler; Add these to `defaults/main.yml` (anything in `defaults/main.yml` is an indication to the user that they can override these variables, even if not documented in `docs/ZABBIX_*.md`)

```yaml
zabbix_proxy_mysql_auth_plugin: ""
zabbix_proxy_mysql_auth_salt: "{{ lookup('password', '/tmp/saltfile', length=20) }}"
```

Then the tasks can become

```yaml

- name: MySQL | Ensure that user selected authentication plugin is supported by the MySQL server
  when: zabbix_proxy_mysql_auth_plugin | length > 0  # User has specified an auth plugin
  ansible.builtin.assert:
    that:
      - zabbix_proxy_mysql_auth_plugin in _zabbix_proxy_auth_plugins_list
  vars:
    _zabbix_proxy_auth_plugins_list: "{{ _zabbix_proxy_auth_plugins.query_result | flatten | map(attribute='PLUGIN_NAME') }}"

- name: MySQL | Determine best authentication plugin
  when: zabbix_proxy_mysql_auth_plugin | length == 0  # User has not specified an auth plugin
  set_fact:
    zabbix_proxy_mysql_auth_plugin: >-
      {{  _zabbix_proxy_auth_plugin_priority | select('in', _zabbix_proxy_auth_plugins_list) | first}}
  vars:
    _zabbix_proxy_auth_plugins_list: ...
    _zabbix_proxy_auth_plugin_priority: ...

- name: MySQL | Ensure that user selected salt for authentication plugin is 20 characters long
  when: zabbix_proxy_mysql_auth_plugin in ['caching...', 'sha256...']
  assert:
    that:
      - zabbix_proxy_mysql_auth_salt | length == 20
```

The logic being,
1. If the user has specified a auth_plugin, verify it's active or fail.
2. If no auth_plugin specified, pick one from active.
3. If auth_plugin requires salt, verify that the user didn't mess that up.

l00d3r: Implemented

eb4x (approved):
Love it 🤩

eb4x:
@pyrodie18 Any objections to merging this?

I think if we want to have non-breaking changes for a release, we should branch out something like "stable-4.x", and cherry-pick fixes into that branch for any updates to that release, and let main just always be the latest and greatest?

pyrodie18:
I'm good, my only question is does #1735 still need to go in or is it already included in this?

l00d3r:
Closed the other one, that was more of a workaround, this one should actually fix the issue.

eb4x:
I think this turned out great! Thank you for your contribution :heart:

PR #1738, opened 2026-05-02
fix: add login_unix_socket to MySQL version check and schema import

tobstone:
When MySQL is configured to listen on a Unix socket (the default on many Linux installations), the community.mysql modules require login_unix_socket to connect locally. Without it, both the schema-version check and the      
  schema import tasks fail because they attempt a TCP connection instead.                                                                                                                                                         
                                                                                                                                                                                                                                  
  This adds an optional zabbix_server_mysql_login_unix_socket variable to those two tasks. It defaults to omit, so existing deployments using TCP are unaffected.


pyrodie18:
@tobstone thanks for the contribution.  This looks like a new variable that you're adding so a few things.  

1.  If you can update this PR with a general description of what you're trying to do here.
2.  If it is a new variable, update the appropriate role docs with it
3.  Add a change fragment describing the update.

tobstone:
> @tobstone thanks for the contribution. This looks like a new variable that you're adding so a few things.
> 
> 1. If you can update this PR with a general description of what you're trying to do here.
> 2. If it is a new variable, update the appropriate role docs with it
> 3. Add a change fragment describing the update.

  When MySQL is configured to listen on a Unix socket (the default on many Linux installations), the community.mysql modules require login_unix_socket to connect locally. Without it, both the schema-version check and the      
  schema import tasks fail because they attempt a TCP connection instead.                                                                                                                                                         
                                                                                                                                                                                                                                  
  This adds an optional zabbix_server_mysql_login_unix_socket variable to those two tasks. It defaults to omit, so existing deployments using TCP are unaffected.

PR #1740, opened 2026-05-11
Added options to allow for custom content in zabbix.conf.php

### LINKED ISSUE #1563: SSO requestedAuthnContext security setting option in /etc/zabbix/web/zabbix.conf.php
##### SUMMARY
We have an issue with Azure Entra ID and Zabbix SSO settings that requires a custom setting in "/etc/zabbix/web/zabbix.conf.php".  Failing that "/usr/share/zabbix/vendor/onelogin/php-saml/src/Saml2/AuthnRequest.php" can be modified but that is very undesirable.  Currently the "community.zabbix.zabbix_web" role sets the "/etc/zabbix/web/zabbix.conf.php" file but does not include settings for "requestedAuthnContext".

Barring specific config to solve this issue, as outlined at the link at the bottom of this post, would it be possible to add a generic `zabbix_conf_php_custom` variable that adds it's content at the end of the template file? 

[https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_web/templates/zabbix.conf.php.j2](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_web/templates/zabbix.conf.php.j2)

##### Example Code

End of "zabbix.conf.php.j2":
```jinja2
{% if zabbix_conf_php_custom | default('') | length > 0 %}
{{ zabbix_conf_php_custom }}
{% endif %}
```

User provided host_var:
```yaml
zabbix_conf_php_custom: |
  $SSO['SETTINGS'] = ['requestedAuthnContext' => false];
```

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
community.zabbix.zabbix_web

##### ADDITIONAL INFORMATION
[https://www.zabbix.com/forum/zabbix-help/476520-zaabix-saml-sign-in-problem?p=503169#post503169](https://www.zabbix.com/forum/zabbix-help/476520-zaabix-saml-sign-in-problem?p=503169#post503169)

and

[https://www.zabbix.com/forum/zabbix-troubleshooting-and-problems/454003-solution-to-azure-saml-please-rectify-in-code](https://www.zabbix.com/forum/zabbix-troubleshooting-and-problems/454003-solution-to-azure-saml-please-rectify-in-code)

InvisibleDuck:
##### SUMMARY
Added option to allow for the addition of raw PHP config in zabbix.conf.php.
Usage example:
```yaml
zabbix_web_custom_zabbix_conf_php: |
  $SSO['SETTINGS'] = [
      'use_proxy_headers'  => true,
      'security' => [
          'requestedAuthnContext' => [
              'urn:oasis:names:tc:SAML:2.0:ac:classes:unspecified',
          ],
      ]
  ];
```
Actually fixes #1563 because #1570 did not.

##### ISSUE TYPE
- Bugfix/Feature Pull Request

##### COMPONENT NAME
Web role

##### ADDITIONAL INFORMATION
There may be a need for very specific options that is not worth/possible to add using mapped variables directly. I.e. in the case of SSO.

### DIFF
--- a/roles/zabbix_web/templates/zabbix.conf.php.j2
+++ b/roles/zabbix_web/templates/zabbix.conf.php.j2
@@ -65,7 +65,7 @@ $SSO['SP_KEY'] = '{{ zabbix_saml_sp_key }}';
 {% endif %}
 {% endif %}
 
-{% if zabbix_web_custom_zabbix_conf_php | default('') | length > 0 %}
+{% if zabbix_web_custom_zabbix_conf_php | length > 0 %}
 {{ zabbix_web_custom_zabbix_conf_php }}
 {% endif %}
 ?>
--- a/roles/zabbix_web/templates/zabbix.conf.php.j2
+++ b/roles/zabbix_web/templates/zabbix.conf.php.j2
@@ -64,4 +64,8 @@ $SSO['SP_KEY'] = '{{ zabbix_saml_sp_key }}';
 {{ (zabbix_server_tls_certificate_subject is defined and zabbix_server_tls_certificate_subject is not none) | ternary('', '// ') }}$ZBX_SERVER_TLS['CERTIFICATE_SUBJECT'] = '{{ zabbix_server_tls_certificate_subject | default('') }}';
 {% endif %}
 {% endif %}
+
+{% if zabbix_web_custom_zabbix_conf_php | default('') | length > 0 %}
+{{ zabbix_web_custom_zabbix_conf_php }}
+{% endif %}
 ?>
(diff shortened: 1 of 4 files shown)


roles/zabbix_web/templates/zabbix.conf.php.j2
```diff
@@ -64,4 +64,8 @@ $SSO['SP_KEY'] = '{{ zabbix_saml_sp_key }}';
 {{ (zabbix_server_tls_certificate_subject is defined and zabbix_server_tls_certificate_subject is not none) | ternary('', '// ') }}$ZBX_SERVER_TLS['CERTIFICATE_SUBJECT'] = '{{ zabbix_server_tls_certificate_subject | default('') }}';
 {% endif %}
 {% endif %}
+
+{% if zabbix_web_custom_zabbix_conf_php | default('') | length > 0 %}
```
eb4x: Put the default into `defaults/main.yml`

eb4x (changes requested):
Yes! I absolutely want this. Can you add a changelog fragment?

@pyrodie18 for 8.0, what do you say we strip zabbix.conf.php down to the bare minimum and let users pile on their own with this. (We could start out by removing saml, the tls stuff and maybe the zabbix_web_env? (I haven't looked into what comes standard, just a thought that occurred.)

eb4x (approved):
Thank you ❤️

PR #1752, opened 2026-05-31
Remove zabbix60

### LINKED ISSUE #1712: Remove Support for Zabbix 6.0

pyrodie18:
##### SUMMARY
This removes all support for versions of Zabbix < 7.0.  This closes #1712


plugins/modules/zabbix_authentication.py
```diff
@@ -496,123 +243,50 @@ def update_authentication(
             if isinstance(http_case_sensitive, bool):
                 params["http_case_sensitive"] = str(int(http_case_sensitive))
 
-            if LooseVersion(self._zbx_api_version) < LooseVersion("7.0"):
-                if isinstance(ldap_configured, bool):
-                    params["ldap_configured"] = str(int(ldap_configured))
             else:
                 if isinstance(ldap_auth_enabled, bool):
                     params["ldap_auth_enabled"] = str(int(ldap_auth_enabled))
```
eb4x: This else clause belonged to the `LooseVersion(self._zbx_api_version) < LooseVersion("7.0"):` just removed and probably shouldn't be apart of the isinstance it now currently is?

plugins/modules/zabbix_user.py
```diff
@@ -566,7 +565,7 @@ def add_user(
             "url": url,
         }
         if user_medias:
-            if LooseVersion(self._zbx_api_version) <= LooseVersion("7.0"):
+            if LooseVersion(self._zbx_api_version) == LooseVersion("7.0"):
                 request_data["user_medias"] = user_medias
             else:
                 request_data["medias"] = user_medias
```
eb4x: This would be better if written as >= 7.2, but that's beside the thing i spotted. Which I'll get to now;

So we're saying zabbix 7.0 **still** operates on `request_data["user_medias"]`, but future versions will operate on just `request_data["medias"]` and yet, in  the logic below we're saying 

```
... < LooseVersion("7.0"):
  try:
    if user_medias:
      request_data["user_medias"] = user_medias
...
...  >= LooseVersion("7.0"):
  try:
    if user_medias:
      request_data["medias"] = user_medias
```

So something is inconsistent here.

plugins/modules/zabbix_user.py
```diff
@@ -644,25 +643,14 @@ def update_user(
...
-                    msg="Failed to update user %s: %s" % (username, e)
-                )
+        try:
+            if user_medias:
+                request_data["medias"] = user_medias
+            user_ids = self._zapi.user.update(request_data)
+        except Exception as e:
+            self._module.fail_json(
+                msg="Failed to update user %s: %s" % (username, e)
+            )
```
eb4x, line 636: See comment about which versions use `user_medias` vs just `medias`

plugins/modules/zabbix_user_directory.py
```diff
@@ -487,19 +438,16 @@ def main():
...
-    """ For future when < 7.0 disappears we should use this, now we cannot do this as at this point Zabbix version is unknown
     module = AnsibleModule(
         argument_spec=argument_spec,
         supports_check_mode=True,
         required_if=[
             ("state", "present", ("idp_type",)),
             ("idp_type", "ldap", ("host", "port", "base_dn", "search_attribute"), False),
             ("idp_type", "saml", ("idp_entityid", "sp_entityid", "sso_url", "username_attribute"), False),
-            ("provision_status", "true", ("provision_groups"))
+            ("provision_status", "true", ("provision_groups",))
```
eb4x, line 448: is the comma introduced on purpose?

roles/zabbix_repo/vars/RedHat.yml
eb4x, line 1: We need an empty file atleast, unless you fix the `include_vars:` in `zabbix_repo/tasks/main.yml` to ignore missing files.

roles/zabbix_server/tasks/initialize-pgsql.yml
```diff
@@ -201,16 +201,15 @@
...
         login_host: "{{ zabbix_server_dbhost }}"
         login_port: "{{ zabbix_server_dbport }}"
         db: "{{ zabbix_server_dbname }}"
         state: restore
-        target: "{{ _zabbix_server_database_scripts_dir }}/{{ _is_pre70 | ternary(legacy_path, modern_path) }}"
+        target: "{{ _zabbix_server_database_scripts_dir }}/{{ modern_path }}"
       vars:
-        _is_pre70: "{{ zabbix_server_version is version('7.0', '<') }}"
         legacy_path: postgresql/timescaledb.sql
         modern_path: postgresql/timescaledb/schema.sql
```
eb4x, line 215: You can probably remove one and inline the other.

tests/integration/targets/test_zabbix_regexp/tasks/main.yml
```diff
@@ -3,24 +3,7 @@
   meta: end_play
   when: zabbix_version is version('4.0', '<')
```
eb4x: You can take this out with the rest of the 🗑️

eb4x (changes requested):
I spotted a couple of things, the most curious one being `user_medias` vs `medias`.

There was raised a concern about zabbix 6.0 lts not going away for some time still, but the amount of duplicated garbage going away here makes me want to tell 6.0 users to just use an earlier collection version.

Great job 👍

plugins/modules/zabbix_user_directory.py
pyrodie18: yes, it actually has to be an iterable (failed initially without it)

PR #1753, opened 2026-06-01
New zabbix_agent2_plugins variable for additional plugin packages

### LINKED ISSUE #1091: Add task to install zabbix-agent2-plugin-* (or a subset)
##### SUMMARY

Currently the role can install ```zabbix-agent(2)```, ```zabbix-get``` and ```zabbix-sender```. But ```zabbix-agent2``` comes with a few extra plugins (only postgresql and mongodb) for now.

Can the ```zabbix_agent2_plugins``` variable be extended to also support the installation of additional packages? Something along the lines of:

```
zabbix_agent2_plugins:
  - name: PostgreSQL
    package: zabbix-agent2-plugin-postgres
```

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

zabbix_agent
##### ADDITIONAL INFORMATION

```yaml

```

CompPhy:
##### SUMMARY
Adds support for installing Zabbix Agent 2 plugin packages by introducing `zabbix_agent_packages`
— a list variable that replaces the scalar `zabbix_agent_package`.

The existing `zabbix_agent_package` variable is deprecated but continues to work via a
double-nested default in `zabbix_agent_packages`, so existing playbooks require no changes.
A deprecation warning is emitted via `tasks/deprecation_checks.yml` when the old variable
is detected.

Fixes #1091

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zabbix_agent

##### ADDITIONAL INFORMATION
`zabbix_agent_packages` defaults to a single-item list matching the previous `zabbix_agent_package`
behavior, so there is no change for existing users.

To install agent2 plugin packages alongside the agent:

```yaml
# Debian/Ubuntu
zabbix_agent_packages:
  - zabbix-agent2
  - zabbix-agent2-plugin-mongodb

# RedHat — include version suffix to stay consistent with the agent package
zabbix_agent_packages:
  - "zabbix-agent2-{{ zabbix_agent_version }}.{{ zabbix_agent_version_minor }}"
  - "zabbix-agent2-plugin-mongodb-{{ zabbix_agent_version }}.{{ zabbix_agent_version_minor }}"

### DIFF
--- a/docs/ZABBIX_AGENT_ROLE.md
+++ b/docs/ZABBIX_AGENT_ROLE.md
@@ -343,6 +343,7 @@ The following table lists all variables that are exposed to modify the configura
 | PersistentBufferPeriod | zabbix_agent_persistentbufferperiod | 1h | Agent 2 Only |
 | PidFile | zabbix_agent_pidfile | /var/run/zabbix/`{{ agent version specific }}`.pid | Linux Systems Only |
 | Plugin | zabbix_agent_plugins |  |  |
+| — | zabbix_agent2_plugins |  | List of agent2 plugin packages to install. Agent 2 Only. Undefined by default. |
 | PluginSocket | zabbix_agent_pluginsocket | /tmp/agent.plugin.sock | Agent 2 Only  |
 | PluginTimeout | zabbix_agent_plugintimeout | 3 | Agent 2 Only |
 | RefreshActiveChecks | zabbix_agent_refreshactivechecks |  |  |
--- a/docs/ZABBIX_AGENT_ROLE.md
+++ b/docs/ZABBIX_AGENT_ROLE.md
@@ -143,7 +143,19 @@ Selinux changes will be installed based on the status of selinux running on the 
 * `zabbix_agent_install_agent_only`: Only install the Zabbix Agent and not the `zabbix-sender` and `zabbix-get` packages. Default: `False`
 * `zabbix_agent_package_remove`: If `zabbix_agent2: True` and you want to remove the old installation. Default: `False`.
 * `zabbix_agent_package_state`: If Zabbix-agent needs to be `present` (default) or `latest`.
-* `zabbix_agent_package`: The name of the zabbix-agent package. Default: `zabbix-agent` if `zabbix_agent2` is false and `zabbix-agent2` if `true`.
+* `zabbix_agent_packages`: List of packages to install. Defaults to `[zabbix-agent]` or `[zabbix-agent2]` depending on `zabbix_agent2`. Use this to add agent2 plugin packages:
+  ```yaml
+  # Debian/Ubuntu
+  zabbix_agent_packages:
+    - zabbix-agent2
+    - zabbix-agent2-plugin-mongodb
+
+  # RedHat — include version suffix to stay consistent with the agent package
+  zabbix_agent_packages:
+    - "zabbix-agent2-{{ zabbix_agent_version }}.{{ zabbix_agent_version_minor }}"
+    - "zabbix-agent2-plugin-mongodb-{{ zabbix_agent_version }}.{{ zabbix_agent_version_minor }}"
+  ```
+* `zabbix_agent_package`: **Deprecated.** Use `zabbix_agent_packages` instead. The name of the zabbix-agent package. Default: `zabbix-agent` if `zabbix_agent2` is false and `zabbix-agent2` if `true`.
 * `zabbix_agent_sender_package`: The name of the zabbix-sender package. Default: `zabbix-sender`.
 * `zabbix_agent_userparameters`: Default: `[]`. List of userparameter names and scripts (if any). Detailed description is given in the [Deploying Userparameters](#deploying-userparameters) section.
   * `name`: Userparameter name (should be the same with userparameter template file name)
@@ -343,7 +355,6 @@ The following table lists all variables that are exposed to modify the configura
 | PersistentBufferPeriod | zabbix_agent_persistentbufferperiod | 1h | Agent 2 Only |
 | PidFile | zabbix_agent_pidfile | /var/run/zabbix/`{{ agent version specific }}`.pid | Linux Systems Only |
 | Plugin | zabbix_agent_plugins |  |  |
-| — | zabbix_agent2_plugins |  | List of agent2 plugin packages to install. Agent 2 Only. Undefined by default. |
 | PluginSocket | zabbix_agent_pluginsocket | /tmp/agent.plugin.sock | Agent 2 Only  |
 | PluginTimeout | zabbix_agent_plugintimeout | 3 | Agent 2 Only |
 | RefreshActiveChecks | zabbix_agent_refreshactivechecks |  |  |
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -8,10 +8,8 @@ zabbix_agent_releases_url: "https://services.zabbix.com/updates/v1"
 zabbix_agent_version_minor: "*"
 #zabbix_agent_version_long: 7.4.2
 zabbix_agent_package_remove: false
-# Deprecated: use zabbix_agent_packages (list) instead
-zabbix_agent_package: "{{ _zabbix_agent_package | default(zabbix_agent2 | ternary('zabbix-agent2', 'zabbix-agent')) }}"
 zabbix_agent_packages:
-  - "{{ zabbix_agent_package }}"
+  - "{{ zabbix_agent_package | default(_zabbix_agent_package | default(zabbix_agent2 | ternary('zabbix-agent2', 'zabbix-agent'))) }}"
 zabbix_agent_package_state: present
 zabbix_agent_get_package: "{{ _zabbix_agent_get_package | default('zabbix-get') }}"
 zabbix_agent_sender_package: "{{ _zabbix_agent_sender_package | default('zabbix-sender') }}"
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -10,7 +10,8 @@ zabbix_agent_version_minor: "*"
 zabbix_agent_package_remove: false
 # Deprecated: use zabbix_agent_packages (list) instead
 zabbix_agent_package: "{{ _zabbix_agent_package | default(zabbix_agent2 | ternary('zabbix-agent2', 'zabbix-agent')) }}"
-zabbix_agent_packages: "{{ [zabbix_agent_package] }}"
+zabbix_agent_packages:
+  - "{{ zabbix_agent_package }}"
 zabbix_agent_package_state: present
 zabbix_agent_get_package: "{{ _zabbix_agent_get_package | default('zabbix-get') }}"
 zabbix_agent_sender_package: "{{ _zabbix_agent_sender_package | default('zabbix-sender') }}"
--- a/roles/zabbix_agent/defaults/main.yml
+++ b/roles/zabbix_agent/defaults/main.yml
@@ -8,7 +8,9 @@ zabbix_agent_releases_url: "https://services.zabbix.com/updates/v1"
 zabbix_agent_version_minor: "*"
 #zabbix_agent_version_long: 7.4.2
 zabbix_agent_package_remove: false
+# Deprecated: use zabbix_agent_packages (list) instead
 zabbix_agent_package: "{{ _zabbix_agent_package | default(zabbix_agent2 | ternary('zabbix-agent2', 'zabbix-agent')) }}"
+zabbix_agent_packages: "{{ [zabbix_agent_package] }}"
 zabbix_agent_package_state: present
 zabbix_agent_get_package: "{{ _zabbix_agent_get_package | default('zabbix-get') }}"
 zabbix_agent_sender_package: "{{ _zabbix_agent_sender_package | default('zabbix-sender') }}"
--- a/roles/zabbix_agent/tasks/main.yml
+++ b/roles/zabbix_agent/tasks/main.yml
@@ -67,6 +67,14 @@
   loop_control:
     loop_var: _install_tasks_file
 
+- name: "Install zabbix_agent2_plugins"
+  ansible.builtin.package:
+    name: "{{ zabbix_agent2_plugins }}"
+    state: "{{ zabbix_agent_package_state }}"
+    update_cache: true
+    disablerepo: "{{ zabbix_agent_disable_repo | default(_zabbix_agent_disable_repo | default(omit)) }}"
+  when: zabbix_agent2_plugins is defined
+
 - name: "Configure Agent"
   when: ansible_facts['os_family'] not in ['Darwin']
   ansible.builtin.include_tasks: "{{ _configure_tasks_file }}"
--- a/roles/zabbix_agent/tasks/main.yml
+++ b/roles/zabbix_agent/tasks/main.yml
@@ -52,7 +52,7 @@
     name: community.zabbix.zabbix_repo
   vars:
     zabbix_repo_version: "{{ zabbix_agent_version }}"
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: roles/zabbix_agent/tasks/main.yml [hunk 1]
```diff
@@ -67,6 +67,14 @@
...
     loop_var: _install_tasks_file
 
+- name: "Install zabbix_agent2_plugins"
+  ansible.builtin.package:
+    name: "{{ zabbix_agent2_plugins }}"
+    state: "{{ zabbix_agent_package_state }}"
+    update_cache: true
+    disablerepo: "{{ zabbix_agent_disable_repo | default(_zabbix_agent_disable_repo | default(omit)) }}"
+  when: zabbix_agent2_plugins is defined
+
```
eb4x: This won't go well over when a Windows, Darwin (and possibly FreeBSD?) runs the playbook.

I'm also not certain we want to install these plugins by default? (I personally install them out of convenience) But I'm all for sane defaults, if they're not a liability somehow.

Anyway, this should probably just be tacked on the task that already exists in `install.yml`.

eb4x (changes requested):

roles/zabbix_agent/tasks/main.yml
CompPhy: It's undefined by default, unless you know you need it.... Hence:  `when: zabbix_agent2_plugins is defined`

Also, it's already broken as is... The templates for the configuration file allow you to include plugin configs.  However, if you can't also make sure the dependencies for that config are installed; then it just breaks on restart.

eb4x: The user supplying custom configuration is more of a foot-gun than broken though. I mean, we do have agent2 deployment tests passing with the "default config" right?

But yeah, the roles/agent code is still in bit of a sorry state :melting_face:, and from the looks of it  there's gonna be more plugins with 8.0.

`is defined` is a bit of a code smell for ansible (yes, there's a lot of this bad practice in our code already)

How about we defaut it to an empty list

```yaml
zabbix_agent2_plugins: []
```

and just tack it to the "Installing zabbix-..." task in `install.yml`

```yaml
- name: "Installing zabbix-{agent,get,sender}"
...
  loop:
    - name: "{{ zabbix_agent_package }}"
      state: "{{ zabbix_agent_package_state }}"
    - name: "{{ zabbix_agent2_plugins }}"
      state: "{{ zabbix_agent_package_state }}"
    - name: "{{ zabbix_agent_get_package }}"
...
```

What do you think? (Should we be checking that agent2 is also getting installed? Or look for some plugin activation in the code?)

eb4x: Actually, I changed my mind. `zabbix_agent2_plugins` is not descriptive enough, because we're actually expecting a list of packages to install.

If anything, I think we could introduce a variable `zabbix_agent_packages` and have it be a list of one.

```yaml
zabbix_agent_packages:
  - "{{ zabbix_agent_package }}"
```

It's (probably?) more intuitively named to override with multiple packages than the `zabbix_agent_package`.

Because if you know how to override the empty list of plugins, you're probably competent enough to override a list of `zabbix_agent_packages` remembering to keep the main package in addition to any plugins you may want/need. (which is something you can already do by making `zabbix_agent_package` a list of packages)

And now I've come full circle to maybe this doesn't require a change at all, maybe it's just a documentation issue?

CompPhy: Following your idea... did I just find more code smell..... shouldn't `zabbix_agent2` be in quoted in the variable below.  This looks to me like it's mixing up the boolean variable zabbix_agent2 with the string "zabbix_agent2".  What's worst, if I do a search for "zabbix_agent2" I find all kinds of occurrences like this...

https://github.com/ansible-collections/community.zabbix/blob/e95b2f3a1d906ca508d181e0f71619cae1bcb4ca/roles/zabbix_agent/defaults/main.yml#L11

I would think we need quotes around the string zabbix_agent2 here:
```yaml
zabbix_agent_package: "{{ _zabbix_agent_package | default('zabbix_agent2' | ternary('zabbix-agent2', 'zabbix-agent')) }}"
```

Happy to follow your thinking and change this PR, but now I'm worried this whole thing is somehow broken.  Even though it's something leading to the right result; these default variables look sus to me.  Maybe I'm just missing something here???

Maybe someone intended to set zabbix_agent2 variable to the package name string, instead of a boolean, but they missed updating the boolean???

CompPhy: NM... ignore me... I'm forgetting how the ternary operator works.  I see what you're trying to get at with the `zabbix_agent_package` variable here...  Let me see about updating the PR so it's just a documentation change.

CompPhy: There is a snag to this logic.... This line would break if we try to pass a list:

https://github.com/ansible-collections/community.zabbix/blob/e95b2f3a1d906ca508d181e0f71619cae1bcb4ca/roles/zabbix_agent/tasks/main.yml#L55

We could do a flatten | first on the list, but then that also assumes the user is passing the list in a specific order...

CompPhy: Maybe another option is to keep the zabbix_agent2_plugins variable but call it through the loop of install.yml?   This could still work with an empty list, so we're not relying on `is defined`.

eb4x: No, that code is correct.

```yaml
zabbix_agent_package: "{{ _zabbix_agent_package | default(zabbix_agent2 | ternary('zabbix-agent2', 'zabbix-agent')) }}"
```

If we have a override (usually prefixed with `_` like this) in the `vars/*` (usually distribution specific) use that, (freebsd is an oddball)
else check what the [`zabbix_agent2`](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L6) boolean flag is, and set to either `'zabbix-agent2'` or `'zabbix-agent'`. You'll note the official package names are inside the `ternary` filter, quoted and with a `-` in them.

We do the same for `zabbix_agent_service` where both alpine and freebsd stand out from the official packaged version.

I'd approve a change to `zabbix_agent_packages` with the additional documentation and examples if you think it's worth it. Just remember to apply a `| flatten` on it in the package install task, as some folks might already have turned `zabbix_agent_package` to a list and we don't want to break things for them.

eb4x: > There is a snag to this logic.... This line would break if we try to pass a list:
> 
> https://github.com/ansible-collections/community.zabbix/blob/e95b2f3a1d906ca508d181e0f71619cae1bcb4ca/roles/zabbix_agent/tasks/main.yml#L55
> 
> We could do a flatten | first on the list, but then that also assumes the user is passing the list in a specific order...

That's a good catch :+1: , it would only prevent debian/ubuntu users from successfully overriding the `zabbix_agent_package` with a list. But a `flatten` and `join` filter would make that section work.

eb4x: > Maybe another option is to keep the zabbix_agent2_plugins variable but call it through the loop of install.yml? This could still work with an empty list, so we're not relying on `is defined`.

Yeah, it's just not a great name for what it's used for though (package names installation).  And we have enough [bad](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L20) [and](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L10) [redundant](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L115) [variables](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L35) [that](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L66) [need](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L114) [to](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L125) [go](https://github.com/ansible-collections/community.zabbix/blob/main/roles/zabbix_agent/defaults/main.yml#L9).

Ideally we'd just deprecate `zabbix_agent_package` for `zabbix_agent_packages`

CompPhy:
Have a look at the new commit and let me know your thoughts.

roles/zabbix_agent/defaults/main.yml
```diff
@@ -8,7 +8,9 @@ zabbix_agent_releases_url: "https://services.zabbix.com/updates/v1"
 zabbix_agent_version_minor: "*"
 #zabbix_agent_version_long: 7.4.2
 zabbix_agent_package_remove: false
+# Deprecated: use zabbix_agent_packages (list) instead
 zabbix_agent_package: "{{ _zabbix_agent_package | default(zabbix_agent2 | ternary('zabbix-agent2', 'zabbix-agent')) }}"
+zabbix_agent_packages: "{{ [zabbix_agent_package] }}"
```
eb4x: Technically yours is valid, but it's not illustrating that it's a list. Better would be

`zabbix_agent_packages: [ "{{ zabbix_agent_package }}", ... ]`

And formatted like this it's glaringly obvious

```yaml
zabbix_agent_packages:
  - "{{ zabbix_agent_package }}"
```

docs/ZABBIX_AGENT_ROLE.md
```diff
@@ -143,7 +143,19 @@ Selinux changes will be installed based on the status of selinux running on the
...
+  ```yaml
+  # Debian/Ubuntu
+  zabbix_agent_packages:
+    - zabbix-agent2
+    - zabbix-agent2-plugin-mongodb
+
+  # RedHat — include version suffix to stay consistent with the agent package
+  zabbix_agent_packages:
+    - "zabbix-agent2-{{ zabbix_agent_version }}.{{ zabbix_agent_version_minor }}"
+    - "zabbix-agent2-plugin-mongodb-{{ zabbix_agent_version }}.{{ zabbix_agent_version_minor }}"
```
eb4x, line 156: I think you can omit the redhat specifics. Minor is `*` by default which means get the latest minor at time of installation.

What are the odds a user find this foot-gun at a later date and manage to get in a state where an agent2 version of 7.0.8 run with a newer plugin of version 7.0.14 , **and** there's an actual conflict between these two binaries? (He said, realizing that **that's** absolutely what's gonna happen)

roles/zabbix_agent/defaults/main.yml
```diff
@@ -8,7 +8,9 @@ zabbix_agent_releases_url: "https://services.zabbix.com/updates/v1"
 zabbix_agent_version_minor: "*"
 #zabbix_agent_version_long: 7.4.2
 zabbix_agent_package_remove: false
+# Deprecated: use zabbix_agent_packages (list) instead
 zabbix_agent_package: "{{ _zabbix_agent_package | default(zabbix_agent2 | ternary('zabbix-agent2', 'zabbix-agent')) }}"
```
eb4x: Deprecation would require some double-nested defaults, and a write up in `tasks/deprecation_checks.yml`.

```yaml
zabbix_agent_packages:
  - "{{ zabbix_agent_package | default( _zabbix_agent_package | default(zabbix_agent2 | ternary('zabbix-agent2', 'zabbix-agent'))) }}"
```

By removing `zabbix_agent_package` from `defaults/main.yml`, but preserving it in `zabbix_agent_packages` we can now detect if someone actually provides an override for this, and warn in `tasks/deprecation_checks.yml` without breaking anything 🤞

roles/zabbix_agent/tasks/main.yml
```diff
@@ -52,7 +52,7 @@
     name: community.zabbix.zabbix_repo
   vars:
     zabbix_repo_version: "{{ zabbix_agent_version }}"
-    zabbix_repo_package: "{{ zabbix_agent_package }} zabbix-get zabbix-sender"
+    zabbix_repo_package: "{{ zabbix_agent_packages | flatten | first }} zabbix-get zabbix-sender"
```
eb4x: `| first` will just pick the first element in the list, so the plugins would be omitted. You'll have to use `| join(' ')` instead.

roles/zabbix_agent/tasks/remove.yml
```diff
@@ -16,7 +16,7 @@
 
 - name: "Remove | Package removal"
   ansible.builtin.package:
-    name: "{{ zabbix_agent_package | replace('agent2', 'agent') }}"
+    name: "{{ zabbix_agent_packages | flatten | first | replace('agent2', 'agent') }}"
```
eb4x: This kinda works, but only if the first package is always the main agent package.

Man this task is so dumb, it's only ever included when `zabbix_agent2` and `zabbix_agent_package_remove` 🫠 There's no opposing logic here, so it's only ever for uninstalling the old agent. Like what else can be achieved by replacing 'agent2' with 'agent' in the package name. 🤯   

You can just simplify this to the package name `zabbix-agent`

eb4x (changes requested):
Overall good, I think with the comments addressed this is good to go 👍 

Sorry 'bout the rant on the remove.yml, it's directed at the old code, not your changes ❤️

docs/ZABBIX_AGENT_ROLE.md
CompPhy: LOL... Yeah, I live in the RedHat world... I think the reminder is needed here so someone doesn't shoot themselves in the foot later.

CompPhy:
I just updated the PR description as well. :wink:

eb4x (approved):
LGTM 🚀

eb4x:
Ok, so the reason it's failing on suse is because of

```
Jun 03 07:02:28 450d9f0dfdd1 systemd[1]: Starting Zabbix Agent 2...
Jun 03 07:02:28 450d9f0dfdd1 zabbix_agent2[1344]: Validating configuration file "/etc/zabbix/zabbix_agent2.conf"
Jun 03 07:02:28 450d9f0dfdd1 zabbix_agent2[1344]: zabbix_agent2 [1344]: ERROR: Cannot register plugins: failed to register metrics of plugin "MongoDB": failed to start plugin: failed to start plugin process "/usr/sbin/zabbix-agent2-plugin/zabbix-agent2-plugin-mongodb": fork/exec /u>
Jun 03 07:02:28 450d9f0dfdd1 systemd[1]: zabbix-agent2.service: Control process exited, code=exited, status=1/FAILURE
Jun 03 07:02:28 450d9f0dfdd1 systemd[1]: zabbix-agent2.service: Failed with result 'exit-code'.
Jun 03 07:02:28 450d9f0dfdd1 systemd[1]: Failed to start Zabbix Agent 2.
```

and if we take a look

```shell
450d9f0dfdd1:/ # ls /usr/sbin/zabbix*
/usr/sbin/zabbix_agent2
```

```shell
450d9f0dfdd1:/ # rpm -ql zabbix-agent2-plugin-mongodb
/etc/zabbix/zabbix_agent2.d/plugins.d/mongodb.conf
/usr/lib/zabbix/zabbix-agent2-plugin-mongodb
/usr/share/doc/packages/zabbix-agent2-plugin-mongodb
/usr/share/doc/packages/zabbix-agent2-plugin-mongodb/LICENSE
/usr/share/doc/packages/zabbix-agent2-plugin-mongodb/README.md
```

which means this package is shipping with broken plugin configuration, where they configure to look for the plugin at `/usr/sbin/zabbix-agent2-plugin/zabbix-agent2-plugin-mongodb`, but install it at `/usr/lib/zabbix/zabbix-agent2-plugin-mongodb`

I'll look up if this has been reported or create a issue for it and link it here before merging.

https://support.zabbix.com/browse/ZBX-27859

eb4x:
Thank you :heart: 

(The failing Suse test is a recent and known issue, https://support.zabbix.com/browse/ZBX-27841 that will get fixed eventually.)

PR #1758, opened 2026-07-06
fix: zabbix_trigger/triggerprototype idempotence - compare only managed fields

tolgakaratas:
##### SUMMARY
`check_trigger_changed` compared the full old vs new trigger dict returned by `trigger.get`. Zabbix auto-populates some fields the module does not manage (for example `event_name`, `url_name`) during `trigger.update`, so the full-dict comparison always reported a change and the module was never idempotent. This compares only the keys the task manages instead. The same fix is applied to `zabbix_triggerprototype`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/zabbix_trigger.py, plugins/modules/zabbix_triggerprototype.py

##### ADDITIONAL INFORMATION
Both dicts being compared come from `trigger.get`, so narrowing the comparison to the managed keys introduces no int/str or expression-normalization skew; it only stops the unmanaged, server-populated fields from producing a spurious `changed`. The modules are exercised by the `plugins-integration` suite, and sanity is green (see the validation note in the comments).

### DIFF
--- a/changelogs/fragments/zabbix_trigger-idempotence.yml
+++ b/changelogs/fragments/zabbix_trigger-idempotence.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zabbix_trigger, zabbix_triggerprototype - fix idempotence. The modules compared the whole trigger object before and after calling ``trigger.update``; when Zabbix auto-populates an unmanaged field (for example ``event_name``) during the update, that comparison reported a spurious ``changed=true``. They now compare only the fields the module manages.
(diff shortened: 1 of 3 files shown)


tolgakaratas:
For transparency: this change was developed and validated as part of a larger modernization of the collection's test suite, then split into focused, single-logical-change PRs per CONTRIBUTING. The whole set was validated green together on our fork's GitHub Actions before splitting.

- repo-sanity (sanity + changelog fragments) green: https://github.com/Denomas/community.zabbix/actions/runs/28761209140

Note on the `Sanity (Ⓐdevel+py3.12)` job: it currently fails at the dependency-install step with `ansible-core requires a different Python: 3.12 not in '>=3.13'`. That is a pre-existing issue (ansible-core `devel` raised its Python floor to 3.13) unrelated to this change - the fix never runs. It is resolved by the companion test-suite modernization PR, which moves `devel` to a supported Python.

tolgakaratas:

Issue #1745, opened 2026-05-15
Support Ubuntu 26.04 LTS (Resolute Raccoon)


Japje:
##### SUMMARY
Support for the newly released Ubuntu 26.06.
The repository of zabbix has builds  for this already.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
zabbix-agent

##### ADDITIONAL INFORMATION
```
[ERROR]: Task failed: Error while evaluating conditional: object of type 'dict' has no attribute '26'

```

tolgakaratas:
Ubuntu 26.04 (Resolute Raccoon) support is addressed by two open PRs:

- #1759 — adds Ubuntu 26.04 to the role supported-version maps (and reconciles those maps with real Zabbix package availability per distribution).
- #1760 — fixes the `zabbix_web` role on modern Debian/Ubuntu Apache. On Ubuntu 26.04 the `zabbix-apache-conf` package serves PHP via php-fpm (`SetHandler "proxy:unix:/var/run/php/zabbix.sock|fcgi://localhost"`) instead of mod_php, but the role did not create the php-fpm pool, so the frontend returned HTTP 503. The role now creates php-fpm when the enabled Apache config references the socket it manages; distributions whose packaged config uses mod_php are left unchanged.

Both were validated against a real Zabbix via molecule (apache, mysql, Zabbix 7.0/7.4) on Ubuntu 26.04 on a fork's CI.

PR #1760, opened 2026-07-06
fix: support modern Debian/Ubuntu apache (php-fpm + proxy modules) in zabbix_web

tolgakaratas:
##### SUMMARY
The `zabbix_web` role assumed the Apache frontend always uses mod_php on the Debian family, which is no longer true on newer distributions:

- The packaged Apache config uses `RewriteEngine` and, on Zabbix 7.0+, `ProxyTimeout`, so `apache2ctl configtest` failed with "Invalid command 'ProxyTimeout'" unless `mod_proxy`, `mod_proxy_fcgi` and `mod_rewrite` are enabled. The role now enables them (proxy before the already-enabled rewrite so the `apache2ctl -M` verification stays valid).
- On newer distributions (for example Ubuntu 26.04) the `zabbix-apache-conf` package serves PHP via php-fpm (`SetHandler "proxy:unix:/var/run/php/zabbix.sock|fcgi://localhost"`) instead of mod_php, but the role skipped php-fpm for Apache, so the socket was never created and the frontend returned HTTP 503. The role now creates php-fpm when the enabled Apache config references the socket it manages; distributions whose packaged config uses mod_php are untouched.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
roles/zabbix_web (tasks/apache.yml, tasks/main.yml)

##### ADDITIONAL INFORMATION
Verified with molecule (apache): Ubuntu 26.04 now serves HTTP 200 and the API-login verify passes (was HTTP 503); Ubuntu 24.04 and Debian 12 (which use mod_php) stay green including idempotence, with the php-fpm autodetection correctly not firing. Green across the role matrix (amd64 + arm64, mysql + pgsql, Zabbix 7.0/7.4) on our fork; see the validation note in the comments. The change also makes the molecule verify wait for the Zabbix backend to finish seeding the database before the API-login assertion (a test-only timing fix). Changelog fragments included. Refs #1745.


tolgakaratas:
For transparency: this change was developed and validated as part of a larger test-suite modernization, then split into focused, single-logical-change PRs per CONTRIBUTING. The whole set was validated green together on our fork's GitHub Actions before splitting.

- zabbix_web role matrix green (390/390 jobs: amd64 + arm64, mysql + pgsql, Zabbix 7.0/7.4, incl. Ubuntu 26.04 apache which was the failing case): https://github.com/Denomas/community.zabbix/actions/runs/28761209178

Note: on branches based on the current upstream `main`, the `Sanity (Ⓐdevel+py3.12)` job fails at dependency install (`ansible-core devel` now requires Python >= 3.13) - a pre-existing issue unrelated to this change, resolved by the companion test-suite modernization PR.

tolgakaratas:

eb4x:
I believe we're actually aware of this, because we do this workaround in testing.
https://github.com/ansible-collections/community.zabbix/blob/main/molecule/zabbix_web/prepare.yml#L65-L75

I think this comes down to a "Separation of Concerns" direction this project took some while ago. We don't install or configure postgresql/timescale nor mysql/mariadb, so we shouln't be the ones to install or configure apache or nginx, just supply configuration to a working installation that meets the requirements.

PR #1761, opened 2026-07-06
ci: import Zabbix arm64 GPG key for fping on EL aarch64 in molecule prepare

tolgakaratas:
##### SUMMARY
On EL aarch64 the Zabbix "non-supported" repository signs packages such as `fping` (pulled in by `zabbix-server` / `zabbix-proxy`) with a different GPG key than the amd64 packages, and the `zabbix_repo` role does not install it. The dnf package step then fails GPG validation with "Public key ... is not installed". This pre-imports that key in the server/proxy molecule prepare so the aarch64 role tests can install their packages. Idempotent and RedHat-only.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
molecule/zabbix_server/prepare.yml, molecule/zabbix_proxy/prepare.yml

##### ADDITIONAL INFORMATION
Verified by installing the fping package under GPG check on an EL aarch64 container; the server/proxy role tests pass on arm64 on our fork (see the validation note in the comments). Test-only change (no user-facing role code), so no changelog fragment.

### DIFF
--- a/molecule/zabbix_proxy/prepare.yml
+++ b/molecule/zabbix_proxy/prepare.yml
@@ -12,6 +12,16 @@
         dest: /usr/sbin/policy-rc.d
       when: ansible_facts['os_family'] == 'Debian'
 
+    # On EL aarch64, Zabbix signs its "non-supported" repo packages (e.g. fping, pulled in by
+    # zabbix-proxy) with a key that the zabbix_repo role does not install (the amd64 packages use
+    # a different key), so the dnf module fails GPG validation with "Public key ... is not
+    # installed". Pre-import that Zabbix key. Idempotent; RedHat only; unused where not needed.
+    - name: "Import Zabbix non-supported repo GPG key (EL aarch64 fping)"
+      ansible.builtin.rpm_key:
+        state: present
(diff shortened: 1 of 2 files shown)


tolgakaratas:
For transparency: this change was developed and validated as part of a larger test-suite modernization, then split into focused, single-logical-change PRs per CONTRIBUTING. The whole set was validated green together on our fork's GitHub Actions before splitting.

Green runs on our fork on a commit already containing this change (server/proxy pass on arm64, i.e. the fping GPG key is imported and packages install):

- zabbix_server: https://github.com/Denomas/community.zabbix/actions/runs/28727055884
- zabbix_proxy: https://github.com/Denomas/community.zabbix/actions/runs/28727055856

Note: on branches based on the current upstream `main`, the `Sanity (Ⓐdevel+py3.12)` job fails at dependency install (`ansible-core devel` now requires Python >= 3.13) - a pre-existing issue unrelated to this change, resolved by the companion test-suite modernization PR.

tolgakaratas:

eb4x:
This feels like something that should be addressed in the `zabbix_repo` role, right?

PR #1762, opened 2026-07-06
ci: modernize and instrument the collection test suite

tolgakaratas:
##### SUMMARY
Brings the GitHub Actions test matrix up to the versions this collection supports today and makes results observable. No production role/plugin code is changed here.

This supersedes #1757 (which carried the first slice - the initial ansible-core/python matrix bump and dropping the generated inventory); everything in #1757 is included here. Running the fuller matrix this adds surfaces several real, mostly pre-existing role bugs; those are fixed in the companion PRs (role version maps #1759, zabbix_web apache #1760, EL aarch64 fping #1761, trigger idempotence #1758) so the matrix is green end to end.

- Add a reusable `role-test.yml` (`workflow_call`) that the agent/server/proxy/javagateway/web workflows call, replacing large blocks of duplicated matrix YAML; parametrize the molecule scenarios via env vars.
- Expand the role matrix to distro x {amd64, arm64} x ansible-core {2.16, 2.21} x Zabbix {6.0, 7.0, 7.4}, with per-role excludes computed from real Zabbix package availability; re-run role workflows when the shared engine changes; add `workflow_dispatch` for manual re-trigger.
- Modernize the integration/sanity matrix to current ansible-core/python; drop EOL Zabbix 7.2; integration covers Zabbix 6.0/7.0/7.4 (plus 8.0 trunk).
- Unpin `ansible-core<2.19` in `molecule/requirements.txt`.
- Add observability: JUnit reporting, per-job step-summary result tables, failure-log artifacts and aggregate gate checks, via composite actions under `.github/actions/`. Retry `ansible-galaxy install` to survive transient Galaxy 5xx.
- Stop tracking the auto-generated `tests/integration/inventory` (it embeds the local `ansible_python_interpreter` path) and gitignore it.

No breaking changes; the support floor (ansible-core 2.16) is preserved.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
.github/workflows (role-test, agent, server, proxy, javagateway, web, plugins-integration-full, plugins-integration-targeted, repo-sanity, build_collection), .github/actions, molecule, .gitignore

##### ADDITIONAL INFORMATION
[...]

PR #1765, opened 2026-07-06
fix: zabbix_inventory use selectHostGroups host groups on Zabbix 7.0+

### LINKED ISSUE #1507: [FIX] zabbix_inventory  parameter add_zabbix_groups is broken with Zabbix 7.2
Versions:

community.zabbix: 3.3.0
Zabbix: 7.2.5

Parameter `selectGroups` which returns groups object  has been deprecated at Zabbix Version 7.0 and removed at version 7.2
https://www.zabbix.com/documentation/7.0/en/manual/api/reference/host/get
https://www.zabbix.com/documentation/current/en/manual/api/reference/host/get

Now we need to use `selectHostGroups` parameter which returns `hostgroups` object

Diff:
https://github.com/ansible-collections/community.zabbix/blob/418fbd7688871e48888a354e2a1fbebdb1b6b9b1/plugins/inventory/zabbix_inventory.py#L463

```
455c455
<                     'selectGroups': ['name']
---
>                     'selectHostGroups': ['name']
463,464c463,464
<                 if len(record['groups']) >= 1:
<                     for group in record['groups']:
---
>                 if len(record['hostgroups']) >= 1:
>                     for group in record['hostgroups']:
```

tolgakaratas:
##### SUMMARY
Fixes #1507. The `host.get` parameter `selectGroups` was deprecated in Zabbix 7.0 and removed in 7.2 (7.4 ignores it and returns hosts without a group field). As a result the `zabbix_inventory` plugin's `add_zabbix_groups` path returned no groups on Zabbix 7.2+, and a user-provided `host_zapi_query` with `selectGroups` silently returned no group data. The legacy `scripts/inventory/zabbix.py` had the same problem.

This makes host-group querying version-aware (mirroring the existing pattern in the modules):
- Use `selectHostGroups` on Zabbix >= 7.0 and `selectGroups` on 6.0, reading the response from `hostgroups` or `groups` accordingly.
- Auto-translate a user-provided `host_zapi_query` group selector by server version, so Zabbix 6.0 LTS keeps working, and document the new `selectHostGroups` sub-option.
- Apply the same version-aware fix to the legacy `scripts/inventory/zabbix.py`.
- Add a real-Zabbix integration target (`test_zabbix_inventory`, depends on `setup_zabbix`) that creates a host in a group and asserts the plugin (`add_zabbix_groups` and `host_zapi_query`) and the legacy script both surface it.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zabbix_inventory (inventory plugin); scripts/inventory/zabbix.py

##### ADDITIONAL INFORMATION
Validated against real Zabbix on both 7.4 (7.4.11) and 6.0 (6.0.46) via the collection's docker-compose stack and `ansible-test integration test_zabbix_inventory` - all three exercises (plugin `add_zabbix_groups`, plugin `host_zapi_query`, legacy script) pass on both versions. The bug is real on 7.4: a raw `host.get` with `selectGroups: ['name']` returns hosts with no group field, while `selectHostGroups: ['name']` returns `hostgroups`. `ansible-test sanity` on the changed files and `yamllint` are clean.

### DIFF
--- a/changelogs/fragments/1507-zabbix-inventory-selecthostgroups.yaml
+++ b/changelogs/fragments/1507-zabbix-inventory-selecthostgroups.yaml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - zabbix_inventory - use the ``selectHostGroups`` host.get parameter on Zabbix 7.0+ (``selectGroups`` was deprecated in 7.0 and removed in 7.2) and translate user-provided ``host_zapi_query`` group selectors by server version, keeping Zabbix 6.0 LTS working (https://github.com/ansible-collections/community.zabbix/issues/1507).
+  - zabbix inventory script - use the ``selectHostGroups`` host.get parameter on Zabbix 7.0+ (``selectGroups`` removed in 7.2) so the legacy ``scripts/inventory/zabbix.py`` works on current Zabbix (https://github.com/ansible-collections/community.zabbix/issues/1507).
--- a/plugins/inventory/zabbix_inventory.py
+++ b/plugins/inventory/zabbix_inventory.py
@@ -87,6 +87,18 @@ options:
                     - Can be limited to different fields for example setting the vaule to ['name'] will only return the name
                     - Additional fields can be specified by comma seperated value ['name', 'field2']
                     - Please see U(https://www.zabbix.com/documentation/current/manual/api/reference/hostgroup/object) for more details on field names
+                    - Removed in Zabbix 7.2; the plugin automatically uses selectHostGroups on Zabbix >= 7.0.
+            selectHostGroups:
+                type: str
+                description:
+                    - query
+                    - Return a hostgroups property with host groups data that the host belongs to.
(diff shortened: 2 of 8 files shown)


tolgakaratas:
Validation was done against real Zabbix (not mocked): the docker-compose stack was brought up for both `zabbix_version=7.4` and `6.0`, and `ansible-test integration test_zabbix_inventory` passed on both - the plugin (`add_zabbix_groups` and `host_zapi_query`) and the legacy script all surface the host's group. On 7.4 a raw `host.get` with `selectGroups` returns no group field while `selectHostGroups` returns `hostgroups`, confirming the fix is what makes group data come back.

Note: on branches based on the current upstream `main`, the `Sanity (Ⓐdevel+py3.12)` check fails at dependency install (`ansible-core devel` now requires Python >= 3.13) - a pre-existing, unrelated issue addressed by the CI test-suite modernization PR #1762.

